PR #17, opened 2016-04-19
apollo-react vs react-apollo

johnthepink:
Which are we going with? Looks like the npm package is `react-apollo`, which means the docs need to be updated here either way.


stubailo:
Yeah I think we went with `react-apollo` in the end because that's the convention for some other stuff.

Issue #30, opened 2016-04-24
Experience with mapQueriesToProps


abhiaiyer91:
first i'll say, `mapQueriesToProps` is pretty cool! To me, this function means we'll be mapping the result of a graphql query. Which in fact we do

```
{
  loading: false,
  result: {
    posts: []
  }
}
```

My question is, why tuck it away under a results key? Wouldn't it be nice to just have our prop
```js
{
  loading: false,
   posts: []
} 
```
I would then in my component much rather do 

```js
function App({postsData}) {
  const { posts = [], loading } = postsData;
  if ( loading) {
    return <Loader/>
  }
  return (
    <div>{posts.map(({ title }) => {
      return (
        <div>
          <h1>{title}</h1>
        </div>
      )
    })}</div>
  )
}

const AppWithData = connect({
  mapQueriesToProps() {
    return {
      postsData: {
        query: `
            {
              posts {
                title
              }
            }
          `
      }
    };
  }
})(App);
```

Essentially when the query yields, can we just extend this state structure? If you guys are down for this change, I can make it happen

jbaxleyiii:
@abhiaiyer91 💯 go for it! Looks great!

jbaxleyiii:
When you do it, will you add a warning about key collision of someone uses `loading` or `errors` as the prop key?

abhiaiyer91:
Yeah i think in mutations we still would want to interact with the values of our data directly rather tucked away in results. And checking for this data should be best practice regardless if result existed or not.

jbaxleyiii:
@abhiaiyer91 PR https://github.com/apollostack/react-apollo/pull/31

PR #31, opened 2016-04-24
remove result in favor of the key's returned by the query

jbaxleyiii:
Discussion in #30 

cc @abhiaiyer91

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,9 +2,11 @@
 
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
-### v0.1.6
+### v0.2.0
 
-Remove `result` key in favor of dynamic key from query or mutation. (https://github.com/apollostack/react-apollo/pull/31)
+**Breaking change:**
+
+Remove `result` key in favor of dynamic key matching root fields of the query or mutation. (https://github.com/apollostack/react-apollo/pull/31)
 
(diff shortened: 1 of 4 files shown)


jbaxleyiii:
I'm going to give this a day for people to chime in if they are interested / opposed to this change.

cc @stubailo @apollostack/community-contributors @johnthepink

johnthepink:
👍

abhiaiyer91:
I'm in :)

abhiaiyer91:
Thanks! I was going to do this but thanks for going for it!

jbaxleyiii:
@abhiaiyer91 I had some down time today 🎈

stubailo:
I think this should be `0.2` to conform to NPM's semver handling!

stubailo:
I think we should add a warning in the case where the query returns a property that conflicts with other things we return like `refetch`, `error`, `loading`. It's easy to work around with an alias, but it would be great for people to find out about the problem early!

It would be a bit weird to have a root query called `error`, but who knows, especially as we add more metadata.

abhiaiyer91:
Yes we can also write some Eslint rules for Apollo Query data, and ship that tooling?

stubailo:
> Eslint rules for Apollo Query data

What would they do? We can add it here maybe: https://github.com/apollostack/eslint-plugin-graphql

Since a plugin can ship multiple rules, we can add as many as we want to that plugin.

jbaxleyiii:
@stubailo I'll update this PR tonight 🎉

jbaxleyiii:
@stubailo this one should be good to go. I'm good with calling it discussed and merging if you are

PR #35, opened 2016-04-29
ApolloProvider

abhiaiyer91:
In reference to this issue [#142](https://github.com/apollostack/apollo-client/issues/142)
This PR is the first step of providing Apollo users a direct integration with the Redux library and set of existing tools.

### What are we shipping?

I think to provide a seamless integration into current Redux workflows, exporting an `ApolloProvider` allows us to maintain the Apollo integration a layer above `ReactRedux`'s `Provider` implementation.

### Usage

```js
import React from 'react';
import { render } from 'react-dom';
import Store from './store';

const client = new ApolloClient();

Meteor.startup(() => {
  render(<ApolloProvider store={Store} client={client}>
    <App />
  </ApolloProvider>, document.getElementById('app'));
});
```

To note: `store` is a required prop

### Benefits
On the surface this change doesn't seem like a big deal. And really it isn't far off then what `react-apollo` ships, but:
1. this allows `react-apollo` to gain upstream changes to the `Provider` component from `react-redux`. 2. helps alleviate confusion between large apps that may be fragmented, meaning they don't have clear component trees e.g. multiple `Providers` who may be confused what `Provider` they are actually using. People don't read 🚨

### DIFF
--- a/src/ApolloProvider.tsx
+++ b/src/ApolloProvider.tsx
@@ -1,0 +1,60 @@
+/// <reference path="../typings/main.d.ts" />
+
+import {
+    Component,
+    PropTypes,
+    Children,
+} from 'react';
+
+import {
+    Provider,
+} from 'react-redux';
+
+import {
+    Store
+} from 'redux';
+
+import ApolloClient from 'apollo-client';
+
+export declare interface ProviderProps {
+    store?: Store<any>;
+    client: ApolloClient;
+}
+
+export default class ApolloProvider extends Component<ProviderProps, any> {
+    static propTypes = {
+        store: PropTypes.shape({
+            subscribe: PropTypes.func.isRequired,
+            dispatch: PropTypes.func.isRequired,
+            getState: PropTypes.func.isRequired,
+        }),
+        client: PropTypes.object.isRequired,
+        children: PropTypes.element.isRequired,
+    };
+
+    static childContextTypes = {
+        store: PropTypes.object.isRequired,
+        client: PropTypes.object.isRequired,
+    };
+
+    public store: Store<any>;
+    public client: ApolloClient;
+
+    getChildContext() {
+        const { client } = this.props;
+        return {
+            client
+        }
+    }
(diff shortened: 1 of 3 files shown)


src/ApolloProvider.tsx
```diff
@@ -0,0 +1,60 @@
+/// <reference path="../typings/main.d.ts" />
+/* tslint:disable:no-unused-variable */
+import * as React from 'react';
```
abhiaiyer91: weird errors in compile. Anyone know what this is all about?

stubailo:
Yeah I think something as simple as changing the export name can be good to avoid confusion between the two.

jbaxleyiii:
@abhiaiyer91 I'm not a fan of having to pass a store. It locks us into redux (which I like redux fwiw) and exposing an internal part of the client (that is exposed for advanced reasons) adds a barrier to usage in my opinion.

abhiaiyer91:
@jbaxleyiii I agree with you, but this component is strictly for Redux implementation no? Unless we want to use this at the forefront, then what I'll do is make it behave like the current react-apollo `Provider` no problems!

jbaxleyiii:
@abhiaiyer91 ah I see, this wouldnt be the most commonly used provider. 

One thing I was wondering about is composing with react-redux provider for ours and exposing that composition method so people already using provider wouldn't need to replace, just compose the Apollo addition 

That way there is only one provider for clarity but we still get the benefits mentioned above. 

Thoughts?

abhiaiyer91:
I am down with that @jbaxleyiii. That will be a lot cleaner. What do you think @stubailo 

 I'll close this PR and make that happen.

jbaxleyiii:
@abhiaiyer91 awesome! Thanks!

stubailo:
Great idea! But I still think we could rename `Provider` to `ApolloProvider` - what do you think?

abhiaiyer91:
Yeah I'll do that in another PR

PR #44, opened 2016-05-06
Update apollo-client, increase performance, and pass polling props

jbaxleyiii:
This PR brings support for polling with the client as well as passes `stopPolling` and `startPolling` to the client.

It also adds a check to prevent rerenders if the data doesn't change.


Changelog
```diff
@@ -2,6 +2,15 @@
 
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
+### v0.3.2
+
+Feature: add `startPolling` and `stopPolling` to the prop object for queries
+Bug: Fix bug where full options were not being passed to watchQuery
+
+### v0.3.1
```
stubailo: Why two separate versions?

jbaxleyiii: 0.3.1 was already published but not updated in the changelog

jbaxleyiii:
@stubailo this should be good to go. If you are good with it, I'll merge!

ambient.d.ts
```diff
@@ -1,9 +1,9 @@
 declare module "lodash.isobject" {
   import main = require('~lodash/index');
-  export = main.isObject;
+  // export = main.isObject;
```
stubailo: What is this about?

stubailo:
Looks good to me! Only thing is I'm not sure what's with the commented typings.

jbaxleyiii:
Oh yeah! I upgraded my typings locally. I think it was seeing ambient.d.ts and the generated files. I'm not really sure, need to investigate more

stubailo:
:shipit: looks good!

Issue #25, opened 2016-04-21
Refactor test


jbaxleyiii:
Right now the tests are overly verbose in their boilerplate, and they don't accurately test common lifecycle events that can happen with changing props.

### Todo:

- [ ] split the connect test into files for `prop api`, `redux integration`, `queries`, and `mutations`
- [ ] create reusable boilerplate
- [ ] better test lifecycle events

Hopefully this will clean up some issues like https://github.com/apollostack/react-apollo/issues/20

jbaxleyiii:
Incrementally better with #50

PR #50, opened 2016-05-11
split tests into more manageable files

jbaxleyiii:
An initial start at #25


stubailo:
This is great! :shipit:

Issue #20, opened 2016-04-19
mapStateToProps params


johnthepink:
It looks like `mapStateToProps` currently gets called with just the redux state. The docs show that it gets called with an object containing `ownProps` and `state`. Is it intended to contain `ownProps` as well? Or should we update the docs? cc @jbaxleyiii

jbaxleyiii:
@johnthepink it should have ownProps. I thought I had a test in place for that?

stubailo:
I had a few issues with various arguments as well. I don't think our tests totally cover the different lifecycle events with changes coming from props, redux, etc.

jbaxleyiii:
@stubailo the tests are pretty minimal right now. @johnthepink would you be up go taking that for me today and tomorrow? Trying to be better about actually being on vacation ;)

stubailo:
I managed to build an example app on top of it, so I'm happy! No rush on making these improvements, I think - vacation away!

jbaxleyiii:
@stubailo @johnthepink is triumphantly moving our stavj to Apollo now so he may need some changes.

I'm really happy it's all decent ly coming together though!

johnthepink:
Yep! I'll jump in on fixing some stuff here soon. Found a couple other things.

jbaxleyiii:
Fixed with #51

PR #51, opened 2016-05-11
Performance improvements, and better lifecycle support

### LINKED ISSUE #41: Inconsistent state in mapMutationsToProps
[Originally posted](https://github.com/apollostack/apollo-client/issues/170) by @deoqc on apollo-client:

------

I am using redux-form and have 2 forms (lets say FormA and FormB, 'living' in `state.form.FormA` and `state.form.FormB`). 

The problem is that `FormB` is not always showing in the state when using `mapMutationsToProps`, but it is correct in `mapStateToProps` or in the store (as shown by devtools).

---

To be more precise, my code is something like this:
```Javascript
const mapMutationsToProps = ({ state }) => ({
  onMutation: () => ({
    mutation: `
      mutation myAwesomeMutation (
        $formKeys: String!
      ) {
        awesomeMutation( 
          formKeys: $formKeys
        ) {
          passThroughtInput
        }
      }
    `,
    variables: {
      formKeys: JSON.stringify(Object.keys(state.form)),
    },
  })
});
```
and the mutation only returns the same input (`passThroughtInput === formKeys`)

To check, I also have the following:
```Javascript
const mapStateToProps = (state) => ({
  formKeys: JSON.stringify(Object.keys(state.form)),
});
```

The callback to the `onMutation` prints the results on console, and the `formKeys` is in a `Text` tag on screen, and also using devtools (so I compare the 3).

As said before, mapStateToProps and store devtools are correct. Mutation is missing (at least most of the time) the state.FormB key (but showing state.FormA).

If you guys don't have any clues, I can try to do a reproducible repo.

jbaxleyiii:
This PR does a few things (it can be split up if needed).

- [x] Queries re run on when the redux state (aside from apollo) change
- [x] `mapMutationsToProps` rebuild on change
- [x] update Changelog.md
- [x] diff queries before rerunning to see if they have changed

Fixes #41 and #20

cc @stubailo @johnthepink

### DIFF
--- a/src/connect.tsx
+++ b/src/connect.tsx
@@ -46,7 +46,7 @@ export declare interface ConnectOptions {
   options?: IConnectOptions;
   mergeProps?(stateProps: Object, dispatchProps: Object, ownProps: Object): Object;
   mapQueriesToProps?(opts: MapQueriesToPropsOptions): Object; // WatchQueryHandle
-  mapMutationsToProps?(opts: MapQueriesToPropsOptions): Object; // Mutation Handle
+  mapMutationsToProps?(opts: MapMutationsToPropsOptions): Object; // Mutation Handle
 };
 
 const defaultMapQueriesToProps = opts => ({ });
--- a/src/connect.tsx
+++ b/src/connect.tsx
@@ -27,6 +27,7 @@ import ApolloClient, { readQueryFromStore } from 'apollo-client';
 
 import {
   GraphQLResult,
+  Document,
 } from 'graphql';
 
 export declare interface MapQueriesToPropsOptions {
@@ -45,11 +46,11 @@ export declare interface ConnectOptions {
   options?: IConnectOptions;
   mergeProps?(stateProps: any, dispatchProps: any, ownProps: any): any;
   mapQueriesToProps?(opts: MapQueriesToPropsOptions): any; // WatchQueryHandle
-  mapMutationsToProps?(opts: MapMutationsToPropsOptions): any; // Mutation Handle
+  mapMutationsToProps?(): any; // Mutation Handle
 };
 
 const defaultMapQueriesToProps = opts => ({ });
-const defaultMapMutationsToProps = opts => ({ });
+const defaultMapMutationsToProps = () => ({ });
 const defaultQueryData = {
   loading: true,
   errors: null,
@@ -113,15 +114,17 @@ export default function connect(opts?: ConnectOptions) {
         client: PropTypes.object.isRequired,
       };
 
-      // react and react dev tools (HMR) needs
+      // react / redux and react dev tools (HMR) needs
       public state: any; // redux state
       public props: any; // passed props
       public version: number;
+      private unsubscribeFromStore: Function;
 
       // data storage
       private store: Store<any>;
       private client: ApolloClient; // apollo client
-      private data: any; // apollo data
+      private data: Object; // apollo data
+      private previousState: Object;
 
       // request / action storage
       private queryHandles: any;
@@ -131,6 +134,7 @@ export default function connect(opts?: ConnectOptions) {
       private haveOwnPropsChanged: boolean;
       private hasQueryDataChanged: boolean;
       private hasMutationDataChanged: boolean;
+      private hasOwnStateChanged: boolean;
 
       // the element to render
       private renderedElement: any;
@@ -150,15 +154,40 @@ export default function connect(opts?: ConnectOptions) {
 
         const storeState = this.store.getState();
         this.state = assign({}, storeState);
+        this.previousState = storeState;
 
         this.data = {};
         this.mutations = {};
       }
 
       componentWillMount() {
-        const { props, state } = this;
-        this.subscribeToAllQueries(props, state);
-        this.createAllMutationHandles(props, state);
+        const { props } = this;
+        this.subscribeToAllQueries(props);
+        this.createAllMutationHandles(props);
+      }
+
+      componentDidMount() {
+        const { store, props } = this;
+        const { reduxRootKey } = this.client;
+
+        this.unsubscribeFromStore = store.subscribe(() => {
+          let newState = assign({}, store.getState());
+          let oldState = assign({}, this.previousState);
+
+          // we remove the apollo key from the store
+          // because incomming data would trigger unneccesary
+          // queries and mutations rebuilds
+          delete newState[reduxRootKey];
+          delete oldState[reduxRootKey];
+
+          if (!isEqual(oldState, newState)) {
+            this.previousState = newState;
+            this.hasOwnStateChanged = true;
+
+            this.unsubcribeAllQueries();
+            this.subscribeToAllQueries(props);
+          }
+        });
       }
 
       componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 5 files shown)


stubailo:
I think I would prefer `{ ownProps, state }` to be the first argument, not the last, because then it's a lot more consistent to find. I think this might get messy when you start having optional arguments in the mutations.

> diff queries before rerunning to see if they have changed

What does this mean?

stubailo:
@jbaxleyiii wait, did we ever decide that rebuilding the mutation functions on each re-render wouldn't work? Is there a downside to that approach, which wouldn't require changing the API to add extra arguments?

jbaxleyiii:
@stubailo since we rerun the `mapQueriesToProps` whenever prop or state changes, I am adding in a check to see if the resulting queries have changed. If not, I don't unsubscribe and rebuild them. So its a performance check to not handle all of the query lifecycle again for no reason.

I can rebuild the mutation functions on each re-render (I have it mostly done). The reason I diverged from that to the `{ state, ownProps }` as arguments was around the use case. Since mutations don't happen on their own (i.e. have to be called within the child component), I thought it would be more efficient to pass the state and props only at execution of the mutation. We can make them first or last for sure. My thought with last was in cases like so:

```diff
mapMutationsToProps = () => ({
  action: (my, args) => { // MutationObject }
})

// vs unused `{ ownProps, state }` 
mapMutationsToProps = () => ({
  action: ({ ownProps, state }, my, args) => { // MutationObject }
})
```

Thoughts?

deoqc:
Thx for the update. Any interface that you guys decide is just **great**, but if I was to do it maybe would be with something like:
```Javascript
const mapMutationsToProps = () => ({ state, ownProps }) => ({
  action: (my, args) => { ... }
});
```
It is much closer to the original one, it feels more natural... 

Also, arguments called in runtime `my` and `args` (not sure if this is the case, but it is like `raw` in the [docs](http://docs.apollostack.com/apollo-client/react.html) right?) should be separate from `state` and `ownProps`, and like this seems clearer to me.

jbaxleyiii:
Thanks for the feedback! I think that's one of the points behind this change in my mind. `my` and `args` are called at runtime, but you also want the current props and state to mutation execution as well. 

Keeping it as the current API is possible! I'll do a comparison branch and try to add in a performance test to see if my concerns about rebuilding the mutations each time is valid or not! (We are rebuilding queries each time so it's probably nothing!)

jbaxleyiii:
Ehhh, I'll just rewrite to keep the same API. The plumbing is there already and the execution won't be that much, if any perf loss.

stubailo:
Sounds good to me!

src/connect.tsx
```diff
@@ -43,9 +44,9 @@ export declare interface ConnectOptions {
   mapStateToProps?: IMapStateToProps;
   mapDispatchToProps?: IMapDispatchToProps;
   options?: IConnectOptions;
-  mergeProps?(stateProps: any, dispatchProps: any, ownProps: any): any;
-  mapQueriesToProps?(opts: MapQueriesToPropsOptions): any; // WatchQueryHandle
-  mapMutationsToProps?(opts: MapMutationsToPropsOptions): any; // Mutation Handle
+  mergeProps?(stateProps: Object, dispatchProps: Object, ownProps: Object): Object;
+  mapQueriesToProps?(opts: MapQueriesToPropsOptions): Object; // WatchQueryHandle
+  mapMutationsToProps?(opts: MapQueriesToPropsOptions): Object; // Mutation Handle
```
stubailo: Why is this now `MapQueriesToPropsOptions` instead of `MapMutationsToPropsOptions`?

jbaxleyiii: haha because I copy pasted wrong. Thanks!

stubailo:
@jbaxleyiii looks great to me! :shipit:

deoqc:
I think something broke =[. When I upgrade to `0.3.4` it is taking away too much time to start the app or really do anything at all. The xcode console (I'm using react native) is just spiting this:
```
equalObjects@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69878:31
baseIsEqualDeep@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69668:20
baseIsEqual@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69610:23
equalObjects@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69878:31
baseIsEqualDeep@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69668:20
baseIsEqual@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69610:23
equalObjects@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69878:31
baseIsEqualDeep@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69668:20
baseIsEqual@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69610:23
isEqual@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:70038:19
http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69118:12
dispatch@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:68541:13
dispatch@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:82792:21
http://localhost:8081/index.ios.bundle?platform=ios&dev=true:71511:21
dispatch@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:68892:17
fetchQuery@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73879:20
startQuery@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73933:16
http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73777:29
subscribe@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:78901:58
subscribe@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:73702:39
handleQueryData@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69220:40
subscribeToAllQueries@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69163:21
componentWillMount@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:69085:27
mountComponent@http://localhost:8081/index.ios.bundle?platform=ios&dev=true:15996:24
 ```

**Comment by stubailo** [comment]
Could this be deep equal failing on a circular reference or something?

**Comment by jbaxleyiii** [comment]
@deoqc could you make a reproduction repo I can debug?

**Comment by jbaxleyiii** [comment]
It may be a circular reference, or the way I'm rebuilding queries when props / state change is getting stuck in a loop

**Comment by deoqc** [comment]
I found **my** mistake. Sorry!

But it could have a test and a nice error message for this.

I think will be a common mistake, if it is indeed a mistake - I had this code in `0.3.3` and didn't crash:

```Javascript
const mapMutationsToProps = () => ({
  myBuggyMutation: (myVariable) => ({
    mutation: gql`
      mutation myMutation($myVariable: String!) { # --> same name here
        myMutation (myVariable: $myVariable) # --> and here
      }
    `,
    variables: {
      myVariable,
    },
  }),
});
```

So I replaced it for:

```Javascript
const mapMutationsToProps = () => ({
  myCorrectMutation: (myVariable) => ({
    mutation: gql`
      mutation onMyMutation($myVariable: String!) { # --> different name here
        myMutation (myVariable: $myVariable) # --> and here
      }
    `,
    variables: {
      myVariable,
    },
  }),
});
```

stubailo:
Also @deoqc can you post a new issue for this?

jbaxleyiii:
@deoqc that's definitely a bug!

PR #73, opened 2016-06-16
Provide watchQuery to components via connect.

dahjelle:
I was recently looking for a way to subscribe to `watchQuery` directly, rather than through the usual `mapQueriesToProps`, and noticed that, while `query` and `mutate` are passed via `connect`, `watchQuery` is not. This PR changes that.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
+### vNEXT
+
+Feature: provide add `watchQuery` to components via `connect`
+
 ### v.0.3.8
 
 Bug: Don't use old props on store change change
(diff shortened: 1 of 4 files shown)


jbaxleyiii:
@dahjelle thanks for the PR and tests. Any reason in particular why you would not use the map method?

dahjelle:
@jbaxleyiii I have two queries, where one query is dependent on the results from the other. Using `query` and `watchQuery` manually seemed like the most straightforward approach to me; is there another approach that you'd recommend? I see my git history has a solution with nested components, but that seems a bit clumsy.

I suppose the other reason is just for symmetry; it seems like it'd be nice to be able to say "the Apollo core API is also available via `props`" without qualification, but that just my 2¢.

jbaxleyiii:
@dahjelle I think that is a great use case. Can you sign the CLA? Then I'll merge this through and release it.

Could you also add the change to the changelog?

dahjelle:
@jbaxleyiii When adding to the changelog, should I do it under v0.3.9 or vNEXT a la [apollo-client](https://github.com/apollostack/apollo-client/blob/master/CHANGELOG.md)?

jbaxleyiii:
@dahjelle under vNEXT 👍

jbaxleyiii:
also, thanks so much for this PR and addition! 🎉

dahjelle:
My pleasure—thank **you**! :-D

jbaxleyiii:
@dahjelle released at `0.3.9` on npm!

dahjelle:
👍

PR #77, opened 2016-06-20
remove typescript declarations from output

AlmaasAre:
Suggesting to remove typescript declarations from the output. When i try to use this library with typescript it uses the definitions outputted, and they don't work because typings is not included. 

Tried to include typings, but got a lot of errors.

Best would be to include the typings folder in the output?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "sourceMap": true,
-    "declaration": true,
+    "declaration": false,
     "noImplicitAny": false,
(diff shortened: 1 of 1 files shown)


AlmaasAre:
Just publish to npm and it should be golden

jbaxleyiii:
@AlmaasAre and this will still let users use the type definitions of this project? I thought if the files don't get generated, then typescript won't seem the from other projects?

AlmaasAre:
No, they have to create their own. But typescript typings are useless now
anyways if you don't include typings or rm the import on top in each file.

PR #75, opened 2016-06-17
added inital test coverage for SSR

jbaxleyiii:
This should also make development of SSR features easier to test

Fixes https://github.com/apollostack/react-apollo/issues/72

PR #80, opened 2016-06-24
Wait until the component *has mounted* to set hasMounted.

tmeasday:
This fixes my repro as well as my other SSR examples!

### DIFF
--- a/src/connect.tsx
+++ b/src/connect.tsx
@@ -165,10 +165,13 @@ export default function connect(opts?: ConnectOptions) {
 
       componentWillMount() {
         const { props } = this;
-        this.hasMounted = true;
         this.subscribeToAllQueries(props);
         this.createAllMutationHandles(props);
         this.bindStoreUpdates();
+      }
+
+      componentDidMount() {
+        this.hasMounted = true;
       }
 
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@tmeasday yep! I came to the same conclusion. We will need to catch all results that come between `componentWillMount` and `componentDidMount` and apply them after mounting. We also need to separate our client test (with jsdom) and our server test (without) to ensure we are correctly testing both environments.

I know this is a pretty big blocker for you right now. I'm swamped at  work for the next couple days, but should be able to get to this first of next week if you haven't gotten to it first

jbaxleyiii:
Fixed by #75

tmeasday:
@jbaxleyiii is it actually possible to receive data back from apollo in between `willMount` and `hasMounted`?

I honestly don't know: Rendering in React is synchronous, so the naive answer would be no. But then I know it uses batching, so it seems *possible*, if unlikely that apollo requests could come back while it is still rendering. 

Happy to take a look at the tests though if they still need fixing. Perhaps open an issue describing what needs to happen and I'll make it so (if I can!)

PR #83, opened 2016-06-24
Integrated SSR

jbaxleyiii:
active work on https://github.com/apollostack/react-apollo/issues/54

```es6
// no changes to client :tada:

// server application code (integrated usage)
import { renderToStringWithData } from "react-apollo/server"

// during request
const markup = await renderToStringWithData(app)


// server application code (custom usage)
import { getDataFromTree } from "react-apollo/server"

// during request
getDataFromTree(app, ).then(({ initialState, store, client }) => {
  // markup with data from requests
});
```

The client does have the option to ignore SSR for particular queries

```es6
const WrappedElement = connect({
  mapQueriesToProps: () => ({
    data: { query, ssr: false }, // wont block the SSR
  })
})(Element);

```

## getDataFromTree

The `getDataFromTree` method takes your react tree and returns an object with `initialState`, the apollo client (as `client`) and the redux store as `store`.
`initialState` is the hydrated data of your redux store prior to app rendering. Either `initialState` or `store.getState()` can be used for server side rehydration. 

## renderToStringWithData
The `renderToStringWithData` takes your react tree and returns a promise that resolves to your stringified tree with all data requirements. It also injects a script tag that includes `window. __APOLLO_STATE__ ` which equals the full redux store for hyrdration. 

## Server notes:
When creating the client on the server, it is best to use `ssrMode: true`. This prevents unneeded force refetching in the tree walking.

## Client notes:
When creating new client, you can pass `initialState: __APOLLO_STATE__ ` to rehydrate which will stop the client from trying to requery data.

@stubailo @tmeasday this needs more tests, but I thinks its pretty close

> I'd like to experiment with adding a streaming API so you can stream the html to the client. But I'll probably do this after the API refactor

### DIFF
--- a/src/getData.ts
+++ b/src/getData.ts
@@ -1,0 +1,69 @@
+
+import { Children } from 'react';
+import ApolloClient from 'apollo-client';
+import flatten = require('lodash.flatten');
+import assign = require('object-assign');
+
+export default function getData(tree, props: any = {}): Promise<any> {
+  let client;
+  let store;
+  function getQueriesFromTree(components, queries = []) {
+    Children.forEach(components, (child: any) => {
+      if (!child) {
+        return;
+      }
+
+      if (!store && child.props && child.props.store) {
+        store = store;
+      }
+
+      // find the client in the tree
+      if (!client && child.props && child.props.client instanceof ApolloClient) {
+        client = child.props.client as ApolloClient;
+        if (!store) {
+          client.initStore();
+          store = client.store;
+        }
+      }
+
+      if (child.type && typeof child.type.mapQueriesToProps === 'function') {
+        queries.push(child.type.mapQueriesToProps);
+      }
+      if (child.props && child.props.children) {
+        getQueriesFromTree(child.props.children, queries);
+      }
+    });
+    return queries;
+  }
+
+  const rawQueries = getQueriesFromTree(tree);
+  if (!rawQueries.length || !client) {
+    return Promise.resolve(null);
+  }
+
+  const ownProps = props;
+  const state = store.getState();
+
+  let queries = rawQueries
+    .map(x => x({ ownProps, state }))
+    .map(x => {
+      const queryOptions = [];
+      for (let key in x) {
+        if (!x[key].query || !x[key].ssr) {
+          continue;
+        }
+        queryOptions.push(x[key]);
+      }
+      return queryOptions;
+    });
+
+  queries = flatten(queries).map(x => {
+    return client.query(x)
+      .then(result => {
+        const { data, errors } = result;
+        return assign({ loading: false, errors }, data);
+      });
+  });
+
+  return Promise.all(queries);
+}
--- a/src/getData.ts
+++ b/src/getData.ts
@@ -1,69 +1,200 @@
 
-import { Children } from 'react';
+import { Children, createElement } from 'react';
 import ApolloClient from 'apollo-client';
 import flatten = require('lodash.flatten');
 import assign = require('object-assign');
 
-export default function getData(tree, props: any = {}): Promise<any> {
-  let client;
-  let store;
-  function getQueriesFromTree(components, queries = []) {
-    Children.forEach(components, (child: any) => {
-      if (!child) {
-        return;
-      }
+/*
 
-      if (!store && child.props && child.props.store) {
-        store = store;
-      }
+React components can return a `falsy` (null, false) value,
+representation of a native DOM component (such as <div /> or React.DOM.div())
+or another composite component. Components can have a render function (for components).
+They can also pass through children which we want to analyze as well.
 
-      // find the client in the tree
-      if (!client && child.props && child.props.client instanceof ApolloClient) {
-        client = child.props.client as ApolloClient;
-        if (!store) {
-          client.initStore();
-          store = client.store;
-        }
-      }
+To get data from `connect()` components we do a few things:
 
-      if (child.type && typeof child.type.mapQueriesToProps === 'function') {
-        queries.push(child.type.mapQueriesToProps);
-      }
-      if (child.props && child.props.children) {
-        getQueriesFromTree(child.props.children, queries);
-      }
+1. if ssr is not falsy, move the query to a place to batch call it
+
+Ideally, we go through the tree and find all `connect()`s (recursively going through tree)
+If we reach the end of all nodes, we kick off the queries. Once queries have returned,
+we try to go through their children components again to see if we discover any
+more queries. Then once we reach th end, we render the dom.
+
+We recursively do this until the tree is done.
+
+So! Given a component:
+
+1. See if it is falsy (end of line)
+2. Bulid the context and props (global props + defaultProps + parent props)
+3. See if the component is a `connect()`
+3a. Get the queries using props + state
+3b. as long as ssr != false, pass the query to the array to be called
+4. Create the component (or child if connect) (`componentWillMount` will run)
+5. Render the component
+6. Repeat
+
+*/
+
+declare interface Context {
+  client?: ApolloClient;
+  store?: any;
+  [key: string]: any;
+}
+
+declare interface QueryTreeArgument {
+  components: any;
+  queries?: any[];
+  defaultProps?: any;
+  context?: Context;
+}
+
+const defaultReactProps = {
+  loading: true,
+  errors: null,
+};
+
+
+export function getPropsFromChild(child, defaultProps = {}) {
+  const { props, type } = child;
+  let ownProps = assign(defaultProps, props);
+  if (type && type.defaultProps) ownProps = assign(defaultProps, type.defaultProps, props);
+  return ownProps;
+}
+
+export function getChildFromComponent(component) {
+  // See if this is a class, or stateless function
+  if (component && component.render) return component.render();
+  return component;
+}
+
+export function processQueries(queries, client): Promise<any> {
+  queries = flatten(queries)
+    .map((queryDetails: any) => {
+      const { query, component, ownProps, key } = queryDetails;
+      return client.query(query)
+        .then(result => {
+          const { data, errors } = result as any;
+          ownProps[key] = assign({ loading: false, errors }, data);
+          return {
+            component,
+            ownProps: assign({}, ownProps),
+          };
+        });
     });
-    return queries;
-  }
-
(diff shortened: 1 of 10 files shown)


Vanuan:
`      ssr: true, // block during SSR render
`

What's this and why it's needed?

jbaxleyiii:
@Vanuan that is an option for each query that tells react-apollo to actually fetch the data during rendering on the server

Vanuan:
What's `getData` for then?

Vanuan:
I mean, why can't it be true by default?

Vanuan:
So it means that some queries aren't meant to be executed during ssr?

Vanuan:
I believe, the most flexible API would be like this:

```javascript
import {  getQueries, runQueries } from "react-apollo"

let queries = getQueries(app, /* props to be pased as ownProps */);

/*
  filter out queries you don't want on server side
  according to routing or some other logic
*/
queries = ...

runQueries(queries).then((data) => {
  // in redux you dispatch query action
  // store.dispatch(querySucceeded({data}));
  // or update the store somehow else
  // fetched data is saved in the store 
  // Provider provides stored state to connected components during rendering
}, (err) => {
  // dispatch and render errors
});
```

Of course, dispatching can be hidden inside apollo, but it would be hard to follow and unclear whether it's dispatched or not.

src/getData.ts
```diff
@@ -0,0 +1,69 @@
...
+        queries.push(child.type.mapQueriesToProps);
+      }
+      if (child.props && child.props.children) {
+        getQueriesFromTree(child.props.children, queries);
+      }
+    });
+    return queries;
+  }
+
+  const rawQueries = getQueriesFromTree(tree);
```
Vanuan: Would be great if `getQueriesFromTree` is exported

jbaxleyiii: @Vanuan I think we would export it for sure once it is working correctly 👍

jbaxleyiii: I think exporting a few options to allow you to build your own rendering process, or just use react-apollo would be best:

i.e:
1. `getQueries, runQueries` for a fully custom solution
2. `getData` which runs all queries tagged as SSR
3. a `renderWithData` that returns back the full SSR payload with the initial state injected into the markup.

src/getData.ts
```diff
@@ -0,0 +1,69 @@
...
+          store = client.store;
+        }
+      }
+
+      if (child.type && typeof child.type.mapQueriesToProps === 'function') {
+        queries.push(child.type.mapQueriesToProps);
+      }
+      if (child.props && child.props.children) {
+        getQueriesFromTree(child.props.children, queries);
+      }
```
Vanuan: Render leaf components here?

    } else if (child.type) {
        let leaf = new child.type(child.props, {store, client});
        let newTree = leaf.render();
        getQueriesFromTree(newTree, queries);
    }

jbaxleyiii:
@Vanuan this is still very much a WIP

stubailo:
OMG can we turn this stupid bot off

stubailo:
I turned it off

src/server.ts
```diff
@@ -0,0 +1,232 @@
...
+const defaultReactProps = {
+  loading: true,
+  errors: null,
+};
+
+
+export function getPropsFromChild(child, defaultProps = {}) {
+  const { props, type } = child;
+  let ownProps = assign(defaultProps, props);
+  if (type && type.defaultProps) ownProps = assign(defaultProps, type.defaultProps, props);
```
tmeasday: Is this how react does it internally?

src/server.ts
```diff
@@ -0,0 +1,232 @@
...
+      const dataRequirements = child.type.mapQueriesToProps({ ownProps, state });
+      for (let query in dataRequirements) {
+        ownProps[query] = assign({}, defaultReactProps);
+        if (dataRequirements[query].ssr === false) continue; // don't run this on the server
+
+        queries.push({
+          key: query,
+          query: dataRequirements[query],
+          component: child.type.WrappedComponent,
+          ownProps,
```
tmeasday: Should we be pushing `context` in here as well so when it renders again after the query is done, it has the correct context?

src/server.ts
```diff
@@ -0,0 +1,232 @@
...
+      }
+
+      Element = createElement(child.type.WrappedComponent, ownProps) as any;
+    }
+
+    // try to see if this is a component, or a stateless component
+    if (!Element && typeof child.type === 'function') Element = { type: child.type };
+    const RenderedComponent = Element && Element.type && new Element.type(ownProps, context);
+
+    if (RenderedComponent && RenderedComponent.context) context = RenderedComponent.context;
```
tmeasday: Are we overriding the `context` that's returned from the function? I'm confused about `context` here. Is it the same thing as `getChildContext` and co use?

jbaxleyiii: Fixed in the latest which pulls context dynamically

src/server.ts
jbaxleyiii: Fixed in the latest code!

src/server.ts
jbaxleyiii: pretty much yes, if I'm reading this correctly: https://github.com/facebook/react/blob/49238b944046abf086ba80ab5e984ce8a8175a1d/src/isomorphic/classic/element/ReactElement.js#L207-L214

jbaxleyiii:
This could use a lot more testing for sure, but I'm going to cut a release so we can start trying it out in some applications. I've added tests for a few tree structures ranging from simple to mildly complex with mixed classes and components.

gauravtiwari:
@jbaxleyiii Awesome work. I can give it a try 👍

jbaxleyiii:
@gauravtiwari 🎉 thanks so much! I'd use the 0.3.17 version (has a bug fix)

gauravtiwari:
@jbaxleyiii I gave it a try and it seems that I am missing something. Could you a post a minimal setup example for SSR - including component and all (both client and on server) ?

```js
import React, from 'react';
import { connect } from 'react-apollo';

const SomeComponent = (props) => {
    const { data } = props;
    return (
      <div className="nav">
        something to render
      </div>
    );
}

function mapQueriesToProps({ ownProps, state }) {
  return {
    data: `{ someQuery }`
  };
};

const SomeComponentWithData = connect({
  mapQueriesToProps,
})(SomeComponent);

export default SomeComponentWithData;

```

```js
// Where to pass initial state or props?
 const htmlResult = await renderToStringWithData(SomeComponentWithData);
```

tmeasday:
Hey @jbaxleyiii -- worked great for me in this commit: https://github.com/apollostack/saturn/commit/9943e3b72e625bf70efc76f594ca4f0de8c4bef9 for an app which did SSR without queries on the server.

However, for githunt, it just hangs: https://github.com/apollostack/GitHunt/tree/saturn-ssr

I see some output from (I'm guessing) `react-apollo`:

```
[0] { shouldForceFetch: false }
[0] -------fetching query { kind: 'Document',
[0]   definitions:
[0]    [ { kind: 'OperationDefinition',
[0]        operation: 'query',
[0]        name: [Object],
[0]        variableDefinitions: [],
[0]        directives: [],
[0]        selectionSet: [Object] } ] }
[0] -------fetching query { kind: 'Document',
[0]   definitions:
[0]    [ { kind: 'OperationDefinition',
[0]        operation: 'query',
[0]        name: [Object],
[0]        variableDefinitions: [Object],
[0]        directives: [],
[0]        selectionSet: [Object] } ] }
```

gauravtiwari:
Hey @jbaxleyiii 

So, I tried again and it seems that, for NON-JS environment this won't work (unless the JS environment supports callbacks). The response returned is always empty as the server doesn't wait for callback to be finished.

jbaxleyiii:
@tmeasday I've seen it hang every so often for my stuff locally too. I'll be working on that today / tomorrow but would love some help if you can spare!

@gauravtiwari ah that makes sense! I need to find a way to make the getData sync for a lot of js use cases anyway. Would it work for you then?

tmeasday:
@jbaxleyiii I'd love to help although I'm not sure I'll have too much time today. I would say that for githunt it consistently, rather than occassionally hangs. Ping me on slack if I can help thgouhg.

tmeasday:
@jbaxleyiii Ok, I couldn't help myself. There are two problems that are causing the issue, when I hack around them, it works great!:

1) If there's a problem in `getDataFromTree`, the error is silently dropped. We should add a `catch` handler to the promise here: https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L185

2) The error comes because this line fails: https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L128

The reason is that (yay, thanks npm!) I end up with 2 copies of `apollo-client` installed in githunt, as I was npm-linking saturn (so it had it's own copy).

So it actually works if you run my `saturn-ssr` branch of githunt (yay!), but I think in general it's risky to check types like that unless you set a `peerDependency` on `apollo-client`[1] to ensure you get the "app's version" of apollo-client as a dep.

Oh, and as an aside on deeper calls to `getQueriesFromTree` you implicitly *don't* check the type, as you allow the context from the component to override here: https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L123

I'm not sure what the best solution is. One option is to just not check the type :/

[1] In short npm is terrible at helping you with this problem. I could rant for a while about it.

gauravtiwari:
@jbaxleyiii Yeah, I think so, because rendering react component normally on the server works and returns the response so, as long as the response from `react-apollo` is synchronous it should work.

src/server.ts
```diff
@@ -0,0 +1,195 @@
...
+      let markup = ReactDOM.renderToString(component);
+      let initialState = store.getState();
+      const key = client.reduxRootKey;
+      // XXX apollo client requires a lot in the store
+      // can we make this samller?
+      for (let queryId in initialState[key].queries) {
+        let fieldsToNotShip = ['minimizedQuery', 'minimizedQueryString'];
+        for (let field of fieldsToNotShip)  delete initialState[key].queries[queryId][field];
+      }
+      initialState = encodeURI(JSON.stringify(initialState));
```
Vanuan: I assume encodeURI is wrong? It'll result in
`Uncaught SyntaxError: Unexpected token %`

voidale:
Hey guys, anyone had any luck doing this with Meteor?

PR #84, opened 2016-06-28
Throw an error if the client is missing

Aweary:
The current error just states that it `"can't read initStore of undefined"`, which is a pretty cryptic error if you're not familiar with the internals of `ApolloProvider`.

This just throws a more friendly error.

### DIFF
--- a/src/ApolloProvider.tsx
+++ b/src/ApolloProvider.tsx
@@ -16,6 +16,8 @@ import {
 
 import ApolloClient from 'apollo-client';
 
+import invariant = require('invariant');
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@Aweary this is great! Could you rebase against master and if you look at the connect file, we are using `invariant` for throwing errors like this, would you switch to that instead of a thrown error?

Thanks!

Aweary:
@jbaxleyiii I rebased, switch to `invariant`, and squashed my commits 👍

PR #89, opened 2016-06-30
Error catching

### LINKED ISSUE #57: All Errors Catched
When using the `connect` method from `react-apollo`, all errors that occur in one of the child components, are catched by apollo and fail silently.

Development gets nearly impossible then.

Any ideas why that could be the case?
Using `apollo-client 0.3.5` and `react-apollo 0.3.3`

Thanks!

jbaxleyiii:
This should finally fix #57 

With patterns from https://github.com/reactjs/react-redux/commit/ea53d6fb076359a864a1d543568d951d4b3eab3d


stubailo:
Wow it's kinda ridiculous that we have to do this haha

PR #98, opened 2016-07-08
Don't strip data away from the component when the query errors

### LINKED ISSUE #97: Don't strip data away when the query errors (if there is existing data) in long-polling situations
See https://github.com/apollostack/apollo-client/issues/334. If I put some long polling on a query and that query fails, the data is not pulled anymore from the store but just disappears.

Example:
Prod example

rricard:
Fixes #97


jbaxleyiii:
Thanks @rricard!

PR #104, opened 2016-07-12
Add apollo-client v0.4.0.

dahjelle:


jbaxleyiii:
@dahjelle I believe there is a breaking change in apollo-client 0.4.0 around where refetch is located? I think we need to better test refetch and make sure this won't break?

stubailo:
Correct, when I ran tests using the new version they didn't pass specifically for refetch and start/stop polling

dahjelle:
Whoops! I didn't read the `apollo-client` change log closely enough. I'll close this PR. :-D

PR #105, opened 2016-07-12
Update to apollo client 0.4.0

stubailo:
A few hacks we should fixe before merging

### DIFF
--- a/src/connect.tsx
+++ b/src/connect.tsx
@@ -29,8 +29,11 @@ import ApolloClient, {
 
 import {
   ObservableQuery,
-  QuerySubscription,
 } from 'apollo-client/QueryManager';
+
+import {
+  Subscription,
+} from 'apollo-client/util/Observable';
 
 import {
   GraphQLResult,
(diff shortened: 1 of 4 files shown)


src/connect.tsx
```diff
@@ -254,33 +264,31 @@ export default function connect(opts?: ConnectOptions) {
...
-        if (isEqual(oldQueries, queryHandles)) {
+        if (isEqual(oldQueries, queryOptions)) {
           return false;
         } else if (oldQueries) {
           // unsubscribe from previous queries
           this.unsubcribeAllQueries();
         }
 
-        if (isObject(queryHandles) && Object.keys(queryHandles).length) {
-          this.queryHandles = queryHandles;
```
stubailo: As far as I know, this line was not correct - it was assigning query options to the same object we use to hold the observables/handles/subscriptions.

typings/modules/apollo-client/index.d.ts
```diff
@@ -274,12 +274,13 @@ import { Observable, Observer, Subscription } from '~apollo-client/util/Observab
 export class ObservableQuery extends Observable<GraphQLResult> {
     subscribe(observer: Observer<GraphQLResult>): QuerySubscription;
     result(): Promise<GraphQLResult>;
-}
-export interface QuerySubscription extends Subscription {
```
stubailo: I couldn't figure out how to get `typings` to do the right thing here, so I edited manually to get things to pass. How do we regenerate the types here?

jbaxleyiii:
@stubailo thank you so much for doing this!

stubailo:
@jbaxleyiii I think it's ready to merge as soon as you have time to look it over!

PR #111, opened 2016-07-14
Point TypeScript users to the root typings file

nevir:
AFAIK this should address #77 so that you can just `npm install apollo-react` and have full typings available to TypeScript

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.3.19",
   "description": "React data container for Apollo Client",
   "main": "index.js",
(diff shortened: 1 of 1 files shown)


stubailo:
What happened with this PR?

nevir:
It doesn't actually help like I thought it did; typescript will look for `.d.ts` files alongside any modules you import (as best I can tell) - and you are including those as part of the build process of this module.  

Something greater is afoot, I think, but didn't have time to track it down :(

PR #123, opened 2016-07-29
propagate fetchMore

Slava:
This is a preliminary change, I didn't test it yet.

I might need more help from @stubailo and @jbaxleyiii. Specifically, I am not sure if we need to bind `fetchMore` the same way we bind `refetch`? I don't really understand all the intricacies of the react integration.


Slava:
I read more code and attempted to bind the `fetchMore` function to set a `loadingMore` property too.

jbaxleyiii:
@Slava I should have the new api done by monday. Are you okay with me adding `fetchMore` in there instead of the old api?

Slava:
Yes sure
2016년 7월 30일 (토) 오후 4:51, James Baxley <notifications@github.com>님이 작성:

PR #134, opened 2016-08-11
Ensure that loading state and errors are consistently propagated when calling refetch()

nevir:
This fixes an issue where if there is a server or network error during `refetch()`, the query state reported by `react-apollo` would be stuck `loading: true, errors: undefined` until some other (unrelated) query executes (and bumps the store)


jbaxleyiii:
@nevir thanks so much!

PR #136, opened 2016-08-13
Document Redux form

jbaxleyiii:
This introduces two test cases for usage with redux-form


Xiphe:
For people landing here and searching a reference implementation: this has been moved to https://github.com/apollographql/react-apollo/blob/8af4384ad2e31afdb7360e74575fc87cdcf21155/test/react-web/client/libraries/redux.test.tsx#L93-L169

PR #137, opened 2016-08-13
Add immutable support

jbaxleyiii:
This add support for an immutable tree via

`<ApolloProvider store={myImmutableStore} client={myClient} immutable={true}><App></ApolloProvider>`


src/ApolloProvider.tsx
```diff
@@ -51,6 +53,8 @@ export default class ApolloProvider extends Component<ProviderProps, any> {
 
     if (props.store) {
       this.store = props.store;
+      // support an immutable store alongside apollo store
+      if (props.immutable) props.client.initStore();
```
stubailo: I feel like we wouldn't need this if we just had `client.getStore()` which initialized it for you before returning one? Or is there some other reason we need it?

PR #176, opened 2016-08-29
Fix loading state on variable change with unchanged results

msimulcik:


jbaxleyiii:
@msimulcik could you rebase this?

jbaxleyiii:
@msimulcik I ended up fixing this with #191 that addressed a similar issue. Thanks for the help!

PR #178, opened 2016-08-30
Bugfix: graphql wrapper doesnt work with fragments (SSR)

wizardzloy:
Run into this issue when was toying with React Router + Apollo SSR example, please check the test for the details


jbaxleyiii:
@wizardzloy this is great! I'll get this into the next release!

PR #179, opened 2016-08-30
refactor `calculateFragments`...

wizardzloy:
...to return a value instead of mutating i…ts argument.

It's not something important but it makes the code a bit more readable and allows to get rid of the type coercion [here](https://github.com/wizardzloy/react-apollo/blob/d7ab9ea505bc6c6406bee7bedccf01734e45da4c/src/graphql.tsx#L340)


jbaxleyiii:
@wizardzloy thanks for this!

PR #190, opened 2016-09-05
Fixes loading state true for skipped queries

### LINKED ISSUE #181: loading state true for skipped queries
as reported by `sircharleswatson` in slack (https://apollostack.slack.com/archives/react-apollo/p1472591117000042)

```js
const CURRENT_USER = gql`
  query currentUser {
    currentUser {
      id
      username
      email
    }
  }
`

const withUser = graphql(CURRENT_USER, {
  options: () => ({skip: !userIsAuthenticated()})
})
const AppWithUser = withUser(App)
```

This should be a pretty easy bug if anyone wants to contribute! Essentially the default `true` value for `data.loading` should be set to false if skipping the query.

jbaxleyiii:
Closes #181

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -328,9 +328,22 @@ export default function graphql(
         const { reduxRootKey } = this.client;
         const queryOptions = this.calculateOptions(this.props);
         const fragments = calculateFragments(queryOptions.fragments);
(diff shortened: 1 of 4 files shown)


jbaxleyiii:
These tests will fail until https://github.com/apollostack/apollo-client/pull/618 is published

PR #194, opened 2016-09-06
added link to recompose's compose method

jbaxleyiii:
This makes including multiple `graphql` HOCs on a single component much easier. It uses the [compose](https://github.com/acdlite/recompose#composition) method from recompose.

```js
// component
const Template = (props) => (<div></div>);

// old
import { graphql } from 'react-apollo';

const withPeople = graphql(peopleQuery, { name: 'people' });
const withShips = graphql(shipsQuery, { name: 'ships' });

const ContainerWithData = withPeople(withShips(Template));

// new
import { compose, graphql } from 'react-apollo';

const ContainerWithData = compose(
  graphql(peopleQuery, { name: 'people' }),
  graphql(shipsQuery, { name: 'ships' })
)(Template);

```


package.json
```diff
@@ -101,6 +101,7 @@
     "lodash.flatten": "^4.2.0",
     "lodash.isequal": "^4.1.1",
     "lodash.isobject": "^3.0.2",
-    "object-assign": "^4.0.1"
+    "object-assign": "^4.0.1",
+    "recompose": "^0.20.2"
```
stubailo: Is this a big dependency? Doesn't redux also have a compose function we could use? Or are they different?

sedubois:
Could this `compose` method be documented somewhere? It doesn't seem to be visible in http://dev.apollodata.com/react/.

Issue #147, opened 2016-08-17
Feat: Refactor tests to use Jest


jbaxleyiii:
Since we don't render any markup in any of the client tests, if we refactor to use jest everywhere, the react-web and react-native tests could all be the same. This would vastly increase our react native coverage and how how easy it is to share code between web and native

jbaxleyiii:
thanks to @johnthepink this is done!

PR #201, opened 2016-09-08
Jest Refactor

johnthepink:
This begins to bring all of our tests from mocha to jest, as noted in #147. There is a lot of work yet to be done, but wanted to get this out there early since it's a big change.

All I have really done so far is convert the existing tests so they run using jest. We will probably want to convert some of the tests to use snapshots, etc.

This uses the `scriptPreproccesor` jest parameter to run the test files through `preprocessor.js` before testing, which converts typescript files using `tsc`, and javascript files using `babel-jest`.

The tests actually run fine using chai, but I went through and adjusted for the jest expect syntax. That seems to be the right direction anyways in moving to jest.

I did comment out one `expect` that was failing and marked it `XXX`. It seemed to be failing correctly? I probably messed it up.

I went through a lot of the tests and made sure they weren't false positives, but it would be good to double check this.

I'm sure I missed something below. Thoughts? :)

- [x] set up jest tests using typescript
- [x] convert tests to jest
- [x] ensure tests aren't false positives
- [x] handle warnings
- [x] fix issues (marked XXX)
- [x] convert to snapshots where applicable
- [x] remove mocha and other old deps
- [x] use jest code coverage


stubailo:
Wow! This is huge.

jbaxleyiii:
@johnthepink ❤️ * 1,000,000

stubailo:
We're going to be so trendy after this... except `apollo-client` is still using mocha with chai assert syntax, lol.

I'm going to take my assert syntax to the grave.

johnthepink:
Interesting issue I ran in to. Basically, using enzyme with react-test-renderer in the same file crashes everything.

https://github.com/facebook/react/issues/7386

Unfortunately, it seems like the answer is "wait for React 15.4", which should be out in the next couple weeks. But, for now I've separated things out in to separate files where necessary, and added to notes for clean up after 15.4 is released.

But, we are getting to a good place https://travis-ci.org/apollostack/react-apollo/builds/159778798

johnthepink:
Jest coverage added. Man that was simple. I'll rebase soon.

stubailo:
Wow does this mean you're basically done?

johnthepink:
@stubailo I think so. There are still improvements to be made, but I figured it's better to get this on master sooner rather than later. Jest seems to be logging out errors that we are testing for right now. I do want to see if I can squelch that somehow.

After I rebase this week, I will see if React 15.4 is out, and if so I'll fix the file splitting issue. If not, I'll send a PR to fix that whenever it does get released.

stubailo:
Yeah, we should definitely merge as soon as we can so we don't have a long-running branch.

jbaxleyiii:
@stubailo @johnthepink merging today :tada

jbaxleyiii:
@johnthepink would you mind rebasing?

test/react-web/client/graphql/shared-operations-1.test.tsx
```diff
@@ -22,7 +16,7 @@ import {
 import graphql, { withApollo } from '../../../../src/graphql';
 import { compose } from '../../../../src/';
 
-describe('shared opertations', () => {
+describe('shared operations', () => {
```
jbaxleyiii: ;)

johnthepink: sorry i couldn't help myself in a few places

jbaxleyiii: ❤️

PR #219, opened 2016-09-19
Fix #218

### LINKED ISSUE #218: SSR: this.context is undefined
Suppose you have the following react component, 
and you want to render it using `getDataFromTree`.

```javascript
class MyComponent extends React.Component {
  constructor(props) {
    super(props); // <-- context parameter is ignored
  }

  componentWillMount() {
    this.context.something; // <-- Cannot read property 'something' of undefined error
  }
}
```

`Cannot read property 'something' of undefined` error occurs.

It appears that this code:

https://github.com/apollostack/react-apollo/blob/4271f68d604e7a6f9c23c7a123a0b15573cafd0e/src/server.ts#L48

```javascript
const Component = new ComponentClass(ownProps, context);
```

Doesn't set context if it isn't used in constructor (for some reason it does in the browser).

I didn't check React's source code, but it looks like it sets context even if it isn't used in constructor.

Here's a suggested fix:

```javascript
    const Component = new ComponentClass(ownProps, context);
    try {
      Component.props = ownProps;
+      Component.context = context; // not sure about this
      Component.setState = (newState: any) => {
        Component.state = assign({}, Component.state, newState);
      };
    } catch (e) {} // tslint:disable-line
```

Vanuan:
Fix #218

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -48,6 +48,7 @@ function getQueriesFromTree(
     const Component = new ComponentClass(ownProps, context);
     try {
       Component.props = ownProps;
+      Component.context = context;
       Component.setState = (newState: any) => {
         Component.state = assign({}, Component.state, newState);
       };


jbaxleyiii:
@Vanuan thanks! This will be in the next release!

PR #223, opened 2016-09-21
Created apolloClient utility to retrieve the apollo state in redux.

rewop:
This PR resolves https://github.com/apollostack/react-apollo/issues/221.

It checks if the deprecated `reduxRootKey` property is set in the client and uses it. Otherwise it uses the new `reduxRootSelector` property of the client. 

Also it adds the deprecation warning when `reduxRootKey` is used in the client.


rewop:
@rricard this is the first attempt :)

I had to run typings command to make the types be recognised, but I also see it added a lot of files. Let me know if I have to revert for those files.

rewop:
Another note, I had to fix apollo-client to **0.14.6** to get the version with the problem.

rewop:
Thanks for the review.

About the tests, I am considering the testing the new function should be enough.
Suggestions where I can implement the test for the function?

About the CI, I will look into it, but think I may need help to understand what is wrong.

rricard:
First for the typings, let's try to see if you revert your changes in `typings/` what it does. The, re-npm install. It my still fail for you but may pass in the CI, in this case we should investigate your env!

Second, for the tests, yes just test the new function but try to cover the edge cases as much as possible. Since you created a new file for the function, try creating a new file, we'll move it around if that makes more sense!

Good luck!

rewop:
@rricard Tested and fixed the CI. However I see #226 could be a cleaner alternative than this one.

rricard:
However, what I like here is we add some tests. If you can get rid of the `typings/` files and rebase it, I think we'll have something good.

@jbaxleyiii or @stubailo do you have a preference on this ?

rewop:
@rricard done what you asked. I hope this is helpful.

package.json
```diff
@@ -52,7 +52,7 @@
     "apollo-client": "^0.4.12"
   },
   "devDependencies": {
-    "apollo-client": "0.4.13",
+    "apollo-client": "0.4.16",
```
rewop: I fixed the version here to test it with the faulty version of **apollo-client**, but I think we should change it to the latest version. These changes should back compatible, because `reduxRootKey` has precedence over `reduxRootSelector`.

rricard: yes you should indeed!

rewop:
I am not familiar with typescript either. The problem is that I could not find steps to set up the module for development. So I just run `npm install` `./node_modules/.bin/typing install` (that is why the typings files added) to get started with the project. 

About the CI I have the same error on my local machine when running `npm run compile`. 

I tried also to run `./node_modules/.bin/typing install --production` to update the definitions only for production modules, but the compilation still fails. 

Here I am stuck. I have been trying to look for [info](https://github.com/apollostack/apollo-client/issues/6) to [contribute](https://github.com/apollostack/apollo-client/pull/10) with typescript, but nothing that helped to fix the CI problem. 

Help wanted! :)

rricard:
@rewop don't worry, they plan to migrate from typings to npm-based typing soon. I think we should wait for the opinion of someone else here!

jbaxleyiii:
closed for #226 but thank you for this PR!

Issue #221, opened 2016-09-20
TypeError: Cannot read property 'queries' of undefined


rewop:
I just updated to `react-apollo@0.5.5` and `apollo-client@0.14.6` and my application is now crashing with the error:
```
 TypeError: Cannot read property 'queries' of undefined
```

The error refers to [this line](https://github.com/apollostack/react-apollo/blob/master/src/graphql.tsx#L460).

I do SSR and I integrate with redux. I use `apollo` as name of the key for the apollo state.
Despite the initial state is passed correctly from the server to the frontend app the frontend fails to find the the name of key in which apollo state is stated. 

The line I linked gets the `reduxRootKey` [from the client](https://github.com/apollostack/react-apollo/blob/master/src/graphql.tsx#L448). However `client.reduxRootKey` is undefined.

My app works if I explicitly set the `reduxRootKey` when I create the client:
```javascript
new ApolloClient({
    ...
    reduxRootKey: 'apollo'
});
```

I know that `reduxRootKey` [has been deprecated](https://github.com/apollostack/apollo-client/blob/master/CHANGELOG.md#v0416) in `apollo-client@0.14.6`, so this may be its bug.

However, the function should be updated to use `reduxRootSelector` instead. 

I can make a PR if needed.

rricard:
@rewop that's exactly the issue! If you can produce a PR to support both Key and Selector and add a deprecation warning in react-apollo if Key is the one used, that would be awesome! In the meantime, the client should also be patched (see https://github.com/apollostack/apollo-client/issues/676)

stubailo:
Yep, the newest release has reverted this change, and we're going to work on making it more backwards compatible before shipping it again.

rewop:
Great. I will try to make a PR asap. My first contribution to `react-apollo` so it may take a moment to get into the code.

rricard:
@rewop great to have new contributors! Hopefully it will not be that hard. The codebase is quite easy to get started with :wink:

rewop:
@rricard I gave it a look, and this would be my approach:

I would create a function called `getReduxRootKey(client : ApolloClient, store: ApolloStore) : string` which would be implemented something like follows:

```javascript
function getReduxRootKey(client) {
  // warn if client is using deprecated apollo client reduxRootKey prop
  if (client.reduxRootKey) {
    console.warn(`
      Property 'reduxRootKey' from apollo client deprecated and its usage
      will be removed from the next releases.
[...]

PR #226, opened 2016-09-22
Use getApolloState instead relying on reduxRootKey

sheerun:
I hope I got it well. It may or may not require major bump.

I'd be really glad if you released it asap, as I use my own redux store.

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -213,7 +213,7 @@ export default function graphql(
       // if this query is in the store, don't block execution
       try {
         readQueryFromStore({
-          store: client.store.getState()[client.reduxRootKey].data,
+          store: client.queryManager.getApolloState().data,
           query: opts.query,
           variables: opts.variables,
           fragmentMap: createFragmentMap(opts.fragments),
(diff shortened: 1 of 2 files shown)


sheerun:
This is simpler and more accurate implementation of #223, should resolve #221

sheerun:
Actually it shoudn't require major bump as queryManager.getApolloState() was available in previous versions.

stubailo:
I'm a little uncomfortable because this isn't strictly a "public API" but it definitely works for now! We'll just need to be careful to not remove that function, or rename `queryManager`.

sheerun:
CI fails for non-related reasons.. I guess if you figure out public API for getting apollo state, you can change it. At least it's less hacky than manually fetching it by `reduxRootSelector` (that can be both string or function..)

stubailo:
Yeah I think this is a great solution right now, let's do it!

sheerun:
@stubailo Any chance to release it soon? :)

jbaxleyiii:
@sheerun I'll get this out today!

Issue #209, opened 2016-09-12
support custom store getter


jbaxleyiii:
https://github.com/apollostack/apollo-client/pull/631

jbaxleyiii:
^^ has landed 💯 so this should be in the next release

jbaxleyiii:
Fixed in `0.5.7`

PR #217, opened 2016-09-18
update to new typescript and remove typings

jbaxleyiii:
Inspired by https://github.com/apollostack/apollo-client/pull/624

PR #229, opened 2016-09-24
Upgrade to TypeScript 2

wizardzloy:
As a part of the implementation of #209 we need to upgrade the version of `typescript` because the newest version of `apollo-client` uses it


wizardzloy:
closing this, as a duplicate of #217

PR #234, opened 2016-09-27
remove nested imports for apollo-client

### LINKED ISSUE #172: Avoid importing from sub-directories of apollo client
Hey, right now it's pretty hard to use react-apollo or apollo-client with `npm link` because this package reaches into sub-directories of the published package.

I'd love to be able to eliminate that completely.

Can we just make a list of everything that needs to be exported from `apollo-client` to avoid the internal imports?

jbaxleyiii:
closes #172

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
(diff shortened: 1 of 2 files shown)


stubailo:
Wow, great!

PR #261, opened 2016-10-11
Use `fail(err)` rather than `done(err)`

tmeasday:
cc @jbaxleyiii


tmeasday:
I tested that it actually does fail btw

PR #291, opened 2016-10-27
Fix issue with usage in TypeScript projects caused by 'compose' re-export

### LINKED ISSUE #279: Cannot find module 'recompose/compose'.
## Steps to Reproduce

Just import `react-apollo`.

### Buggy Behavior

```
[0] /home/zhenghu/coding/qsc/qsc-main-react-typescript/node_modules/react-apollo/index.d.ts
[0] (3,21): error TS2307: Cannot find module 'recompose/compose'.
```

### Expected Behavior

No error.

### Version
- apollo-client@0.4.21
- react-apollo@0.5.11

vlasenko:
Fixes #279 

As suggested by @migueloller `compose` is re-exported from `redux` instead of `recompose`, since implementation is completely the same, that's why we can drop depending on `recompose` completely.

TODO:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import ApolloProvider from './ApolloProvider';
 import graphql, { withApollo } from './graphql';
 
-// expose easy way to join queries from recompose
-import compose from 'recompose/compose';
+// expose easy way to join queries from redux
+import { compose } from 'redux';
 
 export { ApolloProvider, graphql, withApollo, compose };
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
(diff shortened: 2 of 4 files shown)


migueloller:
Would it make sense to rebase so that the commit that removes the `recompose` dependency also imports from `redux`? This way the commit makes a non-breaking atomic change and can easily be reverted if necessary.

src/index.ts
```diff
@@ -2,6 +2,6 @@ import ApolloProvider from './ApolloProvider';
 import graphql, { withApollo } from './graphql';
 
 // expose easy way to join queries from recompose
```
vlasenko: Thank you! Done

vlasenko:
Sure. I will try now. I never did this before.... Any pointers appreciated

vlasenko:
Okay, rebased everything into single commit

jbaxleyiii:
@vlasenko this is great, my only concern is with importing all of redux. This probably isn't an issue since apollo-client imports redux. Technically this integration doesn't actually rely on redux at all. 

That being said, I'm great with this for now! Thanks for the PR

Issue #292, opened 2016-10-27
`options.forceFetch` seems to break SSR


thebigredgeek:
## Steps to Reproduce

Describe how to reproduce this issue.

-  Create a component, decorate with `graphql` from `react-apollo`.  Write a simple query, and in the options of the query set `forceFetch` to `true`.  Render isomorphically using `getDataFromTree` from `react-apollo/server` combined with  `ReactDOMServer.renderToString` against root component, OR `renderToStringWithData` from `react-apollo/server`.  Log out the initial state server side and see the hydrated state tree from the component's query.  Log out the markup generated from the render server side and see lack of proper rendering.  Log out `props` in the component on the server and see that the `...fields` entries are missing from `props.data`.

##### renderer.js

```js
export default async function render (req, res, next) {
  try {
    const { renderContext, routes } = routesFactory();

    const { redirectLocation, renderProps } = await matchAsync({ routes: routes, location: req.originalUrl });

    if (redirectLocation) {
      return res.redirect(redirectLocation.pathname + redirectLocation.search);
    }

    const client = new ApolloClient({
      ssrMode: true,
      networkInterface: createNetworkInterface({
        uri: `${API_FQDN}/graphql`,
        credentials: 'same-origin',
        headers: req.headers
      })
    });

    const root = (
      <ApolloProvider client={client}>
        <RouterContext {...renderProps} />
      </ApolloProvider>
    );

    const initialState = (await getDataFromTree(root)).store.getState();

    const content = ReactDOMServer.renderToString(root);

    //const { markup: content, initialState } = await renderToStringWithData(root);

    //console.log(content);

    //console.log(JSON.stringify(initialState, null, 2));

    const html = <Html content={content} state={initialState} style={__PRODUCTION__ ? renderContext.getCssString() : ''} {...initialState.pageInformation}/>

    const markup = `<!doctype html>\n${ReactDOMServer.renderToStaticMarkup(html)}`;

    setTimeout(() => res.status(200).send(markup), 2500);
  } catch (e) {
    console.trace(e);
    next(e);
  }
}
```

##### usersTable.js

```javascript
import React from 'react';
import withStyles from 'isomorph-style-loader/lib/withStyles';
import { graphql } from 'react-apollo';
import PureComponent from 'components/common/pure';
import { propTypes, defaultProps } from 'lib/decorators/generic';
import gql from 'graphql-tag';

import UserRow from 'components/userRow';

import style from 'components/usersTable/index.css';

[...]

PR #293, opened 2016-10-28
Fix forceFetch on SSR

jbaxleyiii:
Fix for #292 

TODO:


tmeasday:
Interesting. We added a `ssrMode: true` to AC which simply ignores `forceFetch` to get around this problem originally. 

I don't recall if we had an idea at the time of use cases where people might validly want to force fetch in SSR (and thus not set the option). Probably not huh? If so, we should remove `ssrMode` as an option (and make sure the angular SSR works the same way).

PR #301, opened 2016-11-02
Better packaging

vlasenko:
Provide compiled sources from 'lib' for node modules. Provide all API exports from index.js.

This is breaking change. 

As a side effect this change fixes file paths in Stack Traces when react-apollo throws exceptions.

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -1,5 +1,10 @@
 /// <reference path="./node_modules/typed-graphql/graphql.d.ts" />
 
+declare namespace NodeJS {
+  interface Process {
+    browser: boolean;
+  }
+}
 
 /*
   LODASH
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,17 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
+#### Breaking
+```js
+// old
+import { getDataFromTree, renderToStringWithData } from 'react-apollo/server'
+
+// new
+import { getDataFromTree, renderToStringWithData } from 'react-apollo'
+```
+
+- Feature: Better packaging [PR #301](https://github.com/apollostack/react-apollo/pull/301)
+
 ### v0.5.15
 - Feature: Added test utilities and examples to library.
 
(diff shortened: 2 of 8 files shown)


typings.d.ts
```diff
@@ -1,5 +1,10 @@
 /// <reference path="./node_modules/typed-graphql/graphql.d.ts" />
 
+declare namespace NodeJS {
```
tmeasday: Should we use `@types/node` for this?

jbaxleyiii:
@vlasenko I appreciate the effort to make imports easier, however I'm really concerned about this ballooning the file size of the import.

stubailo:
Yeah, I don't think loading all of the SSR code in the client bundle is a good idea. Perhaps the real solution is to make it a separate package somehow? But that might be too much overhead.

vlasenko:
@stubailo @tmeasday The thing is the SSR code IS NOT loaded into the client bundle with this implementation. It is excluded from client bundle by combination of bpb and unreachable-branch-transform browserify plugins

typings.d.ts
vlasenko: Nope. By this declaration we are signalling TypeScript compiler that we have `browser` field in the `process`.

tmeasday: Oh, OK; is this a browserify thing? I guess I wonder if there are some defined typings for it. Probably not a big deal to include this though.

vlasenko:
In this implementation SSR code is excluded from client bundle by utilizing dead-code elimination with browserify plugins. There is another route - to use conditional compilation browserify plugin: https://github.com/janvanhelvoort/browserify-conditionalify I'm not sure which route  is better. With conditional compilation using commonjs requires in index.js and adding `process.browser` typings will not be needed...

Yet another route is, of course, to move SSR into separate package.

vlasenko:
I have submitted alternative implementation, which utilizes `package.json` `browser` field. It has the same effect as this one, but is much cleaner IMO. So I'm closing this one.

Link to new  PR:
https://github.com/apollostack/react-apollo/pull/306

PR #305, opened 2016-11-03
Fix issues w/ react@15.4 & jest-react-native

tmeasday:
Fix as suggested by @vlasenko on Slack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,9 @@
   "jest": {
     "preset": "jest-react-native",
     "testEnvironment": "jsdom",
-    "scriptPreprocessor": "preprocessor.js",
+    "transform": {
+      ".*": "preprocessor.js"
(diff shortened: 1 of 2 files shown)


tmeasday:
Hmm. The tests seem to be failing intermittently now...

vlasenko:
After experimenting with different React versions, I have a proposal to keep React packages on 15.3.2 version, BUT react-test-renderer on 15.4.0-rc.4.

React@15.4.x does not have `react/lib/ReactNative` module and `react-native` 0.36.x is importing it, so we have a problem with this combination.

React-test-renderer@15.3.2 triggers issue:
https://github.com/facebook/react/issues/7386#issuecomment-253935567

This issue is fixed in react-test-renderer@15.4.0-rc.3

PR #308, opened 2016-11-04
Allow renderToStringWithData to run without queries

rattrayalex:
TODO:

Fixes https://github.com/apollostack/react-apollo/issues/304

What, if any, of the TODO's above need to occur here?


tmeasday:
@rattrayalex I suppose this is no longer relevant if we aren't returning state from `renderToStringWithData` any more.

rattrayalex:
Yup! Thanks!

PR #313, opened 2016-11-07
Refactor ssr

tmeasday:
Refactored SSR to use a slightly more principled `walkTree` method.

Tests pass, some points of interest:

1. I got rid of returning the `context` from the SSR functions. This seemed kind of undefined (what was it? The lowest context we saw in the tree? Should we have been just picking the first `ApolloProvider` child context we saw?). All in all it seemed a bit hacky, and the user will have access to an `ApolloClient` anyway, so can't they just get the store themselves?

 - this will require a docs change.

2. I am still assigning `instance.setState`. This seems to *not* throw an exception. How does this fit your understanding @rewop? I was surprised by this but we need to do something like this to allow people to call `setState` in `componentWillMount` (or pursue a different method of rendering[1]).

3. @jbaxleyiii I wonder if rather than passing `{ element, context }` around (and calling `element.type.fetchData`, a static method), we could not just pass the `instance` around, and call `instance.fetchData`? It would make the code a little simpler in `server.js`, but also it would make `graphql.js` a bit simpler too. Is there a reason to *not* instanciate the element on the server before running the query?

[1] I feel like ultimately we should be trying to use React's rendering infrastructure, as although this is no doubt quicker, it feels really brittle and wouldn't be surprised if there's lots of things you can do that break it.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -48,6 +48,9 @@ export function walkTree(
     //   https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L66
     if (Component.prototype && Component.prototype.isReactComponent) {
       const instance = new Component(props, context);
+      // In case the user doesn't pass these to super in the constructor
+      instance.props = instance.props || props;
+      instance.context = instance.context || context;
 
       // Override setState to just change the state, not queue up an update.
       //   (we can't do the default React thing as we aren't mounted "properly"
@@ -63,7 +66,6 @@ export function walkTree(
         instance.componentWillMount();
       }
 
-
       if (instance.getChildContext) {
         childContext = assign({}, context, instance.getChildContext());
       }
@@ -73,12 +75,15 @@ export function walkTree(
       child = Component(props, context);
     }
 
-    walkTree(child, childContext, visitor);
-
+    if (child) {
+      walkTree(child, childContext, visitor);
+    }
   } else { // a basic string or dom element, just get children
     if (element.props && element.props.children) {
       Children.forEach(element.props.children, (child: any) => {
-        walkTree(child, context, visitor);
+        if (child) {
+          walkTree(child, context, visitor);
+        }
       });
     }
   }
--- a/src/server.ts
+++ b/src/server.ts
@@ -3,7 +3,6 @@ import { Children } from 'react';
 import * as ReactDOM from 'react-dom/server';
 import ApolloClient from 'apollo-client';
 import assign = require('object-assign');
-import flatten = require('lodash.flatten');
 
 
 declare interface Context {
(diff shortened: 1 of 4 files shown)


tmeasday:
I believe this will fix https://github.com/apollostack/react-apollo/pull/237 and https://github.com/apollostack/react-apollo/issues/250

rattrayalex:
> Is there a reason to not instantiate the element on the server before running the query?

I'm just a bystander with little context, but I imagine performance would be one reason – especially if users defined any heavy `constructor` methods, but even with standard object/class instantiation, that could add up with an entire react component tree.

tmeasday:
> I'm just a bystander with little context, but I imagine performance would be one reason – especially if users defined any heavy constructor methods, but even with standard object/class instantiation, that could add up with an entire react component tree.

Sure, but we have to create the instances at some point to render the subtree. Also in this case we are talking about `graphql` instances anyway, and we know a lot about their behaviour!

src/server.ts
```diff
@@ -13,107 +12,127 @@ declare interface Context {
...
+    // Are we are a react class?
+    //   https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L66
+    if (Component.prototype && Component.prototype.isReactComponent) {
+      const instance = new Component(props, context);
+
+      // Override setState to just change the state, not queue up an update.
+      //   (we can't do the default React thing as we aren't mounted "properly"
+      //   however, we don't need to re-render as well only support setState in
+      //   componentWillMount, which happens *before* render).
+      instance.setState = (newState) => {
```
tmeasday: We don't call set state, but it's totally possible that any component we are "rendering" will rely on calling in it `componentWillMount`. I think there were some issues about this (and that's why there's a test for it).

src/server.ts
```diff
@@ -13,107 +12,127 @@ declare interface Context {
...
   return getDataFromTree(component)
-    .then(({ client }) => {
-      let markup = ReactDOM.renderToString(component);
-      let apolloState = client.queryManager.getApolloState();
-
-      for (let queryId in apolloState.queries) {
-        let fieldsToNotShip = ['minimizedQuery', 'minimizedQueryString'];
-        for (let field of fieldsToNotShip) delete apolloState.queries[queryId][field];
-      }
+    .then(() => ReactDOM.renderToString(component));
```
rattrayalex: Wow, what a massive cleanup! 

Are you planning to keep `initialState` as a return value?

src/server.ts
```diff
@@ -13,107 +12,127 @@ declare interface Context {
...
+    const skipRoot = !fetchRoot && (element === rootElement);
+    if (typeof Component.fetchData === 'function' && !skipRoot) {
+      const props = assign({}, Component.defaultProps, element.props);
+      const query = Component.fetchData(props, context);
+      if (query) {
+        queries.push({ query, element, context });
+
+        // Tell walkTree to not recurse inside this component;  we will
+        // wait for the query to execute before attempting it.
+        return false;
```
tmeasday: True, but it will eventually get rendered when the query is ready. The assumption is that you have something like

```js
@graphql(...)
const X = ({ loading }) => {
  if (loading) {
    // A. there's no nested calls to `graphql` inside this render tree
    return <Loading/>;
  } else {
    // B. there could be anything in here
  }
}
```

We don't bother rendering the A. subtree as we are going to await the query setup by `graphql` and then render `X` with the results (and `loading: false`). There are probably pathological cases you can think of where this is inefficient, but I think it seems a reasonable assumption and is "eventually consistent" in any case.

src/server.ts
tmeasday: No, see point 1. in the original PR comment; it was a bit of a hack IMO

src/server.ts
rattrayalex: Hmm, is `Component.fetchData` recursive then? What if the `B` subtree has a `@graphql(...)` of its own? 

Thanks very much for taking the time to explain!

src/server.ts
rattrayalex: Ahhh, I see. That makes sense, I believe. 

I imagine another option might be to run `Component.flushState()` or whatever it's called, just before `Component.componentWillMount()` ? But I don't necessarily see that as more elegant. 

Again, thanks for taking the time to explain!

src/server.ts
rattrayalex: Ahh, I see... 

> the user will have access to an ApolloClient anyway, so can't they just get the store themselves

Yes, that sounds much better anyway 😄

src/server.ts
tmeasday: `getQueriesFromTree` is 


> Thanks very much for taking the time to explain!

No problem! Happy to talk about it while I still remember ;)

rattrayalex: Ahhh, of course. Somehow I missed the fact that there are two separate recursions happening: one to find all top-level `graphql` nodes, and one to find/execute their child queries once the first have completed... although, of course, you had clearly stated that. 

> We have to wait for X's query though, as any sub-queries may rely on the results.

Of course. Presumably in the future some introspection could check whether this is true, and batch the queries together if not. 

Hah, well I'm glad to get your thoughts on the public record while we can then! 😄

rattrayalex:
This PR seems like a major step forward in the simplicity and readability of this functionality!

jbaxleyiii:
@tmeasday awesome work! I'm going to try and review this today!

rewop:
@tmeasday sorry I was away on holiday. This PR is great indeed. As for my issue the problems was  setting **props** and **context** to the instance of the component because the properties could not be modified. **setState** was not an issue for me. 

So this should fix my scenario.

rewop:
> I feel like ultimately we should be trying to use React's rendering infrastructure, as although this is no doubt quicker, it feels really brittle and wouldn't be surprised if there's lots of things you can do that break it.

Totally agree with this.

rattrayalex:
> I feel like ultimately we should be trying to use React's rendering infrastructure, as although this is no doubt quicker, it feels really brittle and wouldn't be surprised if there's lots of things you can do that break it.

Speaking as a (potential) user of this library, perf is likely to be a top consideration in whether to adopt Apollo for SSR. I don't know very much about the tradeoffs at hand, so I wouldn't take my "opinion" on the matter too seriously. But If tests can be written to combat the brittleness and detect breakages, that'd be much preferred to a slower-but-more-"correct" approach. 

Writing some benchmarks might be helpful in making that call as well (though again, I don't have the context necessary to tell how helpful that'd be).

jbaxleyiii:
@tmeasday this looks awesome. I'm great to merge it if you are! (Feel free to merge and release if you want!)

tmeasday:
I think it's a step forward in any case, so I am going to merge it. I guess we'll hear about problems as it goes live ;)

tmeasday:
@jbaxleyiii I wonder what your thoughts are on 

> I got rid of returning the context from the SSR functions. This seemed kind of undefined (what was it? The lowest context we saw in the tree? Should we have been just picking the first ApolloProvider child context we saw?). All in all it seemed a bit hacky, and the user will have access to an ApolloClient anyway, so can't they just get the store themselves?

jbaxleyiii:
@tmeasday it was a hack 👍 I think its great

jbaxleyiii:
Also, thank you SO much for the fantastic work with the new SSR renderer!

tmeasday:
Ok, I'll update the docs etc to reflect the change.

PR #317, opened 2016-11-08
Fix issues with context and undefined children

bkniffler:


src/server.ts
```diff
@@ -48,6 +48,8 @@ export function walkTree(
     //   https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L66
     if (Component.prototype && Component.prototype.isReactComponent) {
       const instance = new Component(props, context);
+      instance.props = props;
+      instance.context = context;
```
tmeasday: Why do we need these? I thought `new Component` did that for us?

bkniffler: These three constructors are valid for react:
```
constructor() {
super();
}

constructor(props) {
super(props);
}

constructor(props, context) {
super(props, context);
}
```
But only the last one is covered by `new Component(props, context)`. At least thats what I suppose.

bkniffler:
I can't do it immediately, got some work to catch up. But there is a lot of situations where a child can be null, a stupid example: `<MyComponent>{user.isAdmin ? <AdminPanel /> : null}{user.isSuperAdmin ? <SuperAdminPanel /> : null}</MyComponent>`.

tmeasday:
Thanks @bkniffler , see https://github.com/apollostack/react-apollo/commit/800ce408e79f615299cbdaeb59b5aca3951b2965

Issue #322, opened 2016-11-11
Add networkStatus prop to connected components


cannoneyed:
## Steps to Reproduce

When fetching more data, using a pattern identical to that proposed in the docs, the `loading` prop that is passed down to the wrapped component is `false` after `fetchMore` is invoked. 

```javascript
const Query = gql`
query paginatedQuery ($offset: Int) {
    paginatedItems (offset: $offset) {
        ... item
    }
}
`
@graphql(Query, {
    props:  ({ data }) => {
        const { fetchMore } = data
        const queryResult = data.paginatedItems

        data.loadNextPage = () => fetchMore({
            variables: {
                offset: queryResult.length
            },
            updateQuery: (prev, { fetchMoreResult }) => {
                if (!fetchMoreResult.data) {
                    return prev
                }

                return Object.assign({}, prev, {
                    [queryName]: [
                        ...prev[queryName],
                        ...fetchMoreResult.data.paginatedItems
                    ]
                })
            }
        })

        return { data }
    }
}
class View extends Component {
    render() {
        const { data: { paginatedItems, loading, loadNextPage } } = this.props
        // Loading is false after loadNextPage is invoked
    }
}
```

### Buggy Behavior

It seems the `loading` prop is only ever tied to the initial load of data. For simple components that don't need to update or paginate, this is fine. However, the workaround for this (managing refetch / loadMore status with local component state) is not ideal.  

### Expected Behavior

The following PR https://github.com/apollostack/apollo-client/issues/707 in apollo client exposes a `networkStatus` object for handling the status of non-initial-load network requests, such as refetch, fetchMore, and subscriptions.

It would be very nice to have the `networkStatus` object exposed as a prop to wrapped components (perhaps if specified as an option), so that paginated / more complex query operations loading/error state can be managed by the react-apollo wrapper rather than using local ad-hoc local component state.

### Version
- apollo-client@0.5.0
- react-apollo@0.5.16

helfer:
@cannoneyed FYI, the `fetchingMore` network status isn't implemented in apollo-client yet. All others should be.

wizardzloy:
I would like to implement this functionality since my app depends on it. @jbaxleyiii  @cannoneyed @helfer Is anybody working on it now?

cannoneyed:
@wizardzloy I'm starting to look into this now.

cannoneyed:
[...]

PR #327, opened 2016-11-15
Feature: Add networkStatus to props of the connected components

wizardzloy:
Implements #322


wizardzloy:
When [this issue](https://github.com/apollostack/apollo-client/issues/851) will be implemented in apollo-client `networkStatus = 3 (fetchMore)` should also be available on the connected component if `fetchMore()` is called

tmeasday:
Looks great! Thanks @wizardzloy

Issue #356, opened 2016-12-05
This uses deprecated `createFragment`


SimenB:
See https://github.com/apollostack/apollo-client/pull/984#discussion_r90794327

## Steps to Reproduce

Use `graphql` export

### Buggy Behavior

I get a deprecation warning from react-apollo

### Expected Behavior

No deprecation warning

### Version
- apollo-client@0.5.11
- react-apollo@0.6.0

tmeasday:
Thanks, @SimenB, fixed in https://github.com/apollostack/react-apollo/pull/357, trying to figure out why an unrelated test is failing.

tmeasday:
I think this was fixed in v0.7.1

SimenB:
Yup, no more warning. Thanks!

PR #357, opened 2016-12-06
Removed use of `createFragment` and bumped AC version

tmeasday:
For #356.

@jbaxleyiii any idea why this test is failing? It's the "real" SSR test; it seems to be throwing an error out of `whatwg-fetch`, and started when I updated to latest AC. A couple of things:

1. Latest AC bumped to `whatwg-fetch@2.0.0`.

2. This test shouldn't use `whatwg-fetch` at all, given it's node code, it should use `isomorphic-fetch`. I wonder if Jest is somehow doing it with JSDOM stuff (that's in the stack trace too). I'm out of my depth with this stuff.


tmeasday:
Also, you might want to consider the PR too ;)

tmeasday:
Oh, weird, the tests passed, just the `tsc` failed. The "SSR › `renderToStringWithData` › should work on a non trivial example" test fails here though

jbaxleyiii:
@tmeasday sadly I'll have to get to this tomorrow :( I've got an update to out giving platform I need to babysit out tonight. I'll work on cleaning up some of the test output as well to make it clearer when I work on it tomorrow!

Also, thank you SO much for helping to support this project. I've found my able time reduced recently due to some changes at work and you helping maintain it has been incredible. So thanks 😻

tmeasday:
I fixed most of the problems but the test still fails locally and there are some weird other problems. Would appreciate you taking a look.

SimenB:
Any news here? It messes up my pristine output 😄 Getting errors I can do nothing about sucks, and chrome keeps telling me I get warnings in the console...

jbaxleyiii:
@tmeasday all fixed! I'm going to merge this through

PR #372, opened 2016-12-16
add react-dom as peer depency

koenpunt:
TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
+- Add `react-dom` to `peerDependencies` because since React 15.4 it is no longer "secretly" included.  
+  _(ref: https://github.com/facebook/react/releases/tag/v15.4.0)_
+
 ### v0.7.1
 
 #### Breaking
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
   "peerDependencies": {
     "apollo-client": "^0.4.21 || ^0.5.1",
     "react": "0.14.x || 15.* || ^15.0.0",
+    "react-dom": "0.14.x || 15.* || ^15.0.0",
     "redux": "^2.0.0 || ^3.0.0"
   },
   "devDependencies": {


helfer:
@koenpunt if I'm not mistaken people could use react-apollo anywhere they like, not only in the browser, so I'm not sure why we should add react-dom as a peer dependency?

koenpunt:
> since React 15.4 react-dom is no longer "secretly" included.
>
> ref: https://github.com/facebook/react/releases/tag/v15.4.0

And since `react-dom/server` is required from [`server.ts`](https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L3), this causes errors.

helfer:
Ah, I did not see that, thanks!

Can you sign the CLA?

migueloller:
What will this PR mean for using `react-apollo` in React Native? Will you need to install `react-dom` as a dependency to prevent NPM from complaining?

koenpunt:
> What will this PR mean for using react-apollo in React Native?

I am using `react-apollo` with React Native, and building fails without adding `react-dom` as a dependency due to the before mentioned [require](https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L3)

migueloller:
Having to install `react-dom` as a dependency when using React Native isn't good though.

helfer:
@migueloller don't you effectively already have to do that now if you use SSR? Or is SSR not a thing in react-native?

Since we only use `renderToString`, we could potentially use a pared down react-dom package if such a thing exists.

migueloller:
@helfer,

SSR is not a thing in React Native. The React team keeps the core React library separate from the different renderers (native, DOM/SSR, ART, etc.) so it might make sense to separate DOM/SSR logic from Apollo into a different package.

Perhaps there could be a pared down `renderToString` out there somewhere that doesn't include all of the shadow DOM diffing logic that `ReactDOM.render` needs and simply traverses the component tree and renders it. I haven't come across it though.

jbaxleyiii:
@migueloller we have discussed moving the main logic of this renderer out of this package. For now, I think we should merge this. Up until 15.4 RN builds got react-dom anyway.

I think in a future release we should remove the react-dom integration for a new package or just remove the render to string part. This would also allow this library to potentially be used with react-like renderers (i.e inferno and preact)

jbaxleyiii:
@koenpunt could you update the changelog?

koenpunt:
@jbaxleyiii done.

jbaxleyiii:
@koenpunt sorry this took so long to merge!

Issue #374, opened 2016-12-20
Stale variables when combining `pollInterval`, `skip`, `forceFetch`


jbinto:
This seems similar to https://github.com/apollostack/react-apollo/issues/358 (but possibly unrelated).

Here is my situation:

* When I click a button, I issue a mutation. I get an ID back (e.g. id=1).
* A polling query starts (since the value of a skip became false). options is called with ownProps, and that has the ID from the mutation (e.g. id=1).
* Polling begins, and network inspector shows the query is using the correct ID (e.g. id=1)
* After the polling query sees changes in some state, we flip skip again and polling stops.
* All is well at this point.

But, if I click the button again, this happens:

* The mutation is issued. I get a new ID back (e.g. id=2)
* A polling query starts (since the value of a skip became false). options is called with ownProps, and that has the ID from the mutation (e.g. id=2).
* Polling begins, and network inspector shows the query is using the WRONG ID, from the previous variables. (e.g. id=1)
* This breaks other things in my app because it was not expecting a result for id=1 but for id=2.

## Steps to Reproduce

I've isolated this in a reproduction case:

```
git clone https://github.com/jbinto/apollo_polling_bug/
cd apollo_polling_bug
yarn
npm start
```

To keep things simple I don't do a mutation, just controlling props with `setTimeout` is sufficient to reproduce. I used a mock network interface that just echoes back an argument.

Relevant files:

https://github.com/jbinto/apollo_polling_bug/blob/master/src/index.js
https://github.com/jbinto/apollo_polling_bug/blob/master/src/Foo.js

### Buggy Behavior

Somehow, `apollo-client` or `react-apollo` is holding on to old variables and issuing queries based on those, despite the fact the props of the `graphql` HOC have changed. 

### Expected Behavior

I should be able to combine `skip`, `forceFetch`, and `pollInterval` as such:

* `forceFetch` and `pollInterval` are fixed as `true`, and `1000` respectively. `skip` begins as `true`.
* Set `skip` to false, by re-rendering with new props (including an `id` prop, which is used in `options.variables`)
* Let it poll for a bit
* Set `skip` to true, polling stops
* Let it idle for a bit
* Set `skip` to false, and send a new `id` prop
* Polling starts again, but...
* It should use the new `id` prop I provided to `options.variables`, not the old one

### Version
- apollo-client@0.5.23
- react-apollo@0.7.1

tmeasday:
Thank you @jbinto for the excellent reproduction.

[...]

PR #376, opened 2016-12-21
Ensure that we always change options, even when unsubbed

tmeasday:
For #374. If we go from skip -> no-skip w/ new options, we won't have a subscription, but we will have an observable that we need to update with the new options.


test/react-web/client/graphql/queries.test.tsx
```diff
@@ -514,6 +514,61 @@ describe('queries', () => {
...
   });
 
+  it('allows you to skip then unskip a query with new options (top-level syntax)', (done) => {
+    const query = gql`query people($first: Int) { allPeople(first: $first) { people { name } } }`;
+    const dataOne = { allPeople: { people: [ { name: 'Luke Skywalker' } ] } };
+    const dataTwo = { allPeople: { people: [ { name: 'Leia Skywalker' } ] } };
+    const networkInterface = mockNetworkInterface(
+      { request: { query, variables: { first: 1 } }, result: { data: dataOne } },
+      { request: { query, variables: { first: 2 } }, result: { data: dataTwo } },
+      { request: { query, variables: { first: 2 } }, result: { data: dataTwo } },
```
tmeasday: I'm not sure why I need to include the same request twice here. I suspect it may be a bug in AC but I need to dig deeper.

@helfer anything immediately come to mind? The behaviour is:

1. ObservableQuery with `first: 1`, subscribed.
2. Subscription stopped
3. Variables set to `first: 2`
4. Re-subscribed.

For some reason this fires the network request twice. Any ideas?

tmeasday:
Hold on merging this due to further investigation needed: https://github.com/apollostack/react-apollo/pull/376#discussion_r93367673

test/react-web/client/graphql/queries.test.tsx
tmeasday: > my mocking of the network interface may be to blame

I doubt it! (Although we should find a way to use the MNI from AC.

> so I added the extra test

Oh, what test are you referring to? The above is my own test that I just wrote.

jbaxleyiii: There is another test that has an extra one, let me find it

jbaxleyiii: https://github.com/apollostack/react-apollo/blob/master/test/react-web/client/graphql/queries.test.tsx#L967-L971

tmeasday:
Hmm, this is more complicated than I feared.

The problem referenced [here](https://github.com/apollostack/react-apollo/pull/376#discussion_r93367673) is due to the fact that:

1. We unsubscribe (when skipping).
2. We change the variables, which triggers a `fetchQuery` [here](https://github.com/apollostack/apollo-client/blob/master/src/core/ObservableQuery.ts#L301)
3. Immediately, we re-subscribe, which triggers the `ObservableQuery` to re-fetch [here](https://github.com/apollostack/apollo-client/blob/master/src/core/ObservableQuery.ts#L301)

Because 2+3 happen instantaneously in sequence, both queries hit the server. I'm not sure what we could do differently; if we reversed the order I would expect the same problem, except that the re-subscription would trigger the old query.

I guess this behaviour is better than the broken behavior it is fixing, so I'll merge it, but I wonder if there's any thing we can do about it. Thoughts @helfer, @jbaxleyiii?

helfer:
@tmeasday sorry for the late response on this one. I'm guessing our discussion around not fetching when setVariables is called but there's no subscriber is because of this very issue?

tmeasday:
@helfer, yes, I should have added a note here that I opened https://github.com/apollostack/react-apollo/issues/381 to deal with this.

PR #389, opened 2017-01-03
add shouldResubscribe option

gilesbradshaw:
A should resubscribe option makes subscription queries optionally resubscribe when the props change..

I use it like this:

````
const NODE_SUBSCRIPTION_QUERY = gql`
  subscription  - blah blah ....

const NODE_QUERY = gql`
  query  - blah blah ....

const _Node = (props) => <Blah />

const options = {
  shouldResubscribe: (props, nextProps) => {
    // if the id property changes then we need to resubscribe..
    return nextProps.id !== props.id
  }
}
const Node = compose(
  graphql(NODE_SUBSCRIPTION_QUERY, options),
  graphql(NODE_QUERY)
)(_Node)
````

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
+ [PR #389](https://github.com/apollostack/react-apollo/pull/389) added a shouldResubscribe option to allow subscriptions to automatically resubscribe when props change.
+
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@gilesbradshaw nice! can you add this to the docs site as well?

gilesbradshaw:
sure! apollostack/react-docs#138

PR #413, opened 2017-01-12
Add an `alias` option to `graphql` container

SachaG:
As explained in https://github.com/apollostack/react-apollo/issues/354, your React component tree can get quite confusing when you have many nested `Apollo(...)` containers. This PR adds an `alias` option to the `graphql` HoC factory function:

```
options(ownProps) {
        return {
          alias: 'withCurrentUser',
        };
      },
```

Before:

After: 

TODO:

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -37,6 +37,7 @@ export declare interface MutationOptions {
   optimisticResponse?: Object;
   updateQueries?: MutationQueryReducersMap;
   forceFetch?: boolean;
+  alias?: string;
 }
 
 export declare interface QueryOptions {
@@ -47,6 +48,7 @@ export declare interface QueryOptions {
   noFetch?: boolean;
   pollInterval?: number;
   fragments?: FragmentDefinitionNode[] | FragmentDefinitionNode[][];
+  alias?: string;
   // deprecated
   skip?: boolean;
 }
@@ -156,7 +158,7 @@ export default function graphql(
   const version = nextVersion++;
   return function wrapWithApolloComponent(WrappedComponent) {
 
-    const graphQLDisplayName = `Apollo(${getDisplayName(WrappedComponent)})`;
+    const graphQLDisplayName = `${mapPropsToOptions({}).alias || `Apollo`}(${getDisplayName(WrappedComponent)})`;
 
     class GraphQL extends Component<any, any> {
       static displayName = graphQLDisplayName;
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -37,7 +37,6 @@ export declare interface MutationOptions {
   optimisticResponse?: Object;
   updateQueries?: MutationQueryReducersMap;
   forceFetch?: boolean;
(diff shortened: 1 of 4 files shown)


SachaG:
Also I wasn't sure where/how to write tests, so I thought I'd get feedback on the PR itself first and improve it later.

tmeasday:
Can you write a test that just renders a aliased wrapped component and then checks it's display name?

src/graphql.tsx
```diff
@@ -37,6 +37,7 @@ export declare interface MutationOptions {
   optimisticResponse?: Object;
   updateQueries?: MutationQueryReducersMap;
   forceFetch?: boolean;
+  alias?: string;
```
tmeasday: I think it should be an `OperationOption`

src/graphql.tsx
```diff
@@ -156,7 +158,7 @@ export default function graphql(
   const version = nextVersion++;
   return function wrapWithApolloComponent(WrappedComponent) {
 
-    const graphQLDisplayName = `Apollo(${getDisplayName(WrappedComponent)})`;
+    const graphQLDisplayName = `${mapPropsToOptions({}).alias || `Apollo`}(${getDisplayName(WrappedComponent)})`;
```
tmeasday: If you make it an `OperationOption` (like `name`), you won't need to do this janky thing.

jbaxleyiii:
This is awesome. I wonder if this can also be done with a babel transform which would be cool

SachaG:
@tmeasday is that what you meant?

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -2113,4 +2113,39 @@ describe('queries', () => {
...
+    const data = { allPeople: { people: [ { name: 'Luke Skywalker' } ] } };
+    const networkInterface = mockNetworkInterface({ request: { query }, result: { data } });
+    const client = new ApolloClient({ networkInterface, addTypename: false });
+
+    let queryExecuted;
+    @graphql(query, {
+      alias: 'withFoo',
+    })
+    class Container extends React.Component<any, any> {
+      componentWillReceiveProps(props) {
```
tmeasday: I don't think you need to do all this; it's sort of testing the `query.metadata` feature is working but there's a separate test for that (I assume).

Can you not just do 

```js
const instance = renderer.create(<Container>);
instance.displayName;
```

?

SachaG: I get an error: `Invariant Violation: Could not find "client" in the context of "withFoo(Container)". Wrap the root component in an <ApolloProvider>`. 

But if I wrap the component in `<ApolloProvider>`, I'm not sure how to access its `displayName`…

tmeasday:
Not sure what typescript is doing but `Component.displayName` is all we needed.

SachaG:
Nice! Btw I meant to submit it as a separate PR, but I also added a short note in the README on how to work with the package locally, I thought it might help others.

PR #439, opened 2017-01-27
better skip behaviour

machard:
Hi,

I am proposing a better behaviour for the skip feature.

# Problems of the current implementation

- if skip is true, the component is (re)rendered without the data and consequently does not get update of the data even if it's updated in the query cache because of some other action that happens in the app.
- When skip switches to false, it will run the query again against the server even if forceFetch is not true and that the data is already available
- Missing some infos about knowing if this is in loading state because it's refreshing or because the result are not full (`partial` part of the pr, related to https://github.com/apollostack/apollo-client/pull/1097)

# Proposed solution

Instead of unsubscribing, resubscribing, i suggest to keep it simple and only switch noFetch to true/false depending of the skip value. and keep being subscribed.
**skip is to be seen as a way to prevent request to be run against the server when conditions are not met. But if the component defines it's underlying query, it means he wants the data if it's available!!**
this is enabled by https://github.com/apollostack/apollo-client/pull/1067.
switching noFetch from true to false will only run the request against the server if the result if not fully in cache yet or if forceFetch options is set to yes. Which is what we want

## conclusion

If you are ok with that i can finalize the PR

TODO:

- [X] Rebase your changes on master so that they can be merged easily

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -199,17 +199,12 @@ export default function graphql(
 
         this.type = operation.type;
 
-        if (this.shouldSkip(props)) return;
         this.setInitialProps();
       }
 
       componentDidMount() {
         this.hasMounted = true;
         if (this.type === DocumentType.Mutation) return;
-
-        if (!this.shouldSkip(this.props)) {
-          this.subscribeToQuery();
-        }
       }
 
       componentWillReceiveProps(nextProps) {
@@ -220,6 +215,7 @@ export default function graphql(
         if (this.type === DocumentType.Mutation) {
           return;
         };
+
         if (this.type === DocumentType.Subscription
           && operationOptions.shouldResubscribe
           && operationOptions.shouldResubscribe(this.props, nextProps)) {
@@ -227,13 +223,6 @@ export default function graphql(
           delete this.queryObservable;
           this.updateQuery(nextProps);
           this.subscribeToQuery();
-          return;
-        }
-        if (this.shouldSkip(nextProps)) {
-          if (!this.shouldSkip(this.props)) {
-            // if this has changed, we better unsubscribe
-            this.unsubscribeFromQuery();
-          }
           return;
         }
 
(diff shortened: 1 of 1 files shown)


tmeasday:
I don't really understand; if you want to use `noFetch`, why not just use `noFetch`?

machard:
because noFetch can't be a function getting parent properties

machard:
example of use in a react native app

```
// it attaches userState property
@connect(user)
// it attaches isFocused property
@focusInfos({ properties: true })
@graphql(gql`
{
  ...
}
`, {
  skip: ({ userState, isFocused }) => !utils.isReady(userState) || !isFocused,
  props: ({ data }) => ({
    data: {
      ...
    },
  }),
  options: () =>
    ({
      // i want to refresh the data every time i go back to that page
      forceFetch: true,
      reducer: (previousResult, action) => {
        ...
      },
    }),
})
export default class CurrentCoverages extends CoverageTab {
  ...
  isLoading() {
    // i want to display a loading indicator only if no "full" data is available in the local cache
    return this.props.data.partial !== false;
  }
  ...
}
```

tmeasday:
> because noFetch can't be a function getting parent properties.

But `options` can be.

> But indeed, that what i say, i believe the right behavior for skip is to enable/disable fetching

One obvious problem with this is that some people use `skip` in order to choose between two queries:

```js
@graphql(QUERY_1, { skip: ({ first }) => first })
@graphql(QUERY_2, { skip: ({ first }) => !first })
component ...
```

machard:
Haa i'll take a look at options as a function then!
I did not think of the "query switch" usage indeed

tmeasday:
>       `// i want to refresh the data every time i go back to that page`

Could you do:

```js
  options: () =>
    ({
      forceFetch: !utils.isReady(userState) || !isFocused,
      noFetch: /* the negative of that */
    });
```

machard:
using options as a function to set noFetch accordingly seems to do the same thing as i wanted to achieve with skip!
I was focusing on getting skip to work as I wished. once i made https://github.com/apollostack/apollo-client/pull/1067 i did not realize that i could actually directly use options as a function. 

@tmeasday however I still need that part https://github.com/apollostack/react-apollo/pull/439/files#diff-13fb06b5ebd9e4360f97cefd84b8c762R458. I close this PR and open the a new one just for that ?

tmeasday:
I would say yes, but it seems `partialData` is being removed soon: https://github.com/apollostack/react-apollo/issues/401#issuecomment-275556973

stubailo:
Looks like this PR is outdated and the discussion ended up with the conclusion that the specific feature here was not needed.

PR #448, opened 2017-02-03
Fix failing tests on `master`

calebmer:
This PR will fix the failing tests on `master`, and mute the warnings Jest emit.

The reason tests were failing was because Parse finally went offline! So the old GraphQL SWAPI API we hit in the tests (http://graphql-swapi.parseapp.com/) is no longer reachable. The new SWAPI is hosted on https://graphql.org/swapi-graphql, but now the [schema](http://graphql.org/swapi-graphql/schema.js) runs entirely on the client and not the server. This means there was no URL to point to.

It’s probably not a good idea that we were sending a network request to someone else’s server in our tests anyway, so this PR refactors the test in question to use a local schema.

This PR also hides expected errors from `PropTypes` in tests and upgrades Jest to fix some errors that were being logged.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
   "editor.rulers": [100],
   "files.trimTrailingWhitespace": true,
   "files.insertFinalNewline": true,
-  "editor.wrappingColumn": 100,
+  "editor.wrappingColumn": -1,
   "files.exclude": {
     "**/.git": true,
     "**/.DS_Store": true,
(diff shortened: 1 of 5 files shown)


tmeasday:
Awesome @calebmer

calebmer:
Ironically the tests still were failing because `apollo-client` updated so that the `ApolloQueryResult` type requires a type argument 😊

I just used `any` for now. We can examine how to provide better types for `react-apollo` in another PR.

jbaxleyiii:
@calebmer thank you SO much for doing this! I'm on paternity leave for a few weeks so please feel free to merge / oversee as much as you want here! @tmeasday same to you 💯

PR #461, opened 2017-02-13
Provide data prop even when skipping

### LINKED ISSUE #456: `skip` should still provide `data` prop
When using `skip()` on a query and it evaluates to true, the component no longer receives a `data` prop. This results in ugly code because this needs to be detected; you can't just use destructuring any more (`const {data: {loading, foo}} = this.props`)

I'd be happy with an empty `data` prop, put perhaps `skipped: true` and `refetch()` can be useful.

wmertens:
Fixes #456 
I suppose this is a breaking API change, since it runs `props` at times when before it wouldn't.

TODO:

### DIFF
--- a/test/react-web/client/graphql/queries.test.tsx
+++ b/test/react-web/client/graphql/queries.test.tsx
@@ -398,7 +398,7 @@ describe('queries', () => {
         queryExecuted = true;
       }
       render() {
-        expect(this.props.data).toBeUndefined();
+        expect(this.props.data).toEqual({skipped: true});
         return null;
       }
     };
@@ -424,7 +424,7 @@ describe('queries', () => {
         queryExecuted = true;
       }
       render() {
-        expect(this.props.data).toBeUndefined();
+        expect(this.props.data).toEqual({skipped: true});
         return null;
       }
     };
@@ -474,7 +474,7 @@ describe('queries', () => {
     renderer.create(<ApolloProvider client={client}><Parent /></ApolloProvider>);
   });
 
-  it('doesn\'t run options or props when skipped', (done) => {
+  it('doesn\'t run options when skipped', (done) => {
     const query = gql`query people { allPeople(first: 1) { people { name } } }`;
     const data = { allPeople: { people: [ { name: 'Luke Skywalker' } ] } };
     const networkInterface = mockNetworkInterface({ request: { query }, result: { data } });
@@ -484,14 +484,45 @@ describe('queries', () => {
     @graphql(query, {
       skip: ({ skip }) => skip,
       options: ({ willThrowIfAccesed }) => ({ pollInterval: willThrowIfAccesed.pollInterval }),
-      props: ({ willThrowIfAccesed }) => ({ pollInterval: willThrowIfAccesed.pollInterval }),
     })
     class Container extends React.Component<any, any> {
       componentWillReceiveProps(props) {
         queryExecuted = true;
       }
       render() {
-        expect(this.props.data).toBeFalsy();
+        expect(this.props.data).toEqual({skipped: true});
+        return null;
+      }
+    };
+
+    renderer.create(<ApolloProvider client={client}><Container skip={true} /></ApolloProvider>);
+
+    setTimeout(() => {
+      if (!queryExecuted) { done(); return; }
+      fail(new Error('query ran even though skip present'));
+    }, 25);
+  });
+
+  it('runs props even when skipped', (done) => {
+    const query = gql`query people { allPeople(first: 1) { people { name } } }`;
+    const data = { allPeople: { people: [ { name: 'Luke Skywalker' } ] } };
+    const networkInterface = mockNetworkInterface({ request: { query }, result: { data } });
+    const client = new ApolloClient({ networkInterface, addTypename: false });
+
+    let queryExecuted;
+    @graphql(query, {
+      skip: ({ skip }) => skip,
+      props: ({ data }) => {
+        expect(data).toEqual({skipped: true});
+        return {skippery: data.skipped};
+      }
+    })
+    class Container extends React.Component<any, any> {
+      componentWillReceiveProps(props) {
+        queryExecuted = true;
+      }
+      render() {
+        expect(this.props.skippery).toBeTruthy;
         return null;
       }
     };
@@ -517,7 +548,7 @@ describe('queries', () => {
         queryExecuted = true;
       }
       render() {
-        expect(this.props.data).toBeFalsy();
+        expect(this.props.data).toEqual({skipped: true});
         return null;
       }
     };
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- Also provide props when skipping, so destructuring props is easy [PR #461](https://github.com/apollographql/react-apollo/pull/461)
 
 ### 0.10.1
 - Fix wrong invariant sanity checks for GraphQL document [PR #457](https://github.com/apollostack/react-apollo/issues/457)
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -477,6 +477,9 @@ export default function graphql(
       }
 
       dataForChild() {
+        if (this.shouldSkip()) {
+          return {skipped: true};
+        }
         if (this.type === DocumentType.Mutation) {
           return (mutationOpts: MutationOptions) => {
             const opts = this.calculateOptions(this.props, mutationOpts);
@@ -516,10 +519,6 @@ export default function graphql(
       }
 
       render() {
-        if (this.shouldSkip()) {
-          return createElement(WrappedComponent, this.props);
-        }
-
         const { shouldRerender, renderedElement, props } = this;
         this.shouldRerender = false;
 
--- a/test/react-web/server/index.test.tsx
+++ b/test/react-web/server/index.test.tsx
@@ -249,8 +249,8 @@ describe('SSR', () => {
       );
       const apolloClient = new ApolloClient({ networkInterface, addTypename: false });
 
-      const WrappedElement = graphql(query, { options: { skip: true }})(({ data }) => (
-        <div>{data ? 'loading' : 'skipped'}</div>
+      const WrappedElement = graphql(query, { options: { skip: true }})(({ data: {skipped} }) => (
+        <div>{skipped ? 'skipped': 'dang'}</div>
       ));
 
       const app = (<ApolloProvider client={apolloClient}><WrappedElement /></ApolloProvider>);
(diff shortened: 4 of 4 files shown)


test/react-web/client/graphql/queries.test.tsx
```diff
@@ -474,7 +474,7 @@ describe('queries', () => {
     renderer.create(<ApolloProvider client={client}><Parent /></ApolloProvider>);
   });
 
-  it('doesn\'t run options or props when skipped', (done) => {
+  it.skip('doesn\'t run options when skipped', (done) => {
```
calebmer: Why is this test skipped?

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -484,14 +484,42 @@ describe('queries', () => {
...
+
+    renderer.create(<ApolloProvider client={client}><Container skip={true} /></ApolloProvider>);
+
+    setTimeout(() => {
+      if (!queryExecuted) { done(); return; }
+      fail(new Error('query ran even though skip present'));
+    }, 25);
+  });
+
+  it.skip('runs props even when skipped', (done) => {
```
calebmer: Why is this test skipped?

test/react-web/client/graphql/queries.test.tsx
wmertens: argh :) fixing

wmertens:
Fixed tests. Right about the props and the 0.x land - worst case someone is injecting props when the query runs and in the component tests for their presence. No biggie.

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -484,14 +484,42 @@ describe('queries', () => {
...
+  it('runs props even when skipped', (done) => {
+    const query = gql`query people { allPeople(first: 1) { people { name } } }`;
+    const data = { allPeople: { people: [ { name: 'Luke Skywalker' } ] } };
+    const networkInterface = mockNetworkInterface({ request: { query }, result: { data } });
+    const client = new ApolloClient({ networkInterface, addTypename: false });
+
+    let queryExecuted;
+    @graphql(query, {
+      skip: ({ skip }) => skip,
+      props: ({ data }) => ({ skippery: data.skipped })
```
calebmer: Could we add this assertion here to make sure `data` is correct in the `props` function?

```js
expect(this.props.data).toEqual({skipped: true});
```

calebmer:
I think there are some questions surrounding whether or not we should include the last set of props when skipped. However, these are questions we should be able to answer retroactively.

test/react-web/client/graphql/queries.test.tsx
wmertens: Done.

calebmer:
Looks good to me. I’ll leave this open for a day or so to give people the opportunity to comment and then we can merge 👍

stubailo:
I think this is a good change because it prevents people from adding extra checks in their UI code for missing properties, but I would be worried about adding extra functionality around `skip`. The original intention of `skip` was that it would make the HoC behave as if it weren't there at all - this means no props, no fetching, no reading from the store, etc. Basically, because it's hard to do flow control with HoCs. We should make sure that `skip` remains a reliable way to make the HoC pretend that it's not there.

For example, what does this PR do in this case?

```
@graphql(query1)
@graphql(query2, {skip: true})
MyComponent
```

Based on the contract above, you should definitely receive the results from `query1` in `MyComponent`.

stubailo:
I can't find the previous discussion, but it's also intentional that the `props` function was not executed so that people don't have to put guards for missing data in their `props` function.

wmertens:
The props function always needs to check for missing data, since it is
called during load as well.

The double graphql is a bit special, we could simply merge the two data
props?

stubailo:
I just feel like this feature is starting to go outside of what the "skip" option was designed for, which is specifically to make it look like the query or HoC never existed. The two queries on one component case is exactly why it was added.

wmertens:
Hmmm, so I have the use-case that when I get don't get a particular
prop, the query should not run.
The alternative is creating a bunch or @include or @skip directives
and deriving the required booleans, which is a) annoying and b)
unnecessary overhead… Adding a skip function is much nicer DX.

I think merging the props should handle the two queries case, no? All
that would happen is that the prop data.skipped is added...

stubailo:
@wmertens why not just pass `noFetch`? I guess it seems weird to me that the query doesn't run but you still want it to create props. Alternatively, you can use default props in React and set an empty `data` object yourself.

calebmer:
I wasn’t familiar with the previous context around `skip`. Although, I’m not sure if:

```js
@graphql(query1)
@graphql(query2, {skip: true})
```

Is a compelling use case when there are so many other ways to achieve the same functionality using both JavaScript and Apollo Client features 😣

Ideally, I would think that the `skip` functionality should live outside of `react-apollo`. Something like `applyIf(props, enhancer)`, or another technique like the ones mentioned in the recent [“Apollo and Recompose”](https://dev-blog.apollodata.com/simplify-your-react-components-with-apollo-and-recompose-8b9e302dea51#.chtpj8ypw) article. I’m not convinced on why `skip` should live in `react-apollo` basically. However, it does live here now…

I really don’t have a strong opinion either way. This is an ergonomics issue for what feels like a niche feature. Going back to the original issue, I may have responded too quickly by inviting a PR: `const {data: {loading, foo}} = this.props` does break when skipping, but not if you write it like: `const {data: {loading, foo} = {}} = this.props` (thanks ES6).

wmertens:
@stubailo correct me if I'm wrong, but as far as I know, the two options are:
```js
@graphql(
  gql`query foo($id: ID!) {
    myQ(id: id) {...}
  }`,
  {skip: ({id}) => !!id}
)
```
or
```js
@graphql(
  gql`query foo($id: ID!, $noFetch: Boolean!) {
    myQ(id: id) @skip(if: $noFetch) {...}
  }`,
  {options: ({id}) => ({variables: {id, noFetch: !!id}})}
)
```

I hope you'll agree that the first version, which is both simpler to read and to write, is the winner in terms of maintainability and ease of use…

stubailo:
@calebmer there are three things in `react-apollo` that can be achieved with other methods:

1. `skip` - it's a replacement for using something like recompose `branch` to pick which HoC to apply
2. `props` - it's a replacement for props transformations
3. `compose` - it's literally compose from recompose

These are included directly because they were the things people wanted to do most often.

If the React Apollo `skip` is not usable for this use case, I don't know what other use case it's for. We shouldn't change it into some other feature; perhaps we can remove it in a new API, or we can introduce a different feature, but that is exactly what `skip` was introduced for.

stubailo:
@wmertens I agree that code is nicer but that's just not what the 'skip' feature is intended to do.

wmertens:
@stubailo in that case I would like to request `noFetch` which is exactly like `skip` as currently implemented. I can adjust this PR for it?

wmertens:
@stubailo come to think of it, how about doing this in this PR instead:
* It merges with `this.props.data`
* It doesn't add `skipped: true`
* It doesn't run `props`
  * so if you want some other prop than `data`, you should use another way

Then it is as close as possible to the original behavior, without forcing devs to put extra `= {}` in their components. Less code is better code, better code is happier devs :)

jbaxleyiii:
@wmertens @calebmer @stubailo chiming in from our production usage.

> This is an ergonomics issue for what feels like a niche feature. 
We actually use skip a ton since our site has accounts. A lot of skipping if not logged in.

Initially we had skip implementation which did what this PR adds. However, this ran the opts and props function (with missing info as there was no `data`) so you ended up with a lot of guards in the props usages.

After https://github.com/apollographql/react-apollo/issues/245, we moved skip to bypassing the HOC all together meaning downstream components needed to add in the guards for the data instead of the HOC.

It is great for performance to not call the functions / do any work if you know to bypass. It is annoying to add guards down the tree. In our usage though, I'm happy we adjusted this behaviour.

@stubailo I'm wondering if we should write up some docs + an article on component composition like you did with the recompose article (which I loved btw!). Describing things like skip, loading components, waiting to render until all data exists, error handling, etc). Would that be valuable?

wmertens:
So this PR adds a `data` prop, which means no extra guards are needed,
since the query answers are missing before they loaded, and you need
to guard for that anyway.

calebmer:
I want to emphasize that no guards are actually necessary with ES6:

```js
const { data: { ... } = { skipped: true } } = this.props;
```

Provides the same behavior as this PR in less code.

I apologize @wmertens for encouraging you to open a PR when I didn’t have the full historical context, but I think this should be closed now 😣

I agree that it is very inconvenient, but it’s a small inconvenience that is mitigated by default property values in ES6 destructuring and the other side makes a good argument for completely skipping the HOC. So on balance I don’t think the change is worth it 😣

wmertens:
I suppose - although providing default values is a bit like a guard (it
transpiles to one), and remembering to put the guard is a bit onerous.
Furthermore, the props fn should be able to handle the skip, and if props
returns something different than `data`, it will be consistent.

I mean, I don't mind that this PR doesn't get merged, but IMHO the apollo
APIs are not very elegant nor effortless in their use (except the very
basic query, that's great). I just wanted to improve this edge case :)

calebmer:
@wmertens I agree that the APIs leave a lot to be desired, but it starts with improving the APIs for `apollo-client` which may take a while 😣

Please propose new APIs for `react-apollo`. I have some ideas, but we are definitely listening!

wmertens:
Where is a good place for proposals?

Off the top of my head, some annoyances with react-apollo are:

   - Mutations should be named by their mutation name; `mutate` should be
   the fallback name. That way it's easy to attach several at once.
   - `options` is a very opaque object, and `options.variables` is so
   common that it should be separate. Also it's hard to remember which one is
   the function (`options` or `variables` and whether `ownProps` is just an
   argument or an attribute in an argument object.
   - React-apollo could help with common problems, like only performing a
   query if a value `$foo` is given. This can be done by automatically
   creating e.g. `$fooNonNull` if it is used in the query. That would mean not
   having to do `fooNonNull: !!foo` on every such query, just write
   `query(foo: $foo) @include(if: $fooNonNull)` and it would just work.
   - In the mutation updates, I seem to recall that some of the arguments
   get the `data: ...` wrapped around, but the previous queries just have the
   actual query data. I may be misremembering.

I'd like to point out that whatever apollo-client has as API, it can be
changed a lot by react-apollo through wrapping :)

wmertens:
Another one: it was super confusing to read the example where a mutation
was pre-bound by the props function to have the `updateQueries` set up
correctly so the component only had to call `this.props.mutation(inputs)`.
Instead, I would expect that sort of setup to be part of the @graphql
configuration of the mutation.

From a componentization point of view I think it might be good to have the
queries and mutations defined in some central place, together with their
props and options and whatnot. In my limited experience, using the same
query means using the same other settings. So @graphql might take an object
as its first argument that combines query and all the other options.

Also, see
https://github.com/mattkrick/cashay/issues/150#issuecomment-280438650 where
I describe how I think optimistic updates work. Is this correct? If so, it
would be good to explain that mutation post-updates can be run with
optimistic results and should be stateless; running post-update with
optimistic and then actual server result should give the same result as
only running with server result. If not, would this requirement be a good
thing?

stubailo:
@jbaxleyiii that sounds like it would be a super helpful post! I'd be happy to feature it on our blog, or on yours, up to you.

@wmertens you're getting into a few other concepts here. We definitely believe that writing queries and mutations specific to particular components is the way to go, rather than having general queries that are reused across the app. That way your queries are componentized - in that they are part of a specific component. Unfortunately right now GraphQL itself doesn't provide any information about how a mutation result should be handled, which is why it has to be manually done in the client. But if you would like to discuss that further perhaps open a new issue.

helfer:
@wmertens. I've opened another issue for use-cases and API wish list. Please copy your comments over there so we can get a discussion started! :heart:

https://github.com/apollographql/react-apollo/issues/519

Issue #454, opened 2017-02-08
Reducer is not get called if the host component is not mounted on the mutation executed.


zhaoyao91:
Hi, from the doc, we should prefer `reducer` to `updateQueries`, so the component can take responsibility to keep itself informed, not bothering to inform other component.

But I am stuck in a situation.
Guess A is the querying comp and B is the mutating comp, and set a reducer on A, which will watches out B's mutations and try to update A's query result.
If A and B are rendered together, everything goes well: B makes some mutation and A gets updated.
But if they are rendered exclusively. say render A first, then switch to B, which then makes some mutation, and finally switch back to A, A will keep the old result for it can't see the mutation while it is unmount.

Please see [this branch](https://github.com/zhaoyao91/our-poems/tree/show_query_reducer_problem) as an example.

calebmer:
This is a use case we are exploring providing better support for in the future. For now, you need to keep the components mounted if you want their data to update.

Issue #419, opened 2017-01-15
Update queries that are no longer active, but the results are cached


bradabelgit:
Page A renders Component A which queries and displays contacts. We will call this query "contacts".

Query:

```
const gql = require("graphql-tag");
const { graphql } = require("react-apollo");

const { contactInfo } = require("../fragments");

const query = gql"
  query contacts {
    contacts {
      ...contactInfo
    }
  }
  ${ contactInfo }
";

const apolloQuery = graphql(query, { name: "contacts" })

module.exports = {
  query,
  apolloQuery
};
```

UpdateQueries:

```
graphql(ImportContacts, {
  props: ({ mutate }) => ({
    importContacts: (contacts) => mutate({
      variables: { contacts },
      updateQueries: {
        contacts: (prev, { mutationResult }) => ({
          contacts: [
            ...prev.contacts,
            ...mutationResult.data.importContacts.contacts
          ]
        }),
        accounts: (prev, { mutationResult }) => ({
          accounts: [
            ...prev.accounts,
            ...mutationResult.data.importContacts.accounts
          ]
        })
      }
    })
  })
});
```

Page B allows the user to import contacts. Page B does not render Component A, and so does not have the contacts query active.

Upon navigating back to Page A, the new contact does not exist in the list, but the old ones do (from the cache).

I figured this is where updateQueries would come in. However I have come to find that when Component A is unmounted, the query which received the data no longer exists. And so updateQueries is not called for the "contacts" query. (I confirmed this by adding the contacts query to Page B, and when the query was active, updateQueries was called as expected)

Is this the expected behaviour? If so, how should I go about handling this situation? 

I'd rather not force a refetch every time, Component A is rendered. It is getting the cached results, so shouldn't there be a way to update those cached results even though the query is no longer active?

### Version
- apollo-client@0.6.0
- react-apollo@0.8.1

tmeasday:
Hi @bradabelgit - this is a known issue with `apollo-client` I suppose. You can read a bit more about the rationale here: https://github.com/apollostack/apollo-client/issues/1068#issuecomment-268287026

I'll close this one in favor of https://github.com/apollostack/apollo-client/issues/1068

bradabelgit:
Thanks for the quick reply @tmeasday! Ah yes after following up on those links, I understand the problem. 

[...]

PR #462, opened 2017-02-14
Add recycling for `ObservableQuery`s

calebmer:
This PR is our first attempt to solve the `updateQueries` regression introduced by `apollo-client` version `0.5.22` in a bug fix while still trying to keep the bug that got fixed, well, fixed.

The issue is that when `updateQueries` or a `reducer` references data for a component which has been unmounted then there is no way to update data in the store at that location. So say you had an app where the user moved to a comment entry screen and your comments list component was unmounted. The user enters a comment and submits a mutation with an `updateQueries` that references the query for the unmounted comments list component. The `updateQueries` function would not run in this environment because of our bug fix.

Before `0.5.22` line [502 of `src/core/ObservableQuery.ts`](https://github.com/apollographql/apollo-client/blob/eada9a50e9a47fc4631a789b687e511d80892a33/src/core/ObservableQuery.ts#L502) did not remove the `ObservableQuery` from `QueryManager` even when it was added earlier along in the observable query’s lifecycle. This behavior was incorrect and caused https://github.com/apollographql/apollo-client/issues/993. So we fixed it! However, as a side effect of not removing the `ObservableQuery` from `QueryManager` any `ObservableQuery`s on unmounted components were updated. In short, the bug had a desirable side affect!

This PR aims to mitigate the problem which caused https://github.com/apollographql/apollo-client/issues/993 while still providing the desirable `updateQueries` behavior that many of our users expect. In order to strike this balance this PR implements `ObservableQuery` recycling. Here’s how it works.

When a component is unmounted, instead of throwing away the `ObservableQuery`, we put it in a recycler to keep it alive. You can think of it like putting the `ObservableQuery` in the fridge to be reheated later. When a new component mounts we take our saved `ObservableQuery` out of the fridge, add some new options, and use the old observable query again.

Now for every container component in your app (you create one whenever you call `graphql`) there should only be one `ObservableQuery` that lives forever assuming that you only render our container component once at a time. If you render your component more than once at a time that should still be fine.

When `ObservableQuery` is in the recycler it will receive any `updateQueries` and `reducer` store changes like normal. The only difference is that no component will re-render.

There are probably going to be a few bugs associated with this change. My assumption is that it will be much easier to smash those smaller bugs as they come up instead of letting this `updateQueries` bug live on. Those bugs will also likely be much smaller in scope.

Here are a few of the issues which can provide context on the issue. I may be missing some:

- This is the issue that prompted the bug fix which broke the `updateQueries` behavior. https://github.com/apollographql/apollo-client/issues/993
- Design for a new `updateQueries` (we determined that we needed a fix before we could implement the new design): https://github.com/apollographql/apollo-client/issues/1224
- The first bug report?: https://github.com/apollographql/apollo-client/issues/1129
- Potential duplicate bug report: https://github.com/apollographql/apollo-client/issues/1218
- Another bug report: https://github.com/apollographql/apollo-client/issues/1211
- Another bug report: https://github.com/apollographql/apollo-client/issues/1290
- Another bug report: https://github.com/apollographql/react-apollo/issues/454
- Another bug report: https://github.com/apollographql/react-apollo/issues/419
- Another bug report: https://github.com/apollographql/apollo-client/issues/1068

Basically, there were a lot of bug reports 😊. We didn’t fix this sooner because we saw the bug fix that caused this regression as technically correct, and we’re working on moving towards `apollo-client` 1.0.

I just want to write one more test to assert that the `updateQueries` behavior is ok. (because that’s why this PR was opened!)

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -576,12 +576,17 @@ export default function graphql(
  * An observable query recycler stores some observable queries that are no
  * longer in use, but that we may someday use again.
  *
- * Recycling observable queries avoids a few nasty bugs that may be hit when
- * using the `react-apollo` API. Namely not updating queries when a component
- * unmounts, and calling reducers/`updateQueries` more times then is necessary
- * for old observable queries.
+ * Recycling observable queries avoids a few unexpected functionalities that
+ * may be hit when using the `react-apollo` API. Namely not updating queries
+ * when a component unmounts, and calling reducers/`updateQueries` more times
+ * then is necessary for old observable queries.
  *
  * We assume that the GraphQL document for every `ObservableQuery` is the same.
+ *
+ * For more context on why this was added and links to the issues recycling
+ * `ObservableQuery`s fixes see issue [#462][1].
+ *
+ * [1]: https://github.com/apollographql/react-apollo/pull/462
  */
 class ObservableQueryRecycler {
   /**
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -191,6 +191,14 @@ export default function graphql(
 
     const graphQLDisplayName = `${alias}(${getDisplayName(WrappedComponent)})`;
 
+    // A recycler that we can use to recycle old observable queries to keep
+    // them hot between component unmounts and remounts.
+    //
+    // Note that the existence of this recycler could potentially cause memory
+    // leaks if many components are being created and unmounted in parallel.
+    // However, this is an unlikely scenario.
+    const recycler = new ObservableQueryRecycler();
+
     class GraphQL extends Component<any, any> {
       static displayName = graphQLDisplayName;
       static WrappedComponent = WrappedComponent;
@@ -210,7 +218,8 @@ export default function graphql(
       private type: DocumentType;
 
       // request / action storage. Note that we delete querySubscription if we
-      // unsubscribe but never delete queryObservable once it is created.
+      // unsubscribe but never delete queryObservable once it is created. We
+      // only delete queryObservable when we unmount the component.
       private queryObservable: ObservableQuery<any> | any;
       private querySubscription: Subscription;
       private previousData: any = {};
@@ -284,7 +293,17 @@ export default function graphql(
       }
 
       componentWillUnmount() {
-        if (this.type === DocumentType.Query) this.unsubscribeFromQuery();
+        if (this.type === DocumentType.Query) {
+          // Recycle the query observable if there ever was one.
+          if (this.queryObservable) {
+            recycler.recycle(this.queryObservable);
+            delete this.queryObservable;
+          }
+
+          // Unsubscribe from our query subscription.
+          this.unsubscribeFromQuery();
+        }
+
         if (this.type === DocumentType.Subscription) this.unsubscribeFromQuery();
 
         this.hasMounted = false;
@@ -353,14 +372,23 @@ export default function graphql(
             query: document,
           }, opts));
         } else {
-          this.queryObservable = this.client.watchQuery(assign({
-            query: document,
-            metadata: {
-              reactComponent: {
-                displayName: graphQLDisplayName,
+          // Try to reuse an `ObservableQuery` instance from our recycler. If
+          // we get null then there is no instance to reuse and we should
+          // create a new `ObservableQuery`. Otherwise we will use our old one.
+          const queryObservable = recycler.reuse(opts);
+
+          if (queryObservable === null) {
+            this.queryObservable = this.client.watchQuery(assign({
+              query: document,
+              metadata: {
+                reactComponent: {
+                  displayName: graphQLDisplayName,
+                },
               },
-            },
-          }, opts));
+            }, opts));
+          } else {
+            this.queryObservable = queryObservable;
+          }
         }
       }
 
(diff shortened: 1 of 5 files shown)


calebmer:
Ironically, now the “♻️” in the repo description means something 😊

Siyfion:
Wow, @calebmer this sounds perfect for what is required! Like you said, I assume there will be some minor bugs associated with this change, but hopefully they'll be much easier to squash than the main `updateQueries` bug.

I for one will be pulling this change asap and will start testing immediately.

calebmer:
Ok, just added the test which tests the actual behavior we want to fix with this PR (`updateQueries` not running for unmounted queries). If everyone feels comfortable with the changes made, let’s release 👍 (cc @helfer)

stubailo:
@calebmer I think this is a great approach, there's just one thing I have a question about. Since we are keeping an open subscription to each query, we are going to be reading the results out of the store even though we aren't going to be using them. If the queries are very large and there are a lot of them that could create some CPU impact.

calebmer:
@stubailo this is an unfortunate side effect. We could disable this by exposing some way in `apollo-client` to disable updating certain observers maybe observers that don’t have `observer.next`, or an observer with a private property: `observer._keepAliveOnly`. It’s fixable, but I think it’s only fixable on the `apollo-client` side.

stubailo:
Personally I think I would solve this by adding a flag in Apollo client watchQuery which makes the observable "hot" and doesn't track subscribing and unsubscribing, then always pass that flag in React-Apollo. It doesn't seem like it would take a lot of work and the benefits could be worth it.

stubailo:
We could merge and publish this first then release the other thing as an optimization.

src/graphql.tsx
```diff
@@ -543,3 +571,74 @@ export default function graphql(
...
+   * reuse have been applied.
+   */
+  public recycle (observableQuery: ObservableQuery<any>): void {
+    // Stop the query from polling when we recycle. Polling may resume when we
+    // reuse it and call `setOptions`.
+    observableQuery.stopPolling();
+
+    this.observableQueries.push({
+      observableQuery,
+      subscription: observableQuery.subscribe({}),
```
helfer: We should modify apollo-client so an active subscription isn't necessary to keep the query in the store.

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -2128,4 +2128,184 @@ describe('queries', () => {
...
+      remount();
+
+      setTimeout(() => {
+        expect(Object.keys((client as any).queryManager.observableQueries)).toEqual(['1']);
+        const queryObservable3 = (client as any).queryManager.observableQueries['1'].observableQuery;
+        expect(queryObservable1).toBe(queryObservable3);
+
+        wrapper.unmount();
+        done();
+      }, 10);
```
helfer: How are you picking the timeouts? It's 10 here, it was 5 in the other tests. it might be good if there was a constant for these somewhere.

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -2128,4 +2128,184 @@ describe('queries', () => {
...
+    const queryObservable2 = (client as any).queryManager.observableQueries['2'].observableQuery;
+    expect(queryObservable1).not.toBe(queryObservable2);
+
+    remount();
+
+    setTimeout(() => {
+      expect(Object.keys((client as any).queryManager.observableQueries)).toEqual(['1', '2']);
+      const queryObservable3 = (client as any).queryManager.observableQueries['1'].observableQuery;
+      const queryObservable4 = (client as any).queryManager.observableQueries['2'].observableQuery;
+      expect(queryObservable3).toBe(queryObservable1);
```
helfer: Apollo Client queryIds are strictly incrementing, so looking for `observableQueries[1]` twice will always either return the same object or `undefined` if that observable query has been removed. So I think testing that `queryObservable3` is equal to `queryObservable1` doesn't actually show anything.

test/react-web/client/graphql/queries.test.tsx
```diff
@@ -2128,4 +2128,184 @@ describe('queries', () => {
...
+        const queryObservable3 = (client as any).queryManager.observableQueries['1'].observableQuery;
+        expect(queryObservable1).toBe(queryObservable3);
+
+        wrapper.unmount();
+        done();
+      }, 10);
+    }, 10);
+  });
+
+  it('will not recycle parallel GraphQL container `ObservableQuery`s', done => {
```
helfer: I'm not sure what this test is trying to do. Is it supposed to show that if you have the same query rendered twice you actually get two different observable query objects? If so, why is there a `remount()` in there?

test/react-web/client/graphql/mutations.test.tsx
```diff
@@ -394,4 +394,218 @@ describe('mutations', () => {
     renderer.create(<ApolloProvider client={client}><Container id={'123'} /></ApolloProvider>);
   });
 
+  it('will run `updateQueries` for a previously mounted component', () => new Promise((resolve, reject) => {
```
helfer: This test is really long and keeps track of a lot of stuff, so it would be useful to have a short comment here that summarizes what it does.

Siyfion:
I've been testing this since @calebmer made the PR and so far, so good, no new bugs have been found and it definitely resolves the issue.

src/graphql.tsx
```diff
@@ -543,3 +571,74 @@ export default function graphql(
 
   return wrapWithApolloComponent;
 };
+
+/**
+ * An observable query recycler stores some observable queries that are no
+ * longer in use, but that we may someday use again.
+ *
+ * Recycling observable queries avoids a few nasty bugs that may be hit when
```
calebmer: > It's also important to note that recycling query observables will still not provide the ability to update parts of the cache for variables more than one step back.

Did not think about that as a use case we should look at supporting. I don’t think the cost of keeping around all observable queries is worth that, however. Hopefully imperative `read` and `write` methods will allow users to implement this themselves where necessary.

test/react-web/client/graphql/queries.test.tsx
calebmer: Arbitrarily picking numbers for no reason 😉

This is a very unimportant number that could go anywhere from 0–1000 and the test would behave the same.

test/react-web/client/graphql/queries.test.tsx
calebmer: Yep. It shows that if you render two components with the same query you get two observable queries, but if you unmount and remount one of the components it will recycle that observable query so that in total you only ever have two observable queries.

test/react-web/client/graphql/queries.test.tsx
calebmer: If `queryObservable3` could be `undefined` if it was removed, then this test does show something, right?

What I really wanted to test was the `ObservableQuery` instance directly on the components, but that value is really hard to get.

helfer: Yeah, then maybe compare directly to undefined and write a comment about what we'd really like to check here.

helfer:
@calebmer I've added you to `react-apollo` on npm. Feel free to merge this and release when you consider it ready.

calebmer:
Released in `0.11.1` 🎉

booboothefool:
Thanks @calebmer ur a fuggin beast. 💪

tmeasday:
Rather than making the `recycler` global, would it make sense to attach it to the provider (or the client) @calebmer?

I just got really confused in a testing scenario where I was setting up a new provider and new client for each test case, but queries get reused anyway.

calebmer:
@tmeasday There is one recycler per every HOC component. So it isn’t technically global, but every HOC component has one and only one recycler, but I do see the problem 😊

A solution would be to, as you suggest, create a `WeakMap<ApolloClient, ObservableQueryRecycler>` and then we use the recycler for whichever `client` reference we have at the time. I think since we assume `Promise`s exist in the environment we can also assume `WeakMap`s exist? Because both objects were part of the ES2015 specification. We should really be using `Map`s more in `apollo-client` as well…

tmeasday:
Could we move the `const recycler = new ObservableQueryRecycler();` line inside the `class GraphQL`? -- that way there'd be one per component instance -- which would also fail in some cases (when the `props` of the `ApolloProvider` changed without unmounting for instance), but at least would be easily avoidable via the use of `key={xyz}`

calebmer:
No, the point of the recycler is that observable queries will be recycled _across_ component instances. If we remove that capability then it is basically pointless and we are back where we started as far as the `updateQueries` issue goes 😣

I think the `WeakMap` on instances of `ApolloClient` is safe as long as we are comfortable with requiring a ES2015 collections polyfill (which I think we should for `apollo-client`). We could also have the recycler ask for an instance of `client` on recycle and reuse (or `queryManager`). Then it would only reuse an observable query if the client is the same.

tmeasday:
Makes sense, I opened: https://github.com/apollographql/react-apollo/issues/513

booboothefool:
I was locked down on apollo-client 0.5.21, react-apollo 0.7.3 for a while due to https://github.com/apollographql/apollo-client/issues/1129, however `updateQueries` still not does seem to work after performing the following upgrade.

screen shot 2017-03-26 at 12 15 54 am

Is there a step I'm missing?

booboothefool:
@calebmer I'm pretty sure everything is installed correctly / cache cleared, etc. 

I even tried the new `update: (proxy, mutationResult)` and that works fine for a bunch of cases, however my old `updateQueries` still not firing.

There might be something wrong with my implementation for this particular part I am trying it on because `update` gives me "Store error: the application attempted to write an object with no provided id but the store already contains an id of ____ for this object".

apollo-client 0.5.21, react-apollo 0.7.3: `updateQueries` fires and updates like I wanted it to

apollo-client 1.0.0-rc.6, react-apollo 1.0.0-rc.3: `updateQueries` doesn't fire, `update` gives me that error

PR #476, opened 2017-02-23
Fix silent errors

calebmer:
Logs an error to the console now instead of silently ignoring errors when a user does not explicitly use the `error` property on our `data` object.

This is one of the error changes we wanted to make for AC 1.0.

Interestingly enough, this change caught a couple of bugs in our test suite 😊

cc @helfer, @martijnwalraven


jbaxleyiii:
@calebmer this is great!

src/graphql.tsx
```diff
@@ -530,7 +530,29 @@ export default function graphql(
...
+          // 10 milliseconds is an arbitrary number picked to work around any
+          // potential asynchrony in React rendering. It is not super important
+          // that the error be logged ASAP, but 10 ms is enough to make it
+          // _feel_ like it was logged ASAP while still tolerating asynchrony.
+          let logErrorTimeoutId = setTimeout(() => {
+            if (error) {
+              console.error('Uncaught (in react-apollo)', error.stack || error);
+            }
+          }, 10);
+          Object.defineProperty(data, 'error', {
```
stubailo: Would there be any benefit in only doing this in development mode? Or is the overhead minimal enough that we should just always do it?

jbaxleyiii:
Are console logs still slow in react native?

src/graphql.tsx
calebmer: I’m for always logging because:

1. The overhead is fairly small.
2. In production a user should really be catching the error anyway.
3. Some error tracking libraries may report `console.error` exceptions.

calebmer:
@jbaxleyiii as far as I know it’s a safe assumption that React Native logs will always be slow 😣

However, for just printing an error stack trace which is a few bytes of text I highly doubt there will be any performance impact. The real killers for logging in React Native are sophisticated logging libraries like [`redux-logger`](http://npmjs.com/redux-logger) which dump the entire Redux before/after state on every action.

Also, users should really be expected to catch and handle these errors, so even if there is some performance impact on the rare event there is an error it is a (mostly) one time cost and it has an easy fix on the user’s part.

jbaxleyiii:
That makes total sense!

calebmer:
Fixed CI 👍

helfer:
Neat!

PR #479, opened 2017-02-24
Allow ApolloProvider Prop Updates

wmcbain:
We have a use case in React Native where we switch the GraphQL endpoint dynamically while the app is running. This PR changes the ApolloProvider API by updating it's client and store when new props are delivered to the provider.

TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- ApolloProvider now changes it's client and store when those props change. [PR #479](https://github.com/apollographql/react-apollo/pull/479)
 
 ### 1.0.0-rc.1
 - Update dependency to Apollo Client 1.0.0-rc.1 [PR #520](https://github.com/apollographql/react-apollo/pull/520)
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
-- ApolloProvider now changes it's client and store when those props change. [PR #479](https://github.com/apollographql/react-apollo/pull/479)
+- ApolloProvider now changes its client and store when those props change. [PR #479](https://github.com/apollographql/react-apollo/pull/479)
 
 ### 1.0.0-rc.1
 - Update dependency to Apollo Client 1.0.0-rc.1 [PR #520](https://github.com/apollographql/react-apollo/pull/520)
--- a/src/ApolloProvider.tsx
+++ b/src/ApolloProvider.tsx
@@ -10,7 +10,9 @@ import {
   Store,
 } from 'redux';
 
-import ApolloClient from 'apollo-client';
+/* tslint:disable:no-unused-variable */
+import ApolloClient, { ApolloStore } from 'apollo-client';
+/* tslint:enable:no-unused-variable */
 
 import invariant = require('invariant');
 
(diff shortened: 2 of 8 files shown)


stubailo:
Nice! This looks like a great change.

jbaxleyiii:
@wmcbain if you can add a couple tests for this, we can merge it and release it! Thanks!

src/ApolloProvider.tsx
```diff
@@ -43,6 +43,14 @@ export default class ApolloProvider extends Component<ProviderProps, any> {
   constructor(props, context) {
     super(props, context);
 
+    this._init(props);
+  }
+
+  componentWillReceiveProps(nextProps) {
```
calebmer: I think this should be `componentDidUpdate`. We also wouldn’t need to pass in `props` in that case.

src/ApolloProvider.tsx
```diff
@@ -61,7 +69,6 @@ export default class ApolloProvider extends Component<ProviderProps, any> {
     // intialize the built in store if none is passed in
     props.client.initStore();
     this.store = props.client.store;
```
calebmer: Shouldn’t the client and store be state instead of class properties? Or better yet, why do we even need them to be class properties? We can just call `initStore` on them when they change.

src/ApolloProvider.tsx
calebmer: Also, we should probably put this in an `if` that checks if we actually have a new client.

src/ApolloProvider.tsx
```diff
@@ -10,7 +10,9 @@ import {
   Store,
 } from 'redux';
 
-import ApolloClient from 'apollo-client';
+/* tslint:disable:no-unused-variable */
+import ApolloClient, { ApolloStore } from 'apollo-client';
```
wmcbain: TS compiler was throwing an error without `ApolloStore` not being in scope when potentially returning an `ApolloStore` in `getChildContext`.

stubailo: Yeah this is one weird edge case in TS

Changelog.md
```diff
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- ApolloProvider now changes it's client and store when those props change. [PR #479](https://github.com/apollographql/react-apollo/pull/479)
```
stubailo: It's -> its

wmcbain:
Ready for review

test/react-web/client/ApolloProvider.test.tsx
```diff
@@ -103,4 +132,91 @@ describe('<ApolloProvider /> Component', () => {
...
+    container.setState({ client: newClient });
+
+    expect(initStoreMock).not.toHaveBeenCalled();
+
+    const newStore = createStore(() => ({}));
+    container.setState({ store: store });
+    expect(initStoreMock).not.toHaveBeenCalled();
+  });
+
+  it('child component should be able to query new client and store when props change', () => {
```
wmcbain: @calebmer Test added

helfer:
@calebmer I'll let you finish reviewing this and include it in the next release.

calebmer:
The new test is good, but I really want to see something that uses the `graphql` HOC 😊

When the client changes, does the HOC make a new request to the new client? When the client changes and then `refetch` is called from the HOC, is the request executed with the new client?

These are behaviors I’d like to see asserted with tests.

wmcbain:
@calebmer I'm not sure if I'm familiar enough with this library to provide the depth of tests that you are looking for. I'm struggling to provide these tests within reasonable time as a result.

Is there an example that you can point to that would help me see what you're looking for?

test/react-web/client/ApolloProvider.test.tsx
```diff
@@ -103,4 +156,118 @@ describe('<ApolloProvider /> Component', () => {
...
+      networkInterface: nextInterface,
+    });
+
+    container.setState({
+      client: nextClient,
+    });
+
+    // Both cases fail
+    expect(initialInterface.query).not.toHaveBeenCalled();
+    expect(nextInterface.query).toHaveBeenCalled();
```
wmcbain: Hey @calebmer, revisited this problem after learning a bit more and running into a bug that fit the test case you described exactly in this PR. I'm wondering now if you foresaw this being an issue?

Any help/insight is appreciated.

calebmer: I added a better test and fixed the bugs the test revealed. I’ll merge once CI passes 👍

calebmer:
Released in 1.0.0-rc.3 🎉

wmcbain:
Thanks!

Issue #464, opened 2017-02-16
Proposal: API changes to support multiple clients


1st8:
Hey,

I would like to add support for multiple clients, similar to http://dev.apollodata.com/angular2/multiple-clients.html

Therefore I am proposing the following API changes and would like to have feedback before preparing a PR for this.

### ApolloProvider

```jsx
const ExtraProvider = createApolloProvider({ for: 'extraClient' })
// We could create the default ApolloProvider with createApolloProvider({ for: 'default' }) too
ReactDOM.render(
  <ApolloProvider client={client}>
    <ExtraProvider client={extraClient}>
      {...}
    </ExtraProvider>
  </ApolloProvider>,
  rootEl
)
```

### graphql

New config key `use` to select the client to be used for the given statement.

```js
// like before: uses the default client provided as "client" to execute given query
// new: uses client provided as "extraClient" to execute given specialQuery
export default compose(
  graphql(query),
  graphql(specialQuery, { use: 'extraClient' }),
)(Component);
```

Cheers!

jbaxleyiii:
I think this would be great!

1st8:
My mistake, this API is not possible (maybe there is a complicated solution), because `childContextTypes` has to be defined statically on `ApolloProvider`.

I will update my proposal to something that should work easier.

1st8:
I just pushed a work-in-progress commit, in case someone wants to see working code: https://github.com/1st8/react-apollo/commit/63c42dbebeed4f718a6a98b151f1a694cc5ebdb2#diff-ff711df5e0aa7417faaf90767b79ac8bR35

Looking forward to get feedback. Cheers!

calebmer:
Instead of implementing a `provideClient` function, could you instead add functionality to the `ApolloProvider` component? So:

```jsx
<ApolloProvider client={client} as="extraClient">
  {...}
</ApolloProvider>
```

I am curious, why would you want this feature? Also, why would this be better than passing a client instance directly to the `graphql` function? Like so:

```js
export default graphql(MyQuery, {
  client: ...,
})(MyComponent);
```

…or if a subtree needed a different client rendering an `<ApolloProvider>` just for that subtree.

I would prefer just passing in clients directly for an initial implementation.

1st8:
> Instead of implementing a provideClient function, could you instead add functionality to the `ApolloProvider` component?

[...]

PR #481, opened 2017-02-24
Extend ApolloProvider and graphql API to support multiple clients.

1st8:
Related discussion: #464 
Documentation PR: https://github.com/apollographql/react-docs/pull/163

TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- Extend ApolloProvider and graphql API to support multiple clients. [PR #481](https://github.com/apollographql/react-apollo/pull/481)
 
 ### 0.11.2
 - Remove `@types/chai` dev dependency which called a reference to the `chai` types in the production build. [PR #471](https://github.com/apollographql/react-apollo/pull/471)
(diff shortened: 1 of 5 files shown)


calebmer:
I’m still opposed to this design 😣

1. It is a feature that assume a certain use case instead of a general feature that gives users the highest power to do what they want.
2. `use` is static.
3. It complicates `ApolloProvider`s alternate usage as a Redux context provider.
4. It is challenging to maintain and extend by adding new features which also require plucking the client off of context.

I’m more interested in providing the basic tools to make this implementation possible and have specific `use`/`as` semantics implemented outside of `react-apollo` in a user’s codebase or in a community maintained package.

It’s unfortunate that we had to remove the prop behavior in the past because of prop naming collisions (see commit: https://github.com/apollographql/react-apollo/commit/626dbb715da31d1de4d416d27e51c62c6b885bef), so instead of reintroducing that we should add `client` to the `options` function which can be computed from props. Such a design may look like so:

```js
graphql(query, {
  options: props => ({ client: ... }),
})(MyComponent)
```

If a `client` does not exist, or it is null then we use the client from `context`. This is the minimum we can add to enable the maximum number of use cases including enabling you to build this functionality into a community maintained package.

Thoughts?

jbaxleyiii:
@calebmer I'm good with the options version

helfer:
I'd be okay with the `options` way that @calebmer proposed.

1st8:
I won't be preparing a new PR in the near future (vacation coming up 🎉 ).
Changing this one doesn't make sense as well.
Still think that I don't want any clients in my props, but maybe thats just me.

Thanks for your time and feedback, I am gonna go ahead and close this.

flexzuu:
Hey I implemented changes proposed by @calebmer. Currently i can't get the test-setup to work. [#717](https://github.com/apollographql/react-apollo/issues/717)
But my integration-test in a simple create-react-app is working fine. I will open a new PR linking to this one and [#464](https://github.com/apollographql/react-apollo/issues/464) once the testing issue is resolved.

PR #490, opened 2017-02-27
WIP: Include peers

helfer:
TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- Make apollo-client and graphql-tag dependencies and re-export them from this package [PR #490](https://github.com/apollographql/react-apollo/pull/490)
 
 ### 0.12.0
 - Update Apollo Client to 0.9.0 and bump a lot of other dependencies [PR #484](https://github.com/apollographql/react-apollo/pull/484)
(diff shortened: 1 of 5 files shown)


jaydenseric:
@helfer After updating to v0.13.0, `ApolloClient` or `ApolloClient` is not available for import. They simply don't appear in the compiled dist.

helfer:
@jaydenseric Yeah, I figured that out today as well. I'll have to dig into it.

helfer:
@jaydenseric actually, I'm not sure why `dist` is included, because in package.json main is specified as `lib/index.js`. My guess as to why the exports are not included is that maybe the bundler uses the `browser` field instead, which is `lib/browser.js`, but as far as I can tell react-apollo's build setup doesn't actually write to that file!

cc @calebmer @jbaxleyiii

jaydenseric:
I can't be much help here because it's TypeScript :(

While messing around with this stuff, what are the chances of adding a  [package.json `module`](https://github.com/rollup/rollup/wiki/pkg.module) entry for modern tools (such as Webpack 2) to consume ESM?

jaydenseric:
Created a proper issue regarding the `module` enhancement: https://github.com/apollographql/apollo/issues/78.

calebmer:
Yeah, I’m not sure how the build is setup in `react-apollo`. I need to take a look at the builds for both `apollo-client` and `react-apollo` 😣

helfer:
@calebmer: please focus on react-apollo first, because the build here seems somewhat broken while for Apollo Client it's still working.

calebmer:
@helfer do we want to schedule this as a fix for 1.0?

helfer:
Yes, for this repo definitely.

helfer:
@calebmer I was stupid and should have put the exports in `browser.ts` instead of `index.ts`. Feel free to ignore what I said before.

PR #493, opened 2017-03-01
Stop polling in server side rendering

calebmer:
This is a quick solution to https://github.com/apollographql/apollo-client/issues/1334. The correct solution would be to never start polling in the first place. If someone wants to add a test for this, it would be much appreciated because I may not get to it in a bit.

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -432,7 +432,9 @@ export default function graphql(
         const result = observable.currentResult();
 
         if (result.loading) {
-          return observable.result();
+          const result = observable.result();
+          observable.stopPolling();
+          return result;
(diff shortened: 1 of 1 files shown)


helfer:
@calebmer shouldn't we just not poll if Apollo Client is run in SSR mode?

calebmer:
@helfer yes. Is this something we are able to do in `apollo-client`?

helfer:
@calebmer I think the `ssrMode` flag in Apollo Client should give us all the information we need. I'll make sure it's taken care of in Apollo Client.

calebmer:
Ok. Closing 👍

Issue #174, opened 2016-08-28
[0.4.7] Weird problem with react-hot-loader@3.0.0-beta.2


HriBB:
I am seeing some really weird behavior with `react-hot-loader@3.0.0-beta.2` and `react-apollo@0.4.7`.

To make things easier to explain, this is my component hierarchy:
`Admin > Modal > AddAlbum > [Title, Content, Footer]`

`Admin` is my `react-router` `Route`.

If I enhance `AddAlbum` with `graphql()` and put it inside `Modal`, first hot reload does not work. When I change `Title` for example, nothing is updated. However, when I change `Title` again, or change `Content`, hot reload works and components are updated.

If I put `AddAlbum` directly in `Admin`, hot reload works all the time.
If I do not enhance `AddAlbum` with `graphql()`, hot reload works all the time.

I tried to create a reproducible example with `react-hot-boilerplate@next`, but I could not do it. Hot reloading always worked. I suspect that this problem has got something to do with my project setup, but I was not able to pinpoint it. It's really frustrating to work like this, when hot reload works for everything but graphql components.

Is anyone else seeing this issue?

HriBB:
I understand that this is a very low-priority bug, if it can even be considered a bug. You can close it if you wish, just wanted to get some feedback from the community.

stubailo:
I think it's a pretty good bug to fix, but it's definitely hard to know what's going on without a concrete reproduction.

HriBB:
I will try to create a reproducible example, but it might take some time, because I'm pretty busy these days.

nickpisacane:
After upgrading from 0.3.x to 0.4.7 (now 0.5.4) I am noticing issues with hot-reloading as well (since the upgrade to 0.4.7). I will look more into this, but it seems to be something odd with the graphql higher order component. The project I am working on is using redux, and react-router. When Components are decorated with `connect`, or `withRouter` from redux, and react-router respectively on top of `graphql`, hot-reloading works. Yet, by itself, hot-reloading doesn't seem to work with graphql.

jbaxleyiii:
@HriBB @nickpisacane any chance you could set up a sample repo with the issue? If love to fix this but don't use HMR myself

HriBB:
> but it seems to be something odd with the graphql higher order component
@nickpisacane my thoughts as well.

@jbaxleyiii here you go https://github.com/HriBB/react-apollo-rhl-problem Let me know if you need any assistance.

[...]

PR #505, opened 2017-03-03
Check if WrappedComponent implementation changed since last render to fix HMR

brunoabreu:
As documented in: 
https://github.com/apollographql/react-apollo/issues/174
https://github.com/apollographql/apollo-client/issues/764
https://github.com/HriBB/react-apollo-rhl-problem

Hot Module Replacement doesn't work when your component is wrapped by the GraphQL HOC. This happens because `render()` reutilizes `this.renderedElement` even though the `WrappedComponent` implementation changed during hot reloading.

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -577,7 +577,7 @@ export default function graphql(
         const clientProps = this.calculateResultProps(data);
         const mergedPropsAndData = assign({}, props, clientProps);
 
-        if (!shouldRerender && renderedElement) {
+        if (!shouldRerender && renderedElement && renderedElement.type === WrappedComponent) {
           return renderedElement;
         }
 


src/graphql.tsx
```diff
@@ -577,7 +577,7 @@ export default function graphql(
         const clientProps = this.calculateResultProps(data);
         const mergedPropsAndData = assign({}, props, clientProps);
 
-        if (!shouldRerender && renderedElement) {
+        if (!shouldRerender && renderedElement && renderedElement.type === WrappedComponent) {
```
calebmer: How does this fix your problem?

firstdoit: Hi @calebmer! I've diagnosed and debugged this with @brunoabreu.
Basically, when a hot module replacement happens, the `render` method of the `GraphQL` is called again and has the correct, new implementation of `WrappedComponent` in scope. 

However, because the wrapped element was already rendered, this `if` is true and we don't get to re-create the element using the new `WrappedComponent` implementation.

We initially checked it works by manually doing `renderedElement = createElement...` and the new component would be rendered.

I believe the intention here is to avoid re-rendering the wrapped component if the data didn't change (which is great), but the original implementation didn't consider that the Component _itself_ might have changed (as is desired during HMR).

This basically checks that the renderedElement is up-to-date with the implementation. If the WrappedComponent didn't change (as it wouldn't without HMR), then it's referentially equal to `renderedElement.type`.

Or maybe we're mistaken, but this seems to be the case from what we've gathered. 😅

calebmer: I find it hard to believe that `WrappedComponent` would ever referentially change without a new instance of the `GraphQL` component being created 😣

Unless React Hot Loader copies over all of the properties on the old component to the new component, so a `renderedElement` with the incorrect `type` may be copied over. Can you confirm this happens?

More importantly, can you confirm that this actually fixes your issue?

If so, this looks fine to me. If we could just get a comment that reads something like this 😊:

```js
// If we were not marked to perform a new render, and we have a previous
// rendered element of the same type as the component we are wrapping then let
// us just return that previous element instead of rendering a new one.
//
// We check the type because it may be possible with some React hot loading
// implementations that a `renderedElement` is copied over with the old
// component type. Since we do not want to render the old element again, the
// type check guarantees we render a new element in this case.
```

firstdoit: > I find it hard to believe that WrappedComponent would ever referentially change without a new instance of the GraphQL component being created 😣

We too were dumbfounded by this. Here's an idea: we'll create a repo reproducing the problem and we take it from there. However it's Friday night already in Rio so we will continue on Monday. Deal? 😃 

Thanks for the help!

firstdoit: Okay, @calebmer, here goes: https://github.com/firstdoit/graphql-hmr

calebmer: Thanks so much @firstdoit for the example! After some research it does look like that every time this version of React Loader reloads your component it tries to keep the same instance. So your React component instance will be `===` no matter the prototype. The prototype will, however, be updated as necessary to actually hot load the component. Since we don’t change out the component instance we will still have an old rendered element with an old type, so just like we thought 😊

I think this is an implementation detail specific to this version of `react-hot-loader`, however, this change is a logical one to make and has no negative impacts, so let’s merge 👍

calebmer:
Released in `0.13.3` 🎉

Thanks so much @firstdoit and @brunoabreu you were both a pleasure to work with 😊

firstdoit:
Likewise!

PR #506, opened 2017-03-06
fix: Move constructor initializing of props to componentWillMount

drager:
React outputs setState warnings (constructor side-effects are an anti-pattern), because of that initializing of props in the constructor and by moving that code to componentWillMount fixes that.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 - Make sure that the cached rendered element has the correct type before returning it. [PR #505](https://github.com/apollographql/react-apollo/pull/505)
+- Move constructor initializing of props to componentWillMount. [PR #506](https://github.com/apollographql/react-apollo/pull/506) ([Issue #509](https://github.com/apollographql/react-apollo/issues/509)).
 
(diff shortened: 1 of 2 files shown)


calebmer:
We don’t set state anywhere in that code 😣

Could you make a small reproduction with [`react-apollo-error-template`](https://github.com/apollographql/react-apollo-error-template) so I can see where the error is being triggered?

calebmer:
It looks like this error was reproduced in this issue: https://github.com/apollographql/react-apollo/issues/509

@drager can you confirm this is the same issue?

I want to take some time to look at the reproduction and understand what is really going on to see if we can get a more specific fix.

calebmer:
After more investigation in https://github.com/apollographql/react-apollo/issues/509 it looks like this does indeed fix the problem. I cleaned it up a little, @drager could you add a changelog entry and then we can merge? 😊

drager:
@calebmer: Thanks, sorry for the late answer but great that #509 could confirm this issue. Will add a changelog entry as soon as possible. :)

morgante:
When can we expect this to be released? Has anyone found a workaround for now?

calebmer:
Released in `0.13.3` 🎉

PR #515, opened 2017-03-08
Rewrite the README

calebmer:
I was just going to add some stuff to the `README`, but it was pretty bare anyways so I ended up rewriting it.

cc @stubailo, @helfer


README.md
```diff
@@ -1,38 +1,137 @@
...
-Use your GraphQL server data in your React components, with the [Apollo Client](https://github.com/apollostack/apollo-client).
+React Apollo unlike many other tools in the React ecosystem requires _no_ complex build setup to get up and running. As long as you have a GraphQL server you can get started building out your application with React immeadiately. React Apollo works out of the box with both [`create-react-app`][] and [React Native][] with a single install and with no extra hassle configuring Babel or other JavaScript tools.
 
-- [Install](#install)
-- [Docs](http://docs.apollostack.com/apollo-client/react.html)
+[`create-react-app`]: https://github.com/facebookincubator/create-react-app
+[React Native]: http://facebook.github.io/react-native
 
-### Documentation
+The Apollo frontend toolchain is built to be:
```
helfer: How about we just say "Apollo is:"

README.md
```diff
@@ -1,38 +1,137 @@
...
-Running tests locally:
+That’s it! You may now use React Apollo in any of your React environments.
+
+For an amazing developer experience you may also install the [Apollo Client Developer tools for Chrome][] which will give you inspectability into your React Apollo data.
+
+[Apollo Client Developer tools for Chrome]: https://chrome.google.com/webstore/detail/apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm
+
+## Usage
 
+To get started you will first want to create an instance of [`ApolloClient`][] and then you will want to provide that client to your React component tree using the [`<ApolloProvider/>`][] component. Finally, we will show you a basic example of connecting your GraphQL data to your React components with the [`graphql()`][] enhancer function.
```
helfer: I think a lot of the sentences in this PR would be more straight forward to read if instead of saying: "TO X you will want to Y" it just said "Do Y to get X".

README.md
```diff
@@ -1,38 +1,137 @@
-# react-apollo
+# [React Apollo](http://dev.apollodata.com/react/) [![npm version](https://badge.fury.io/js/react-apollo.svg)](https://badge.fury.io/js/react-apollo) [![Get on Slack](https://img.shields.io/badge/slack-join-orange.svg)](http://www.apollostack.com/#slack)
 
-[![npm version](https://badge.fury.io/js/react-apollo.svg)](https://badge.fury.io/js/react-apollo)
-[![Get on Slack](https://img.shields.io/badge/slack-join-orange.svg)](http://www.apollostack.com/#slack)
-[![Build Status](https://travis-ci.org/apollographql/react-apollo.svg?branch=master)](https://travis-ci.org/apollographql/react-apollo)
-[![Coverage Status](https://coveralls.io/repos/github/apollographql/react-apollo/badge.svg?branch=master)](https://coveralls.io/github/apollographql/react-apollo?branch=master)
```
helfer: I think build status and coverage status are useful, we should keep them. If they pass and coverage is high, that usually indicates that the project is maintained and more or less in shape.

calebmer: I don’t really like them because sometimes they have false negatives, like a build could be failing on a branch and so the build status goes red. That’s definitely the last thing we want 😣

IMO build status and coverage status badges are pretty weak signals compared to things like commit count, issue count, stars, last commit timing, etc. and they have the possibility of displaying a false negative.

If we want to get a badge that always says “build passing” and 90%+ coverage no matter what the temporary status looks like, I’m down for that 😉

Thoughts?

README.md
calebmer: The thing is that would put “get started” last which is a trigger phrase for a lot of people 😣

How would you rephrase this particular paragraph?

helfer: "To get started" can definitely go first, I'd just try to use "you will want to" a little bit less.

README.md
helfer: If you're concerned about build status sometimes going red, how about we just put coverage there? I think that rarely accidentally goes to zero. I actually think it's also quite rare for the build on master to fail, so I'm not too worried about having both badges there.

helfer:
I'm going to merge it as-is so I can release a new version. We should make a team decision about badges for our repos at some point.

PR #537, opened 2017-03-16
Fix #312

### LINKED ISSUE #312: graphql is caching the variables when using object instead of function in options
Usually I have this type of code:
```Javascript
graphql(query, {
  options: ({ id }) => ({ variables: { id }, fragments }),
})
```

But since I don't need to pass any args explicitly, it could be write as:
```Javascript
graphql(query, {
  options: () => ({ fragments }),
})
```

Or:
```Javascript
graphql(query, {
  options: { fragments },
})
```

**BUG**: this third case, without a function, is using a cached `variables`. So, the first time a item is rendered, it runs correctly. But if I the navigate to another item, with different `id`, it uses the same old `id` of the first item. 

*ps1: I prefer to always explicitly pass the variables, but found this bug while prototyping.*
*ps2: v0.5.16*

calebmer:
Fixes https://github.com/apollographql/react-apollo/issues/312

cc @drager, @steventebrinke

We were mutating the `options` object where we weren’t supposed to. So if a user had used an object instead of a function for their `options` then in some cases we would be setting the `variables` property incorrectly. I found the bug by calling `Object.freeze` on `options` if it was not a function.

There error was reproduced in: https://github.com/steventebrinke/react-apollo-error-template/commit/e80ca5ed7a89cbed0b7d39f44d24ffb433f3e111

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
+- Fix bug where `options` was mutated causing variables to not update appropriately. [PR #537](https://github.com/apollographql/react-apollo/pull/537)
 
 ### 1.0.0-rc.1
 - Update dependency to Apollo Client 1.0.0-rc.1 [PR #520](https://github.com/apollographql/react-apollo/pull/520)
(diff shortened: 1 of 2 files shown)


drager:
@calebmer: This solves the bug for me. Thank you!

steventebrinke:
Thanks, this also solves it in the real app in which I discovered the bug.

calebmer:
Released in 1.0.0-rc.3 🎉

alfonsoperez:
This solves also an instance in which a query was not being triggered if specified the options as an object. Thanks!

PR #548, opened 2017-03-20
Include cached data with error

StevePotter:
This PR includes data cached from the Apollo store when there is an error while fetching the query.  Now useful data can be displayed along with any error message.

For example, the server is down.  An error will be returned as it always was.  But until now, the `data` prop was empty.  A good UI for this situation would be to display an error message like "There was an error reaching our servers.  Until the problem is fixed, you can still use the app.  Your data might just be a bit off." and show the cached data.  Now you can!

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -558,8 +558,16 @@ export default function graphql(
             // while loading, we should use any previous data we have
             assign(data, this.previousData, currentResult.data);
           } else {
-            assign(data, currentResult.data);
-            this.previousData = currentResult.data;
+            let dataToMerge = currentResult.data;
+            // if we have an error, include any previously cached data
+            if (error) {
+              let previousResult = this.queryObservable.getLastResult()
+              if (previousResult && !previousResult.error) {
+                dataToMerge = assign({}, previousResult.data, currentResult.data)
+              }
+            }
+            assign(data, dataToMerge);
+            this.previousData = dataToMerge;
           }
         }
         return data;
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -561,9 +561,9 @@ export default function graphql(
             let dataToMerge = currentResult.data;
             // if we have an error, include any previously cached data
             if (error) {
-              let previousResult = this.queryObservable.getLastResult()
+              let previousResult = this.queryObservable.getLastResult();
               if (previousResult && !previousResult.error) {
-                dataToMerge = assign({}, previousResult.data, currentResult.data)
+                dataToMerge = assign({}, previousResult.data, currentResult.data);
               }
             }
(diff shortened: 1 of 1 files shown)


stubailo:
Can you add a test for this as well?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-apollo",
-  "version": "1.0.0-rc.3",
+  "version": "1.0.0-rc.4",
```
helfer: We'll update the package.json ourselves, this should not be in here.

Changelog.md
```diff
@@ -4,6 +4,9 @@ Expect active development and potentially significant breaking changes in the `0
 
 ### vNext
 
+### 1.0.0-rc.4
```
helfer: You can just put this under vNext, and we'll figure out which version we'll release it under.

src/graphql.tsx
```diff
@@ -573,8 +573,16 @@ export default function graphql(
...
             assign(data, this.previousData, currentResult.data);
           } else {
-            assign(data, currentResult.data);
-            this.previousData = currentResult.data;
+            let dataToMerge = currentResult.data;
+            // if we have an error, include any previously cached data
+            if (error) {
+              let previousResult = this.queryObservable.getLastResult();
+              if (previousResult && !previousResult.error) {
+                dataToMerge = assign({}, previousResult.data, currentResult.data);
```
helfer: I think it's pretty hard to predict what data you'd get here if there ever is any data in `currentResult`. I think it might be better to just return the previous result.

StevePotter: Hey @helfer thanks for the review.  I'm happy you're encouraged by this PR.  Please give me a few days to update the PR as I'm swamped on a few higher priority issues.  I'll get the appropriate testes in there and clean up the currentResult stuff and just use getLastResult().

Thanks again!

StevePotter:
@helfer I tightened up the code and added test coverage.  This one is good to go.

The test:
1. Run a successful query
2. Refetch, receiving an error
3. Verify the data from 1 is included alongside the error.

The test fails without my change to `src/graphql.tsx` and passes with the change.

StevePotter:
@helfer how about now?  It's clean, tested, and green.

StevePotter:
Another PR I submitted to apollo-client that doesn't affect this one but it needed for a solid offline UX https://github.com/apollographql/apollo-client/pull/1601

helfer:
Thanks @StevePotter, the test looks good!

This could technically break apps if a component only checks for data but not `error`, but I'm inclined to include it in the next minor release because we've been printing a warning to the console for quite some time now for unhandled errors, and I think this is a feature that a lot of people want.

PR #550, opened 2017-03-21
ApolloProvider: don't override store if not passed

### LINKED ISSUE #547: ApolloProvider/graphql should not put `store` on context
If you use Redux and react-apollo without wanting to share the store, `ApolloProvider` and `graphql` will put Apollo's `store` on the child context.

This overrides the `store` passed to react-redux's `Provider`, even if `Provider` is mounted inside `ApolloProvider`.

IMHO, react-apollo should only add `client` to context, and if you want access to Apollo's Redux store, you can store that on e.g. `client.store`. (even better: name the context `apollo`, and you can put whatever you like under that without clashing with other context provided by other projects)

wmertens:
Fixes #547

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 Expect active development and potentially significant breaking changes in the `0.x` track. We'll try to be diligent about releasing a `1.0` version in a timely fashion (ideally within 1 or 2 months), so that we can take advantage of SemVer to signify breaking changes from that point on.
 
 ### vNext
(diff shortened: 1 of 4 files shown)


jbaxleyiii:
@wmertens this looks great! If you can add a entry to the changelog that would be great!

wmertens:
@jbaxleyiii done.

jbaxleyiii:
@wmertens thanks!

PR #551, opened 2017-03-21
fix: Pass store prop from MockedProvider to ApolloProvider for parity

mmissey:
MockedProvider did not pass any `store` prop down to the ApolloProvider where it can be handled appropriately. This caused the MockedProvider to override the redux store and cause my tests to fail. 

This change simply passes it along so that ApolloProvider can use it like normal.
Also fixed a typo.


jbaxleyiii:
Thank you!

PR #561, opened 2017-03-23
chore(package): update @types/node to version 7.0.10

helfer:


helfer:
Was breaking before, seems to work now, so nothing to do.

Issue #450, opened 2017-02-03
Suggest feature: Debounce option in query operation.


zxlwbi199101:
### usage scenario
I use redux, and wrap MyListComponent by `graphql(queryStr, { options })`, when some options changes, it fetch from the server immediately. Have a look at 2 use scenarios: 

* there's a search input, display the result based on the search value dynamically,  when use is typing, graphWrapper fetches rapidly.
* the options are stored in the url, like `list?filter=name&sort=age&...`, the params are parsed in componentWillmount then dispatched to redux store. So when first loaded, the graphWrapper made a fetch with init options, then made another fetch with the dispatched values. So the first one is a waste.

I understand for both scenario is not a apollo specific problem, and can be fixed easily through _.debounce or wrap another component (maybe a little bit harder to debounce when binding the redux store value to that input). But I think the code will be more clean if apollo support a debounce option.

### what I have tried
* I wrote a apollo-client middleware, to make debounce at the network level.
```js
import _ from 'lodash';

const operationMap = {
  // operationName: {timer}
};

export default wait => ({ request, options }, next) => {
  // skip all the mutations
  if (_.get(request, 'query.definitions[0].operation') === 'mutation') {
    return next();
  }

  const operationName = request.operationName;

  let operationObject = operationMap[operationName];
  if (!operationObject) {
    operationObject = operationMap[operationName] = {};
  }

  // TODO: apollo-client uses promise to apply middleware,
  //       pending lots of promise will cause memory leak ??
  if (operationObject.timer > 0) {
    clearTimeout(operationObject.timer);
    operationObject.timer = -1;
  }

  operationObject.timer = setTimeout(() => {
    operationObject.timer = -1;
    next();
  }, wait);
};
```
Apparently this just debounce queries with their name, and if there are more than one queries to the same api, this middleware will cause bug. And if added a whitelist or blacklist to operationName, I don't think it's a graceful design.

### thoughts
* Is there any other mechanism can I apply such a debounce feature ?
* Is it feasible to add another level middleware / plugin (not network level) ?

helfer:
Hey @zxlwbi199101 and @ksmth. Thanks for the issue and PR! Here are my thoughts:

Even if it would be a little bit nicer to use if it's integrated with react-apollo, it's something that I think should clearly be kept outside because it can be done by simply wrapping the component.

[...]

PR #565, opened 2017-03-24
add `debounce` feature

ksmth:
TODO:

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
     "graphql-tag": "^1.3.1",
     "hoist-non-react-statics": "^1.2.0",
     "invariant": "^2.2.1",
+    "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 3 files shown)


ksmth:
Before updating the `CHANGELOG.md` or updating the docs, I'd like to discuss if this is even the right approach for adding the debounce feature. I have the same use case as described in #450 - the user is supposed to search via a search bar, but that is triggering one request per key stroke, which is not ideal.

helfer:
Thanks for the PR @ksmth! I've written down my thoughts on the issue. Let me know what you think. Maybe we can turn this into a PR to the docs that just documents how folks can implement this themselves with a wrapper. What do you think? Maybe recompose already has a wrapper like that?

jbaxleyiii:
@helfer @ksmth was the consensus made that this should be an external package?

PR #626, opened 2017-04-13
FIXES don't lock down React peer dependency to exact version

kimpers:
Version 1.0.2 changes the peer dependency to exactly `15.5.0`, this seems to have been due to a green keeper update and may not have been intentional. This new exact version lockdown causes a lot of conflicts with other dependencies and should be reverted.
TODO:
- [X] Update CHANGELOG.md with your change


stubailo:
This looks good, but we need to merge #628 first to make the tests pass I think.

stubailo:
Thanks!

Issue #574, opened 2017-03-27
Make weigh in report gzipped size


stubailo:
We should change this line to point at a gzipped size. https://github.com/apollographql/react-apollo/blob/2d5d5f1f796f4d39bb789f249a301cf21b37e12b/.travis.yml#L26

helfer:
#640 :tada:

PR #640, opened 2017-04-19
Add script to create gzipped file in development & report gzip size on weigh in

kevinzwhuang:
Description
----------

Currently, react-apollo uses `gzip-size` to get the gzip size without writing a `.gz` file to disk. `travis-weigh-in` requires an actual file to work with, making it incompatible with `gzip-size`. 

To solve this, this PR replaces `gzip-size` with a simple script that saves the compressed gzip and uses that file for file size calculation for both `npm run filesize` and `travis-weigh-in` scripts. 

The compression strategy utilized by `/scripts/gzip.js` is identical to what's used in `gzip-size`, with both using `zlib` set to level 9. The main change is that `gzip.js` creates a `.gz` file.

Also, the added `.gz` file in `/dist` might be helpful for people wanting to use gzipped versions of `react-apollo` in development.

Summary of changes:
----------

- Add a simple `gzip.js` script to `/scripts` to create compressed gzips (used by `compress:browser`)
- Add `compress:browser` script to `package.json` to handle creation of compressed `index.js.gz` for consumption by the `filesize` and `travis-weigh-in` scripts
- Append `npm run compress:browser` to the end of `compile:browser`
- Remove `gzip-size` as a devDependency

---------

This PR addresses #574

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
   - npm run compile
   - cd examples/create-react-app && npm test && cd ../../
   - npm run filesize
-  - python node_modules/travis-weigh-in/weigh_in.py ./dist/index.min.js || true # ignore size errors
+  - python node_modules/travis-weigh-in/weigh_in.py ./dist/index.min.js.gz || true # ignore size errors
 
 # Allow Travis tests to run in containers.
 sudo: false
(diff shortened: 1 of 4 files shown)


helfer:
This looks great. Thanks @kevinzwhuang !

PR #647, opened 2017-04-21
Add support for React Native 0.43

ligaz:
React Native 0.43 depends on React 16 alpha so we need to add it to the
version ranges.


nerdmed:
@stubailo @helfer if it helps: we are also running on rn43 with the new react version and everything seems to work fine with apollo

helfer:
Alright, let's go for it. Thanks @ligaz !

PR #650, opened 2017-04-23
Re-export everything from apollo-client

### LINKED ISSUE #521: Make sure all important things from Apollo Client are re-exported
Right now this package only re-exports `ApolloClient` and `createNetworkInterface`, but it should probably re-export a lot more things.

benhjames:
Fixes #521. :)

TODO:


thebigredgeek:
@helfer lgtm.  thoughts?

src/browser.ts
```diff
@@ -1,10 +1,9 @@
 export { default as ApolloProvider } from './ApolloProvider';
 export { default as graphql, withApollo, InjectedGraphQLProps } from './graphql';
 
-
 // expose easy way to join queries from redux
 export { compose } from 'redux';
 
 // re-exports of close dependencies.
-export { default as ApolloClient, createNetworkInterface, createBatchingNetworkInterface } from 'apollo-client';
+export * from 'apollo-client';
```
helfer: Will that export the default as ApolloClient? We should make sure that's the case. Otherwise we should just list all the things that we want to export. It's okay if there are a few things missing occasionally, because people can always import directly from apollo-client as a workaround.

benhjames: `ApolloClient` is also a named export from `apollo-client` (as well as the default export) so it should make no difference.

Ref: https://github.com/apollographql/apollo-client/blob/master/src/index.ts#L121

Issue #655, opened 2017-04-25
Prop-Types missing


Elliryanna:
Hey :)

i installed via npm the react-apollo "lib" and wanted to add this to my code, just like the tutorial says. But my npm compiler throw's the error that:

### Buggy Behavior
Error in ./~/react-apollo/lib/ApolloProvider.js
Module not found: 'prop-types' in "some local link stuff"\node_modules\react-apollo\lib

Error in ./~/react-apollo/lib/graphql.js
Module not found: 'prop-types' in "some local link stuff"\node_modules\react-apollo\lib

### Expected Behavior

Normaly the Prop-Types should be installed/downloaded via the NPM client...

### Version
- latest, i think 1.1.0

helfer:
Hi @Elliryanna, thanks for reporting! This should be fixed in 1.1.1 thanks to @thebigredgeek!

PR #656, opened 2017-04-25
added prop-types as dependency and removed as devDependency

thebigredgeek:


thebigredgeek:
Fixes https://github.com/apollographql/react-apollo/issues/655

We need to deprecate 1.1.0 with this

helfer:
Thanks a lot @thebigredgeek!

PR #664, opened 2017-04-28
lint

TAQ2:
Very simple. Doesn't seem appropriate to add my change to the CHANGELOG as I only linted the example project


helfer:
Thanks @TAQ2! You're right, no changelog entry is needed 🙂

Issue #170, opened 2016-08-25
`loading` stuck true after variable change


sgoll:
I have the following problem: when the props of my GraphQL-connected component change, and I use one of the props in the options (as variables), `loading` is stuck to the value `true` whenever both the previous value and the current value resolve to no result.

Here is an example:

``` js
import ApolloClient from 'apollo-client';
import gql from 'graphql-tag';
import React from 'react';
import { graphql } from 'react-apollo';

// Display props.
const Demo = (props) => (
  <pre>{JSON.stringify(props, null, 2)}</pre>
);

// Connect to some GraphQL query.
const ConnectedDemo = graphql(gql`
  query getStuff($name: String!) {
    stuff(name: $name) {
      payload
    }
  }
`, {
  options: ({ name }) => ({ variables: { name } }),
})(Demo);

// Set up client.
const client = new ApolloClient(/* ... */);

const render = (props) => {
  ReactDOM.render((
    <ApolloProvider client={client}>
      <ConnectedDemo {...props} />
    </ApolloProvider>
  ), document.getElementById('main'));
};

// Initial render.
render({
  name: 'does-not-exist',
});

// Prop update: fetch.
window.setTimeout(() => {
  render({
    name: 'nothing-either',
  });
}, 1000);
```

Here, during the initial render, everything is fine: `Demo` receives `{ loading: true, stuff: null }`. After some milliseconds, this is changed to `{ loading: false, stuff: null }`.

However, when the prop is updated, `Demo` receives `{ loading: true, stuff: null }` and it gets stuck there, even though the network request completed as did the first one.

The problem happens with `react-apollo` version 0.4.7 as well as 0.4.6. It does not happen whenever the first fetch did return an object (i.e. `stuff` was something other than `null`).

jbaxleyiii:
@sgoll I can't seem to replicate in a test? https://github.com/apollostack/react-apollo/pull/173

jbaxleyiii:
Can you create a repo or a failing test?

sgoll:
@jbaxleyiii Sure. I created a failing mock demo at https://github.com/sgoll/react-apollo-170. Please let me know if that helps.

jbaxleyiii:
Fixed with #191

hiei189:
I'm having this error after last update. had to downgrade.

SachaG:
[...]

PR #666, opened 2017-04-29
failing test for #170 

jbaxleyiii:
With the introduction of recycled queries, components that remount with different variables that get the same data (i.e. a null response from the server) get stuck in loading state.

This is currently just an error reproduction from https://github.com/comus/react-apollo-issue

TODO:

### DIFF
--- a/npm-debug.log.3540064644
+++ b/npm-debug.log.3540064644
@@ -1,0 +1,24 @@
+0 info it worked if it ends with ok
+1 verbose cli [ '/usr/bin/nodejs', '/usr/bin/npm', 'run', 'test--watch' ]
+2 info using npm@3.10.10
+3 info using node@v6.10.0
+4 verbose stack Error: missing script: test--watch
+4 verbose stack     at run (/usr/lib/node_modules/npm/lib/run-script.js:151:19)
+4 verbose stack     at /usr/lib/node_modules/npm/lib/run-script.js:61:5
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@helfer it's been a while 👋 but I'm trying to be an actual maintainer again! I merged a couple PR's and tackled a couple low hanging fruit. I need to get my local AC => RA setup working on windows before I could fix this one outright. 

Do you still have the tuesday meeting? How could I best help triage / add features/ contribute again?

thebigredgeek:
Well, at least we are able to lock it down to a test case now :D

helfer:
@jbaxleyiii I think this test could be broken. When you change variables it's expected that the component is first notified while data is still loading. To test for #170, we would have to make sure that the last time the handle is called, loading is false. I'm trying to build a reproduction for apollo-client, but I'm not having much luck because I think it involves a race condition that's hard to reproduce.

jbaxleyiii:
@helfer hmm, the test was written from behavior I noticed on one of the sample error templates. The route was changing (which was the variable, but the result of the two queries was the same (no person data loaded). Instead of showing no person loaded, it maintained a loading state.

helfer:
@jbaxleyiii sure, I don't doubt that. I'm just saying that it might not be wrong that the component first receives the props with `loading: true`, and then later goes to `loading: false`. The test should make sure that `loading` stays `true`.

Issue #557, opened 2017-03-22
ApolloProvider's shouldComponentUpdate causes silent issues when used with other providers, such as react router's


ryancole:
Does `ApolloProvider` need to implement `shouldComponentUpdate`?

I just updated my react-apollo version and suddenly my URL routing broke. Because I had only updated react-apollo, I started to track down what was now causing my application to no longer update when the URL changed (using react-router 4 for routing). After a bit of tinkering, I realized that if `ApolloProvider` was a parent component of react-router's `Router` then the routing would work, but if `Router` was the parent of `ApolloProvider` things broke.

So, I looked at git blame for `ApolloProvider`'s `shouldComponentUpdate` and see that it was added in a version more recent than the one I had been using. When `ApolloProvider` is a child of `Router` the entire routing breaks because `ApolloProvider` will silently refuse to update itself.

In a typical web application, it's almost guaranteed that there will be several "provider" components and none really specify how they need to be used in conjunction with other providers. I think most providers appear to skip implementing `shouldComponentUpdate` though because it can cause the whole application to never update.

Does `ApolloProvider` really need to implement this? If so, I think it should be noted in the docs, if it's not. Ideally somehow there could even be a runtime warning if it detects that you've got a `Router` child or something, but that's definitely too specific of a warning for this tool.

helfer:
@ryancole out of curiosity, can you try and see what happens when you remove `shouldComponentUpdate`? Which `react-apollo` tests break?

The PR that introduced it seems to have added it mostly out of caution.

ryancole:
Hmm. I'm trying to run the tests but it says there are no matching tests ...

```
PS C:\Users\Ryan\Projects\react-apollo> npm test

> react-apollo@1.0.0-rc.3 test C:\Users\Ryan\Projects\react-apollo
> npm run compile && jest


> react-apollo@1.0.0-rc.3 compile C:\Users\Ryan\Projects\react-apollo
> tsc

No tests found
  2258 files checked.
  roots: C:\Users\Ryan\Projects\react-apollo - 2258 matches
  testMatch:  - 2258 matches
  testPathIgnorePatterns: \\node_modules\\ - 65 matches
  testRegex: C:\Users\Ryan\Projects\react-apollo\test\.*.test.(ts|tsx|js)$ - 0 matches
----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
[...]

PR #669, opened 2017-04-30
fix(shouldComponentUpdate): removed from ApolloProvider

CodogoFreddie:
Is causing issues when `ApolloProvider` is used with other Providers (eg: react-router).
`shouldComponentUpdate` is an unnecessary here, it is used to stop unnecessary updates caused by higher components. But as `ApolloProvider` should be near the top of the tree, it can be assumed that ALL updates passed to is are necessary.

fixes issue: #557

### DIFF
--- a/src/ApolloProvider.tsx
+++ b/src/ApolloProvider.tsx
@@ -51,12 +51,6 @@ export default class ApolloProvider extends Component<ProviderProps, any> {
     }
   }
 
-  shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@CodogoFreddie this is a great point! And all the test pass! I'm great with this. Can you update the changelog and I'll merge it in!

jbaxleyiii:
@CodogoFreddie I'll update it actually so we can do a release! Thanks again!

PR #679, opened 2017-05-02
move type imports to root of AC for easier local dev / ci testing

jbaxleyiii:
Moved to using root import for types from AC. Needs to wait on https://github.com/apollographql/apollo-client/pull/1644


benhjames:
Beat me to it, thanks @jbaxleyiii! :tada:

Issue #328, opened 2016-11-15
Can't get full markup from "renderToStringWithData"


novus57:
I can't get full page markup, only part without queries. Intial state come with data but markup doesn't full. Sorry for my English. Thanks a lot for help.
```js
app.use((req, res) => {
  const memoryHistory = createMemoryHistory(req.url);
  const user = Object.assign({}, { current: req.user });
  if (user.current) {
    delete user.current.password;
  }

  const apolloClient = new ApolloClient({
    ssrMode: true,
    // Remember that this is the interface the SSR server will use to connect to the
    // API server, so we need to ensure it isn't firewalled, etc
    networkInterface: createNetworkInterface({
      uri: 'http://localhost:3000/graphql',
      credentials: 'same-origin',
      // transfer request headers to networkInterface so that they're
      // accessible to proxy server
      // Addresses this issue: https://github.com/matthew-andrews/isomorphic-fetch/issues/83
      headers: req.headers,
    }),
  });

  const initialState = { user };
  const store = configureStore(memoryHistory, initialState, apolloClient);
  const history = syncHistoryWithStore(memoryHistory, store);
  match(
    { history, routes: createRoutes(store), location: req.originalUrl },
    (error, redirectLocation, renderProps) => {
      if (error) {
        res.status(500).send(error.message);
      } else if (redirectLocation) {
        res.redirect(302, redirectLocation.pathname + redirectLocation.search);
      } else if (renderProps) {
        const component = (
          <ApolloProvider store={store} client={apolloClient}>
            <MuiThemeProvider muiTheme={getMuiTheme(null, { userAgent: req.headers['user-agent'] })}>
              <RouterContext {...renderProps} />
            </MuiThemeProvider>
          </ApolloProvider>
        );

        renderToStringWithData(component).then((content) => {
          res.send(renderInitialView(content.markup, content.initialState));
        }).catch(dataError => winston.error('RENDERING ERROR:', dataError));
      }
    }
  );
});
```

tmeasday:
Hi @novus57, which version of RA are you using?

amannn:
[...]

PR #688, opened 2017-05-04
Fix server side rendering when using the `cache-and-network` fetchPolicy

amannn:
As discussed in #328.

I've added a test that passes now with the adapted logic in `graphql.tsx`. I'm not quite confident if it is the right way to go though, so I'd appreciate a review 🙂.

TODO:


jbaxleyiii:
@amannn this is great! Thank you!

PR #736, opened 2017-05-30
added sugar for graphql

jbaxleyiii:
Is this useful to anyone? Does it make things more confusing?

cc @helfer @stubailo 

TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,7 +6,7 @@
 #### BREAKING FOR TYPESCRIPT USERS
 - Feature: Enhanced typescript definitions to allow for more valid type checking of graphql HOC [PR #695](https://github.com/apollographql/react-apollo/pull/695)
 - Feature: Flow types: [PR #695](https://github.com/apollographql/react-apollo/pull/695)
-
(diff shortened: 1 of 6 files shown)


stubailo:
Hmm this is odd, we should at least make sure it validates the correct type of operation

jbaxleyiii:
@stubailo haha for sure! I just wanted to push this out to see if it was actually useful or if it would make things more confusing (greater surface area) for people

stubailo:
I do think this makes sense now that I think about it since what you actually get in the props is so different for mutations vs queries.

eug48:
These might be useful to further improve the great work done recently on TypeScript support. In particular some common properties are currently optional (e.g. `data` & `mutate`) so strictNullChecks needs to be needlessly overridden (i.e. in the `props` function have to use `data!.loading` instead of `data.loading`).

jbaxleyiii:
@eug48 @stubailo I'm going to continue to pursue this direction! However, with the upcoming prettier PR, plus a few others, this will be out of date and there wasn't really any meat here anyway. I'm going to close it and revisit it soon!

Issue #718, opened 2017-05-18
Synchronously re-mounting a graphql-wrapped component gets it stuck on `loading`


dallonf:
This seems related to #170, but I think it'll be something much simpler to solve.

### Steps to Reproduce

Here's a sample project: https://github.com/dallonf/react-apollo-bug-sync-remount

In a nutshell, a `graphql`-wrapped component is rendered. On user input, the component is unmounted and replaced by another component that immediately, in `componentDidMount`, triggers a re-render that causes the `graphql` component to be re-mounted.

In real-world usage, this type of scenario can be caused by React Router's `<Redirect />` component, for example.

### Buggy Behavior

The `graphql`-wrapped component's `data.loading` prop will be stuck on `true`, even though the network request has been completed, and the query is correctly updated in the Redux store.

### Expected Behavior

The `graphql`-wrapped  component should behave as it did on its initial mount, where `loading` will eventually be set to `false` and its data correctly updated.

### Some more research

The difference between a "stuck" component and a working one seems to be that the `APOLLO_QUERY_STOP` event triggered when it mounted, which effectively resets the recycled query. I'm not sure exactly why this event is so important, but I do know roughly why it isn't being called.

This event is triggered when you unsubscribe from a query; specifically, the `tearDownQuery` call in the `ObservableQuery.prototype.onSubscribe()` method. This only happens when there are _no more observers_ on the query. In this scenario with a synchronous re-mount of a component, there is no point where there are no observers. Here's what happens instead:

1. Component mounts, triggering an observer, let's call it **subscribe1**. There is now 1 observer active.
2. Component loads its data, everything is fine
3. Component unmounts
4. `graphql` starts to recycle the query, calling `setOptions`, which in turn calls `setVariables` and `result`, which ultimately sets up an observer. Let's call it **setOptions**. There are now 2 observers active.
5. As part of the recycling process, a dummy observer is also created, to keep the query alive. Let's call it **recycle**. There are now 3 observers active.
6. Finally, the component unsubscribes from **subscribe1** to finalize the unmounting process. There are now 2 observers active.
7. Component re-mounts
8. Seeing that there is a recycled query available, the component picks it up and tears down its dummy observer, **recycle**. There is now 1 observer active.
9. _Uh oh! This is the part where `APOLLO_QUERY_STOP` should have been called, but there's still an observer active: **setOptions**_
[...]

Issue #667, opened 2017-04-30
networkStatus gets stuck on "2" (setVariables)


SachaG:
Something I've noticed recently: when passing new options to my query container, it can get stuck on `networkStatus = 2` even once the request succeeds. Here's a short screencast to demonstrate the issue:

https://cl.ly/kEWd

Note that this only seems to happen the first time these new options are passed. If I go back and forth between both pages, it'll work the second time. This is with apollo-client 1.1.2 and react-apollo 1.1.1.

Other (maybe related) `networkStatus` issues: 

https://github.com/apollographql/apollo-client/issues/1186
https://github.com/apollographql/react-apollo/issues/170

grossberg:
i have the same issue

Josiang:
i also have the exact same issue

helfer:
Thanks @SachaG! Would anyone be able to provide a reproduction using react-apollo-error-template? 🙂

hiei189:
Is there a workaround? like using another version?

alikhani:
I'm having the same issue.. Any progress?

helfer:
@hiei189  @alikhani I think this issue is similar to another one we're trying to debug (where the loading state doesn't change back), but we could still use a reproduction for this, in case it's not the same.

andreasgruenh:
Hello everybody, I encountered this problem as well and I was able to reproduce this in this repo: https://github.com/andreasgruenh/react-apollo-error-template

Take a look at the PersonDisplay component. It receives an ID as prop and should render the corresponding person. To prevent some old state in the PersonDisplay component, I want it to completly unmount and remount it (by giving it a key prop). When I give this prop from the App component, everything works, when I hand this prop from mapProps HOC, it stops working. It also does not happen with react-apollo 1.0.0

alikhani:
I can try to reproduce it later when I have time. I can confirm that the issue manifests itself in react-apollo@1.0.1. So 1.0.0 works fine.

helfer:
@alikhani what's the latest version of apollo-client (not react-apollo) that it works with and the first version where it breaks?

alikhani:
@helfer I'm importing apollo-client from react-apollo. Not sure how the dependencies work.. 
react-apollo@1.0.0 that works for me has the dependency: apollo-client "^1.0.0-rc.9". In my node_modules the version I find is apollo-client 1.2.2.

[...]

PR #740, opened 2017-06-01
recycled query teardown

jbaxleyiii:
Fix for #718, #667, and probably a lot of other loading issues by removing the current subscription before recycling the observable 

I'm going to add a test in the morning for this, but for all those impacted, this will be released tomorrow after the flow PR is released in AC. 👍 

TODO:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,7 +6,7 @@
 #### BREAKING FOR TYPESCRIPT USERS
 - Feature: Enhanced typescript definitions to allow for more valid type checking of graphql HOC [PR #695](https://github.com/apollographql/react-apollo/pull/695)
 - Feature: Flow types: [PR #695](https://github.com/apollographql/react-apollo/pull/695)
-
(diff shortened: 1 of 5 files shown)


jbaxleyiii:
@helfer we do have tests, but I'm going to work to improve them some. I think this change isn't the full fix we want and may require apollo-client changes. With this change, on remount you get a loading state then the data. Ideally it shouldn't pass loading state.

One question for thought though:
If you contents from your query, is the query loading? If data means no loading, then we could return loading unless we have the data in which case we could set loading to false

jbaxleyiii:
This will fail until apollo-client PR is approved and released

munyirik:
@jbaxleyiii could you point me to the commit or PR in apollo-client that coincided with this fix? I'm running into https://github.com/apollographql/react-apollo/issues/667 on an apollo-client fork. I'm hoping I can apply the same fix.

Issue #745, opened 2017-06-02
Importing MockedProvider in 1.4.0


anthonator:
When using `react-apollo` 1.2.0 I was able to import `MockedProvider` like:

```js
import { MockedProvider } from 'react-apollo/lib/test-utils';
```

Doing this in 1.4.0 causes the following error:

```
import * as React from 'react';
^^^^^^

SyntaxError: Unexpected token import
```

I'm using `mocha` and `babel` and wasn't having issues prior to 1.4.0.

**.babelrc**

```
{
  "presets": ["env", "react", "stage-2"],
  "env": {
    "test": {
      "plugins": [
        "dynamic-import-node"
      ]
    }
  }
}
```

**mocha.opts**

```
--compilers js:babel-register
--recursive
--require ./test/setup
--require babel-polyfill
--require jsdom-global/register
src/**/*.test.js
src/**/*.test.jsx
```

anthonator:
If I import `MockedProvider` with the following:

```js
import { MockedProvider } from 'react-apollo';
```

I get the following error:

```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of `WrapperComponent`.
```

Here's the code causing this:

```js
mountedContainer = mount(
  <MockedProvider store={ store }>
    <FormContainer { ...props } />
  </MockedProvider>,
);
```

Again, this was working properly prior to 1.4.0.

jbaxleyiii:
I'm so sorry about this, this was most likely broken with the move to tree-shaking bundle. I will have a fix out tonight for this. 

Thanks for the report!

jbaxleyiii:
fixed in 1.4.1! (which is going to npm now!)

PhilippSpo:
@anthonator Is this issue resolved for you? We still see this issue in 1.4.3 right now. Not sure If we are doing something wrong 😅

anthonator:
@PhilippSpo yes, this was fixed for me

PR #748, opened 2017-06-03
Fix test utils

jbaxleyiii:
Fix for #745


tdeekens:
Hi, thanks a lot for the fix. However this still breaks for us by importing

`import { MockedProvider } from 'react-apollo/lib/test-utils';`

We get an error due to non transpiled code at 

```js
react-apollo/lib/graphql.js:19

import { Component, createElement, } from 'react';
^^^^^^
SyntaxError: Unexpected token import
```

as the ES bundle seems to require `graphql` in it's UMD factory function. Any thoughts on how to avoid this?

Issue #750, opened 2017-06-03
interface MutationOpts lacks some properties


pawsong:
### Buggy Behavior

I found [`MutationOpts`](https://github.com/apollographql/react-apollo/blob/ddd3d8faabf135dca691d20ce8ab0bc24ccc414e/src/graphql.tsx#L38) does not contains `refetchQueries`, which is documented [here](http://dev.apollodata.com/react/api-mutations.html#graphql-mutation-options-refetchQueries) and already in `apollo-client`'s [`MutationOptions`](https://github.com/apollographql/apollo-client/blob/ca17c34dc6f6da3eba682eadd166f82c3913692e/src/core/watchQueryOptions.ts#L105). Is there any reason to redefine mutation option typing in `react-apollo`?

### Version
- apollo-client@1.4.0
- react-apollo@1.4.1

beagleknight:
+1 to this. The `MutationOpts` doesn't contain `update` as well like the documentation suggests:
http://dev.apollodata.com/react/cache-updates.html#directAccess

helfer:
This was probably just an oversight. A PR to fix it would be appreciated! 🙂

beagleknight:
@helfer @pawsong This is already fixed here: https://github.com/apollographql/react-apollo/pull/770

PR #770, opened 2017-06-12
Missing properties in MutationOpts (fix for #750)

eug48:
These are based on definitions from [apollo-client](https://github.com/apollographql/apollo-client/blob/ca17c34dc6f6da3eba682eadd166f82c3913692e/src/core/watchQueryOptions.ts#L143).


jbaxleyiii:
@eug48 would you be able to update the flow types as well?

eug48:
@jbaxleyiii Sure, I've had a go. I don't use flow personally but `npm run flow-check` still passes.

jbaxleyiii:
@eug48 thank you so much! I'll update the changelog once this is merged through!

Issue #776, opened 2017-06-14
data.networkStatus is defined in absence of options.notifyOnNetworkStatusChange


lorensr:
## Steps to Reproduce

https://codesandbox.io/s/N2DMB7pv

Displays `props.data.networkStatus` on the page. 

### Buggy Behavior

On load, starts as `1` and changes to `7`. 

### Expected Behavior

Should instead be `undefined`. (As discussed [here](https://github.com/apollographql/react-apollo/issues/753#issuecomment-306098020))

### Version
- apollo-client@1.2.2
- react-apollo@1.2.0

helfer:
This is not a bug, it's currently the intended behavior. Happy to talk about changing it if it's a real issue.

lorensr:
Okay! Idc, was just following up on Sashko's comment 😊

helfer:
Just to be clear, I think it would be better if we changed it, but that doesn't change the fact that it's currently the intended behavior 😄

PR #777, opened 2017-06-14
Include CodeSandbox template in CONTRIBUTING

lorensr:
Thoughts? It was useful eg in #776


helfer:
Hey @lorensr, i think that's a good idea! Can you also include a link to react-apollo-error-template for more complicated bugs?

stubailo:
Tweet about this when merged!

PR #819, opened 2017-07-05
run prettier on all files

jbaxleyiii:
MUCH PRETTY

But really, this PR adds support for using prettier for formatting of all library code. Thanks to the work by @vjeux, @stubailo, and others, this PR formats typescript and inline gql tags!

It also sets up a post commit hook to format prior to committing to ensure consistency


jbaxleyiii:
Note to self, fix flaky mobx test

Issue #801, opened 2017-06-27
Server-returned GraphQL error does not reset data on component


alexzielenski:
## Information
I'm running a query along the lines of:

```graphql
query BoardDetail($id: ID!) {
    board(id: $id) {
        id
        status
        name
    }
}
```

I'm also using react router to page between a component which lists such boards and links to the detail page. For one of my board objects my `status` field result is incorrectly returned by the server (it is a GraphQL enum type and I'm returning an invalid value). 

## Steps to Reproduce

1. Start at board list
2. select a board whose `status` field is valid and has no error. It's page renders fine.
2. Go back to board list.
3. Select a board whose `status` field is invalid and returns a GraphQL error. It's page displays the error and shows the board from step 2.

### Buggy Behavior

`props.data.board` is populated with the result of a past render while `props.data.error` is populated with the error for the current query. This puts my UI in an inconsistent state showing both an error and a successful render of a board viewed in the past. I know that I could do something like `if (data.error != null) {...} else { <Board /> }` but what if I want to display an error at the same time as receiving cache data?

I think this problem stems from some behavior Apollo has that when a query fails it reverts to a past result to use as data for the component. **This is fine except for when we use variables in our queries.** When I run my query in the devtools with "Load from Cache" enabled and correct `id` variable, I get back the (stale) data that I want.

### Expected Behavior

`props.data.board` is `null` or cache-loaded (with correct variables) and `props.data.error` is non-null as the server returns.

### Version
- apollo-client@1.6.0
- react-apollo@1.4.2

helfer:
Hey @alexzielenski, thanks for filing an issue. You're right that having old data along with the new error is inconsistent. The reason for it is that Apollo Client currently doesn't put errors in the store, which also means that it won't return data and errors at the same time. If there's an error, it only returns the error. In the past this has led to an inconvenient situation for people where they lose all the data in their component when an error occurs, so we decided to accept a PR that provided the old data with the new error.
What I recommend doing is simply not displaying the data if there's any error, like you suggested yourself. We're working on a new store that will have more intuitive behavior around errors and pass them through to components along with the new data.

alexzielenski:
Hi @helfer! Thanks for your rapid response.

Would it be feasible to change react-apollo in the interim so it runs your query on the cache again, rather than simply returning the last-rendered data by that component? This would be acceptable to me, since then my detail page would not display the wrong document; but a stale version.

helfer:
[...]

Issue #807, opened 2017-06-29
Component won´t rerender with new data after client.resetStore()


piotrblasiak:
Love apollo, but I have this problem...

## Steps to Reproduce

1. Render component:

```
@graphql(gql`
  {
    me {
      id
      name
    }
  }
`)
class MyComponent extends PureComponent { ... }
```
2. Call apolloClient.resetStore()

### Buggy Behavior

The component is not rerendered with a clared store. In this case I had logged out the user so the query should give me {me: null} but it still gives me the old object.

### Expected Behavior

It should give me a null object for "me" since I´ve logged out and cleared the store like the docs say.

### Version
- react-apollo@1.4.2

jbaxleyiii:
Bringing in an issue from AC to see if we can improve this:


**Intended outcome:**

resetStore clears the Apollo data

**Actual outcome:**
resetStore only refetches, there is no point where data in components is cleared


**How to reproduce the issue:**

error repo: https://github.com/stantoncbradley/apollo-reset-store
1. Run the app :)
2. Clicking resetStore button triggers Apollo resetStore
3. App logs nextProps as the come in through componentWillReceiveProps
4. At no point in the logs is the data cleared

Our actual use case is a user signing out of our app.  We noticed if another user logs in and one of the queries fails, data from the last user displays!  It was never cleared!!!!

stantoncbradley:
really like to see this get fixed!  it's imperative that we completely clear our user data after the user logs out!!! 🙏

stantoncbradley:
this is still a big issue.  We are locked into `react-apollo "1.1.3"` so we don't accidentally leak private data until this issue is resovled

stantoncbradley:
@jbaxleyiii this is still an issue as far as I'm aware.  related to #858 and #890 too

mitjade:
Still and issue also in apollo-client 2.0 beta.

mitjade:
Please add this to https://github.com/apollographql/apollo-client/milestone/9.

tsnieman:
I believe this is still a problem in 2.0 and should be re-opened.

stantoncbradley:
this is a really big issue and I'm afraid it's not well known, nor getting the attention it deserves. To be clear, react-apollo isn't properly clearing user data and there are scenarios where react-apollo can leak (sensitive) data after the store has been "cleared".  If more people realized they have this known issue in production, I don't think they would be happy with Apollo

jbaxleyiii:
[...]

PR #858, opened 2017-07-17
Suggestion: Don't return last result, return actual cache data

alexzielenski:
This is a quick code change suggestion from #801. It brings the code more in line with what the actual comment says is the writer's intent. We will simply re-run the same query on the store, with fresh options.

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -622,7 +622,11 @@ export default function graphql<
             assign(data, this.previousData, currentResult.data);
           } else if (error) {
             // if there is error, use any previously cached data
-            assign(data, (this.queryObservable.getLastResult() || {}).data);
+            try {
+              assign(data, this.store.readQuery({ query: document, opts }));
+            } catch (exception) {
+              // raised if possibly no data. do nothing. 
+            }
           } else {
             assign(data, currentResult.data);
             this.previousData = currentResult.data;
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -625,7 +625,7 @@ export default function graphql<
             try {
               assign(data, this.client.readQuery({ query: document, variables: opts.variables }));
             } catch (exception) {
-              // raised if possibly no data. do nothing. 
+              // raised if possibly no data. do nothing.
             }
           } else {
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@alexzielenski this is a great change to be more inline with stock Apollo-client usage! Thank you! Would you be able to write a test so we can verify this is happening?

If you have any questions, I'm happy to help!

alexzielenski:
@jbaxleyiii No thanks. I identified this as an issue almost a month ago and was brushed off (I suggested this exact fix in my first post but I guess no one really read it) I just got tired seeing this bug in my project. This just seemed like the only way I could possibly express what's wrong with the current implementation to you guys since you were closing my issue.

jbaxleyiii:
@alexzielenski I'm sorry to hear that you have a bad experience suggesting improvements to Apollo. The past month has seen some changes to our team and honestly, we have always been small and relied on community support. This entire library was developed by contributors so I really hate to see that you didn't have a positive experience.

I'm happy to write a test case for this and get it into an upcoming release. If there is anything that I can do to improve your experience next time please don't hesitate to let me know.

All the best 👍

akomm:
This is obviously a buggy behavior and should be fixed. I have another bug, which make this one leak data and there might be other situation where this leads to leaked data.

1. After authentication user information is loaded via GraphQL. After logout I clear all state from redux (apollo is run in same store and is also cleared -> confirmed via dev tools)

2. After this user logs out, when another user wants to login an an error is raised while fetching new data, the data from last query is loaded, even though the store was cleared.

stantoncbradley:
would this pertain to #807 at all?  these cases around logging in and seeing old users data is really concerning

lourenco:
Hi, the project which I work on is being directly affected by this bug. Please, approve this pull request.

marlonmleite:
Hey, in my company have the same problem. 
It's bad for user and my QA report many bugs. 
It's necessary to do several jobs of client.resetStore () to all code in a way not cause the problem.

This is very bad :(

Ardhimas:
Why was this not merged? It looks like a good PR.

PR #865, opened 2017-07-19
Fix/804

### LINKED ISSUE #804: getWrappedInstance returns undefined, when the skip option resolves true
## Steps to Reproduce

create:
```js
import {graphql, gql, compose} from 'react-apollo';

const Comp = () => (<div>dummy</div>);
const Composed = compose(
  graphql(gql`query ...`, {withRef: true, skip: true}),
  graphql(gql`query ...`, {withRef: true}),
)(Comp);
```
use it with ref:
```js
import Composed from './Composed';
import React from 'react';

class App extends React.PureComponent {
  render() {
    return (
      <div>
        <button onClick={() => console.log(this.refs.composedRef.getWrappedInstance())}></button>
        <Composed ref="composedRef" />
      </div>
    );
  }
};
```
### Buggy Behavior

getWrappedInstance returns undefined. When checking with react dev tools, I can see that the first graphql HOC has a wrappedInstance ref, but the second one does not.

### Expected Behavior

ref should be propagated down the HOC chain and getWrappedInstance should return the underlying component.

### Version
- apollo-client@^1.4.0
- react-apollo@^1.4.0

akomm:
Test & fix for issue #804

fixes #804

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ### vNext
 
+- Fix: Fix issue where `withRef`-option of `graphql` did not work when the query was skipped [#865](https://github.com/apollographql/react-apollo/pull/865)
+
 ### 1.4.4
 - Fix: Fix issue around hoisting non react statics for RN [#859](https://github.com/apollographql/react-apollo/pull/859)
 - Fix: Fix issue where options was called even though skip was present [#859](https://github.com/apollographql/react-apollo/pull/859)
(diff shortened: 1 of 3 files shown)


akomm:
Same test passed locally. Weird. must be environment-dependent. Looks like the console.error mock did not work. I see the expected console.error, which is actually expected in the test, coming shortly before the failed test.

jbaxleyiii:
@akomm I'm afraid the tests have been flaky of late. I need to look into why

jbaxleyiii:
@akomm thank you for this great contribution! I'll merge and release it asap!

PR #888, opened 2017-07-23
update example to react-apollo 1.4.8

bossbossk20:
update example to react-apollo 1.4.8 

remove unused package such as gql, apollo-client 

added react-apollo


jbaxleyiii:
@bossbossk20 the example apps actually imported code from the library itself instead of npm. This is done to run tests in example applications during each PR.

PR #920, opened 2017-08-01
chore(deps): Add timezone to Renovate config

rarkins:
Renovate is currently configured to use the schedule "after 10pm and before 5am on every weekday" however without a configured "timezone" then these times will default to UTC. This PR adds the America/Los_Angeles IANA time zone to the `renovate.json` configuration file, assuming that's the preferred timezone for this schedule.


jbaxleyiii:
@rarkins you have been SO helpful with setting up renovate! Thank you SO much! It is incredible and I'm so excited to move all of Apollo to use it

Issue #674, opened 2017-05-01
Unable to get mockedData into props from MockedProvider in test


jasonmorita:
Following some ideas from this gist: https://gist.github.com/jbaxleyiii/b3634aeeab7bdb80ed4119ea5a07ba4a, I am unable to get a component test to work as expected.

Everything seems to work except the `props` never get populated with the `data` from the response.

Any reason why this isn't working as I expect it to?

Here is the simplest real-world example I could make based on an existing component:

```
// test-component.js
import React from 'react';
import { graphql } from 'react-apollo';
import userQuery from './userQuery.graphql';

function TestComponent (props) {
    console.log('props from component', props); // eslint-disable-line
    return <div>The component...</div>;
}

export default graphql(userQuery)(TestComponent);

// userQuery.graphql
query userQuery {
    user {
        items {
            firstName
            id
            lastName
        }
    }
}

// test-component-test.js
import React from 'react';
import { MockedProvider } from 'react-apollo/lib/test-utils';
import { mount } from 'enzyme';
import toJson from 'enzyme-to-json';
import userQuery from '../userQuery.graphql';
import { TestComponent } from '../';

const mockedData = {
    user: {
        items: [
            {
                firstName: 'userF',
                id: 'hhhh-gggg-iiii',
                lastName: 'userL',
            }
        ]
    }
};

describe('<TestComponent />', () => {
    describe('Markup should stay consistent', () => {
        const component = (
            <MockedProvider
                mocks={[
                    {
                        request: {
                            query: userQuery
                        },
                        result: { data: mockedData }
                    },
                ]}
                store={{
                    getState: () => {},
                    dispatch: () => {},
                    subscribe: () => {},
                }}
            >
                <TestComponent />
            </MockedProvider>
        );

        const mountedComponent = mount(component);

        it('should not have any changes without a new snapshot', () => {
            const tree = toJson(mountedComponent);
            expect(tree).toMatchSnapshot();
        });
    });
});

// console.log output during lifecycle
props from component { data:
 { variables: { offset: null, limit: null },
   refetch: [Function: bound ],
   fetchMore: [Function: bound ],
   updateQuery: [Function: bound ],
   startPolling: [Function: bound ],
   stopPolling: [Function: bound ],
   subscribeToMore: [Function: bound ],
   loading: true,
   networkStatus: 1,
   error: [Getter] } }

props from component { data:
 { variables: { offset: null, limit: null },
   refetch: [Function: bound ],
[...]

PR #973, opened 2017-08-11
MockNetworkInterface match mock requests regardless of variable order

L1fescape:
Currently mock data needs to exactly match the request being made within a component being tested. This is because requests object are [stringified](https://github.com/apollographql/react-apollo/blob/master/src/test-utils.tsx#L259-L266) and used as keys when [registering](https://github.com/apollographql/react-apollo/blob/master/src/test-utils.tsx#L121) and [looking up](https://github.com/apollographql/react-apollo/blob/master/src/test-utils.tsx#L138) the corresponding response objects to be returned. This change sorts the request object when it is stringified so the order of variables doesn't matter.

This issue is discussed in #674

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNext
+- MockNetworkInterface match mock requests regardless of variable order [#973](https://github.com/apollographql/react-apollo/pull/973)
 
 ### 1.4.11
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@L1fescape thank you! This looks great to me!

PR #1001, opened 2017-08-19
Remove typenames from MockedProvider

Carlows:
This PR allows to pass a removeTypenames prop to the MockedProvider, this way there's no need to add the __typename field to the mocked data.

Right now, to add mocked data in tests, it's required to use `addTypenameToDocument` in all of the queries:

```
const query = addTypenameToDocument(HERO_QUERY);
```

Together with this, the mocked data needs to include the __typename field in all of the models:

```
export const hero_no_friends = {
  __typename: 'Droid',
  name: 'r2d2',
  id: '1',
  friends: null,
};
```

I find this a bit of a pain, and I would prefer not having to do this. As the apollo-client accepts an `addTypename` field, we can allow the MockedProvider to configure that option as we'd like.

This can be used this like this:

```
<MockedProvider mocks={mocks} removeTypename>
```

I didn't find any tests for test-utils, so I only added an example test within the base example.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 - Feature: Support apollo-client 2.0
 - Fix: Scope query recyclers by client [#876](https://github.com/apollographql/react-apollo/pull/876)
 - MockNetworkInterface match mock requests regardless of variable order [#973](https://github.com/apollographql/react-apollo/pull/973)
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@Carlows this is such a great addition 😍 

Thank you! And congrats on your first contribution here!

PR #1025, opened 2017-08-24
1005

### LINKED ISSUE #1005: Bug: getClient calls `options` with with current props
When `skip` transitions from `true` to `false` due to a prop change, the `options` method is called several times. The times it is called to retrieve for example the variables for the query it is called with the components `nextProps`. But the internal `getClient` method calls it with `this.props`. As this hasn't been brought up as of yet, I recon nobody has used `skip` in the way I did.

Take for example this options object: 

```js
{
  options: ownProps => ({
    variables: {
      gender: ownProps.person.gender,
    },
  }),
  skip: ownProps => !ownProps.person,
}
```
So as long as there's no `person` property available I don't want to execute the query.

But the moment the `person` property is set, `options` is called several times. Not sure of the order, but at least for `skip`, `variables` and `client`. When it is called for `skip` or `variables`, the `nextProps` are passed, and `ownProps.person.gender` will return the persons gender.

But when options is called for the client, `this.props` is passed, which at that point does not yet contain the `person` prop, and thus accessing `ownProps.person.gender` will fail with an error.

**Intended outcome:**

For the client, `options` should also be called with the value of `nextProps` 

**Actual outcome:**

App throws an error: `Uncaught TypeError: Cannot read property 'gender' of undefined`
image

I've created a minimal example demonstrating the issue using the error-template (note the branch):
Just run it with the developer console open and the problem will present itself.

https://github.com/koenpunt/react-apollo-error-template/tree/get-client-prev-props

**Version**
- apollo-client@1.9.1
- react-apollo@1.4.14

jbaxleyiii:
Fixes #1005 

### Checklist:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # Change log
 
 ### vNext
-- Fix: ensure options gets up to date pros [#1025](https://github.com/apollographql/react-apollo/pull/1005)
+- Fix: ensure options gets up to date props [#1025](https://github.com/apollographql/react-apollo/pull/1005)
 - Fix: ensure queryRecycler exists before using it
 - MockNetworkInterface match mock requests regardless of variable order [#973](https://github.com/apollographql/react-apollo/pull/973)
 - Allow to pass removeTypenames to MockedProvider [#1001](https://github.com/apollographql/react-apollo/pull/1001)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNext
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNext
+- Fix: ensure options gets up to date pros [#1025](https://github.com/apollographql/react-apollo/pull/1005)
```
koenpunt: That was quick service ;) You have a small typo here tho, pros => props

PR #1027, opened 2017-08-24
1007

### LINKED ISSUE #1007: Possible Race Condition
Hello - just talked to @jbaxleyiii on Slack and he asked me to open this issue.

I've SSR'd my app and i need to refetch a Header component once it loads in the client, since that is when I have the JWT.

The problem I'm running into is that if I call refetch() on componentDidMount, the user data doesn't make it into props. This is easily fixed by setting a 1ms timeout, which makes it work. 

In both cases - with and without the setTimeout - the network panel shows the correct data coming back, so this isn't an issue with the token being grabbed before it's available. 

I very well may be doing something wrong here - please let me know if that is the case.

Here is the code:

```js
import { Component } from 'react'
import { graphql, compose } from 'react-apollo'
import { CURRENT_USER_QUERY } from '../queries';

class Header extends Component {

  componentDidMount() {
    // This fetches the new data, but doesn't populate the user via props
    // this.props.currentUserQuery.refetch();
    // This fetches the new data, and populates the user via props
    setTimeout(this.props.currentUserQuery.refetch, 1);
  }

  render() {
    console.log(this.props.currentUserQuery.user);
    const user = this.props.currentUserQuery.user || {};
    const { email = '' } = user;

    return (
      <div>
        <p>{email} I'm the header!</p>
      </div>
    )
  }
}

const userEnhancer = graphql(CURRENT_USER_QUERY, { name: 'currentUserQuery' });
export default compose(userEnhancer)(Header)
```

**Version**
- react-apollo@1.4.14

jbaxleyiii:
Fixes #1007 

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNext
+- Fix: handle calling refetch in child componentDidMount
 - Fix: ensure options gets up to date pros [#1025](https://github.com/apollographql/react-apollo/pull/1005)
 - Fix: ensure queryRecycler exists before using it
 - MockNetworkInterface match mock requests regardless of variable order [#973](https://github.com/apollographql/react-apollo/pull/973)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNext
+- Fix: ensure options gets up to date pros [#1025](https://github.com/apollographql/react-apollo/pull/1005)
 - Fix: ensure queryRecycler exists before using it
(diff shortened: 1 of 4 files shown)


kdong007:
@jbaxleyiii could you explain the actual purpose of 'refetch'? At first I thought its is a retry handle for things like network error.

eg
```
render(){
  return ....
    {this.props.data.error && <Button onPress={this.props.data.refetch} title="retry"/>}
}
```
By ready though this commit, I think is may not be what Im looking for.

Issue #878, opened 2017-07-21
Can't import test-utils after update to 1.4.5


anthonator:
It seems that after updating to 1.4.5 `react-apollo/lib/test-utils` has disappeared.

```
module.js:487
    throw err;
    ^

Error: Cannot find module 'react-apollo/lib/test-utils'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/usr/src/app/src/components/accounts/List/ListContainer.test.jsx:8:1)
    at Module._compile (module.js:569:30)
    at loader (/usr/src/app/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .jsx] (/usr/src/app/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at /usr/src/app/node_modules/mocha/lib/mocha.js:230:27
    at Array.forEach (native)
    at Mocha.loadFiles (/usr/src/app/node_modules/mocha/lib/mocha.js:227:14)
    at Mocha.run (/usr/src/app/node_modules/mocha/lib/mocha.js:495:10)
    at Object.<anonymous> (/usr/src/app/node_modules/mocha/bin/_mocha:469:18)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
```

anthonator:
This isn't an issue on 1.4.4.

anthonator:
This is happening for whatever the latest release for `react-apollo` is as of 7/4/2017. NPM and GitHub releases are giving different version numbers so it's a little confusing.

jbaxleyiii:
@anthonator `test-utils` can be found at `react-Apollo/test-utils` instead of at `/lib/test-utils`. I'll work on updating the releases today! Sorry about that!

anthonator:
This fixed the issue.

[...]

PR #1033, opened 2017-08-25
Fix import of test-utils in example

mklopets:
See https://github.com/apollographql/react-apollo/issues/878

Makes example code match actual import usage of test-utils.

### DIFF
--- a/examples/base/src/__tests__/App.js
+++ b/examples/base/src/__tests__/App.js
@@ -1,9 +1,7 @@
 import * as React from 'react';
 import * as renderer from 'react-test-renderer';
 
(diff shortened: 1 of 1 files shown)


mklopets:
Hmm, the tests fail. I might be wrong in my assumption that this change should be made, but it's a change I had to make in my own project using the testing tools. Can someone point out what I'm missing? Namely, why the test-utils should be pulled from `lib` during development? 🤔

jbaxleyiii:
@mklopets during development in the repo, the compiled files live at `/lib` instead of the root like they do on deployment. So the example file has to import it from the `/lib` directory (which is why the test is failing)

mklopets:
Alright, thanks @jbaxleyiii

PR #1068, opened 2017-09-06
Move graphql from devDependencies to dependencies to fix flow errors.

vejersele:
Since graphql flow types are imported in index.js.flow, flow expects the graphql package to be listed as a dependency instead of a devDependency. 

I've noticed the diff is quite large, I suspect a beautifier has changed the file as part of the pre-commit hook.

### Checklist:


jbaxleyiii:
@vejersele is there another way around this? Since flow is only used in development I don't understand why this would be a problem?

PR #1126, opened 2017-09-22
Upgraded to 0.54.1 Flow

jbaxleyiii:
Upgrades flow to `0.54.1` but has to wait until apollo-client is updated to support it


jbaxleyiii:
closed via https://github.com/apollographql/react-apollo/pull/1241

Issue #1083, opened 2017-09-10
Typescript: ChildProps.data should not be optional


choffmeister:
**Intended outcome:**

The typescript definition of `ChildProps.data` should be non-optional, since according to the documentation (http://dev.apollodata.com/react/api-queries.html#graphql-query-data) the `data` property has a `loading`/`error` subproperty to check for completion/errors.

**Actual outcome:**

The actual type definition is

```js
export declare type ChildProps<P, R> = P & {
    data?: QueryProps & R;
    mutate?: MutationFunc<R>;
};
```

forcing users to always check if `data` is defined before being able to access `data.loading` or other properties.

One example from the documentation

```js
render() {
  const { data: { loading, error, todos } } = this.props;
  if (loading) {
    return <p>Loading...</p>;
  } else if (error) {
    return <p>Error!</p>;
  } else {
    return (
      <ul>
        {todos.map(({ id, text }) => (
          <li key={id}>{text}</li>
        ))}
      </ul>
    );
  }
}
```

would fail if using typescript because of `data?`.

**Version**
- apollo-client@1.4.0
- react-apollo@1.4.15

leoasis:
I'd like to add to this one that instead of data being optional, it is the results that should be optional, in the `ChildProps` case, `P` should have all nullable fields (for the case where they are loading).

choffmeister:
What you call `Nullable<T>` should be typescripts `Partial<T>` I guess.

leoasis:
Yeah, duh, good call! haha

choffmeister:
Did look up here https://www.typescriptlang.org/docs/handbook/advanced-types.html. Actually both `Nullable` and `Partial` exist (did only know about `Partial`). Still if I recall correctly the unloaded props are `undefined`, so `Partial` should be the correct one. If there are now objections I would create a PR tomorrow for that.

Jontem:
Looks as this was merged and included in version 1.4.16 but `data` is still optional but now with `Partial<R> ?

```typescript
export type ChildProps<P, R> = P & {
  data?: QueryProps & Partial<R>;
  mutate?: MutationFunc<R>;
};
```
https://github.com/apollographql/react-apollo/blame/3e245205097d427daaac7cb89c0f97eea78032ae/src/types.ts#L70

leoasis:
[...]

PR #1143, opened 2017-09-26
Make ChildProps.data non-optional

choffmeister:
Please see issue #1083

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 - fix shallowEqual bug.
 - Added notifyOnNetworkStatusChange to QueryOpts and MutationOpts Typesccript definitions [#1034](https://github.com/apollographql/react-apollo/pull/1034)
 - Added variables types with Typescript [#997](https://github.com/apollographql/react-apollo/pull/997)
+- Made `ChildProps.data` non-optional [#1143](https://github.com/apollographql/react-apollo/pull/1143)
 
 ### 1.4.15
 - Fix: handle calling refetch in child componentDidMount
--- a/src/types.ts
+++ b/src/types.ts
@@ -67,7 +67,7 @@ export interface OptionProps<TProps, TResult> {
 }
 
 export type ChildProps<P, R> = P & {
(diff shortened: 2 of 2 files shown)


leoasis:
I think you also need to update `OptionProps` defined above https://github.com/apollographql/react-apollo/pull/1143/files#diff-41df78d86ed4de80aae9b0b9560feb2bL65

choffmeister:
True thx. Still there are tests failing which is weired, since I did not change any actual code. Also the tests suite does also not run through locally on `master` branch...

leoasis:
The build fail seems to be related to some typescript checks that now fail because of the changed types: https://travis-ci.org/apollographql/react-apollo/builds/279954551#L773

choffmeister:
OK, these error is from making `data` optional also on `OptionProps`. Before there was no typing error but just some unit tests failed. Will have to look into it.

choffmeister:
@leoasis I reverted the part for `OptionProps` so this is at least correct for now. I am also not really sure that `OptionProps.data` is indeed optional. At least how it is implemented now it is not.

leoasis:
Hmm I think `data` can be optional in both cases, since you won't get it either in `ChildProps` when you set up a mutation.

leoasis:
I think this one is a bit hard to typecheck, since the decision on wether this HoC acts as a mutation or as a query, really depends on the graphql Document that is passed to it, and it's determined at runtime.

jbaxleyiii:
I had no idea this even existed! So cool!

jbaxleyiii:
hmm these test failures seem to be everywhere? I wonder what it is. Anyhow, its not related to this PR so I'll merge this through!

leoasis:
@jbaxleyiii but won't data be undefined if we just use `graphql` for a mutation? I guess that why it was optional in the first place.

Also, I think `OptionProps` here https://github.com/apollographql/react-apollo/pull/1143/files#diff-41df78d86ed4de80aae9b0b9560feb2bL65 should have the same `Partial` treatment to `TResult`

PR #1146, opened 2017-09-27
Upgrades

jbaxleyiii:
This PR upgrades everything but flow 54 since that is still erroring 🎉


pleunv:
Shouldn't `enzyme-adapter-react-16` be in the devDependencies? 😃

PR #1157, opened 2017-09-27
Fix getDataFromTree for functions returning arrays

bkniffler:
Applies to react@16

```
export default ({}) => [<Component1 key={1} />,<Component2 key={2} />/]
```


jbaxleyiii:
@bkniffler can you add a test for this?

PR #1159, opened 2017-09-27
Move the enzyme-adapter to the devDependencies.

pleunv:


kreba:
Thank you. The dependency on enzyme had caused unwarranted peer dependency warnings when I updated to 1.4.16 today. Downgraded to 1.4.15 for now. Glad to see this fixed in 1.4.17 👍

Issue #556, opened 2017-03-22
FetchPolicy 'network-only' returns cached value


davidye:
## Steps to Reproduce

Not sure if this is an issue with react-apollo or apollo-client.

1) Create graphql() component with fetchPolicy: 'network-only'
2) Render the component once - data is fetched from server
3) Print out this.props.data
4) Update the underlying query data
5) Render the component again - this.props.data is stale

### Buggy Behavior

When using fetchPolicy: 'network-only', Component receives the previously cached value.

### Expected Behavior

When using fetchPolicy: 'network-only' should only return fresh results from server.

### Version
- apollo-client@1.0.0-rc.6
- react-apollo@1.0.0-rc.3

helfer:
Thanks @davidye, this sounds like a bug. Could you provide a reproduction with https://github.com/apollographql/react-apollo-error-template? That would help us figure out what's going on.

irfanlone:
I am also seeing the same issue, fetchPolicy: 'network-only', Component receives the previously cached value. versions `apollo-client@1.0.0-rc.3` `react-apollo@1.0.0-rc.3`. I will try to reproduce it.

helfer:
Thanks @irfanlone that would be great! Let me know if you need any help with it 🙂

irfanlone:
@helfer, Here is the PR reproducing the issue, https://github.com/apollographql/react-apollo-error-template/pull/2. Let me know if it makes sense. Thanks for looking into it. 🙂

helfer:
Thanks @irfanlone. Based on your reproduction it looks like this runs two separate queries. What I'm not sure yet is how the code path is different from the one taken in this test here: https://github.com/apollographql/apollo-client/blob/master/test/client.ts#L1282

helfer:
@irfanlone this is embarrassing, but I figured out what's wrong. The error is actually in our docs, and I realized it a few days ago for `pollingInterval`, but I didn't fix it yet. In react-apollo you have to pass options to Apollo Client wrapped in an options object:

```
export default graphql(
  gql`{
    food {
      id
      name
    }
  }`,
  {
    options: { fetchPolicy: 'network-only' },
  }
)(Food)
```
I've fixed it in the docs with this commit: https://github.com/apollographql/react-docs/commit/dd23b58303b6c8415d307b1c48e5ffbec2d45cfc

@davidye I think you most likely had the same problem, so I'll close the issue here. If this doesn't fix your issue, please re-open it.

[...]

PR #1177, opened 2017-10-02
Fixes issue with network-only and cache-and-network fetchPolicy being changed

mhuggins:
I mostly want to open this PR for the sake of conversation.  I can't for the life of me figure out why this logic exists inside of react-apollo, and it seems to be causing issues for many users as pointed out in #556.

I'm interested in some feedback, as I have not been using apollo for long, and it's not clear to me why the desired `fetchPolicy` would be changed despite what the developer intends.

### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
     "colors": "1.1.2",
     "danger": "1.2.0",
     "enzyme": "3.0.0",
+    "enzyme-adapter-react-16": "^1.0.0",
     "enzyme-to-json": "1.6.0",
     "flow-bin": "0.52.0",
     "graphql": "0.11.4",
@@ -149,7 +150,6 @@
   },
   "dependencies": {
     "apollo-client": "^1.4.0",
-    "enzyme-adapter-react-16": "^1.0.0",
     "graphql-tag": "^2.0.0",
     "hoist-non-react-statics": "^2.2.0",
     "invariant": "^2.2.1",
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "enzyme": "3.0.0",
     "enzyme-adapter-react-16": "1.0.0",
     "enzyme-to-json": "1.6.0",
-    "flow-bin": "0.52.0",
+    "flow-bin": "0.56.0",
     "graphql": "0.11.4",
     "immutable": "3.8.1",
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@mhuggins the desired behavior for SSR is to fetch the first time, then to read from the cache during the second loop when creating the components. Since react-apollo uses a two round render on the server, it is important to not send queries (force-fetch) the second time around

mhuggins:
@jbaxleyiii By using cache on the server, every subsequent request has stale data since the data is updating constantly.  How do you make the cache bust between requests?

milanvdmria:
@jbaxleyiii It mainly doesn't make sense when you are also using Apollo to validate some user input. The user enters a certain product, the front end logic changes based on the fact if that product already exists yes or no.
If the caching strategy is forced, the check will always be performed on the very first product the user wants to add and all the consequent products are never correctly checked.

As mentioned in the PR, this forced caching strategy is causing unexpected behavior and issues for quite some people without adding extra value. What could possibly be the advantage of taking away some configuration options from the user?

Mati365:
@mhuggins @milanvdmria ping

mhuggins:
Not sure why I'm being pinged.

Mati365:
Any changes?

mhuggins:
My recommended change is in this PR.

KaySackey:
@jbaxleyiii Typically the client/server implementations will diverge quite a bit during the setup-phase of the app by necessity.

Even, React itself doesn't have an identical API for rendering on the server, versus attaching to the DOM.

Accordingly, I think the best practice would be creating a server-side apollo client without caching, and a client-side apollo client *with* caching. Having the option being forced without warning was a bit surprising, especially since the vanilla client doesn't do this.

jbaxleyiii:
@mhuggins @KaySackey @milanvdmria are you creating a new apollo-client per request or sharing one across server side renders?

I really don't recommend sharing them across requests

milanvdmria:
@jbaxleyiii I dont use any server side rendering. And the apollo client is shared in my front end code.

jbaxleyiii:
@milanvdmria how are you running this code on the client then? Or rather in what use case?  The `fetchData` is designed primarily for SSR?

Issue #1232, opened 2017-10-12
How to use the "client" option on the server.


j:
See this comment: https://github.com/apollographql/react-apollo/issues/1232#issuecomment-339783150

I noticed when using multiple clients, it's a pain on the server side to pass around client instances per request.  This is managed with a single client b/c it's in the context.

A comment suggests to use it like so:

```js
import MyOtherClient from "../other-client"

export default graphql(MyQuery, {
  client: MyOtherClient,
})(MyComponent);
```

However, on the server, this is a lot harder.  You'd have to have a container hoc that stores the extra clients and passes them down as a prop to the graphql wrapped component which seems kind of ugly.

What's a clean way to do this?

j:
I'm trying to create a pullable example of this issue, but just realized that my app creates a client and passes the initialState to it on redux execution.  The client option would be simple without this.  I think the "client" option should have instead been "networkInterface" to switch interfaces and re-use the same client?  Or how can I go about re-using a created client.  The only way I can think of is to use something like context to store the created clients, but that will get ugly fast.

j:
Also, unrelated maybe, but why does ApolloProvider require a redux store and the client have an initialState option?

j:
Just saw that 2.0 is coming soon.  Redux is no longer the store, which makes me now question if the clients and defaultClient props fit now since collisions shouldn't happen if stores are managed in each client independently.

j:
Just got back to this issue.  What I'm finding a pain is that the graphql hoc re-uses client form the ApolloProvider context.  So on server render, it's easy to pass the client + cache down to my wrapped component.  The way to get around it is to create my own wrapped component and pass my second client in the context and use recompose or so to add it to the props then use that prop in the client options configuration.. This just seems way harder and more painful to do than just having it be like:

```js
// this uses defaultClient
const Store = graphql(..., {
  // ...
};

const Repositories = graphql(..., {
  options: () => ({
    client: 'github'
  })({ ... }) => {
  // ...
};

<ApolloProvider clients={{ shopify, github }} defaultClient={shopify}>
  <App>
    <Store />
    <Repositories />
  </App>
</ApolloProvider>
```

In comments, I see this example:
```js
import MyOtherClient from "../other-client"

export default graphql(MyQuery, {
  client: MyOtherClient,
})(MyComponent);
```

which is great and clean... until you want to do this on the server side.  Thoughts?

/cc @jbaxleyiii @flexzuu

j:
[...]

PR #1292, opened 2017-10-26
added clients & defaultClient option.

j:
So, I the client option was merged where users can pass in a Client to override the client in ApolloProvider.  One pain that I've come across using this method is that for SSR, it's complicates things when you need to create a new client per HTTP request.  Without multiple clients, in typical react-apollo setups, this works fine because each request wraps the app in ApolloClient, runs getDataFromTree, etc, etc.  When using SSR, this example in the comments when the client option was originally merged makes it difficult (because the cache is somewhat lost, unless you pass around the original created one)

Current Implementation:

```js
// This is fine on the browser since you can locally cache "other-client" 
// and re-use... but a pain for SSR.
import MyOtherClient from "../other-client"

export default graphql(MyQuery, {
  client: MyOtherClient,
})(MyComponent);
```

This PR adds this functionality:

```js
// Store.js
export default graphql(gql(`...`))(Products); // uses default client

// Repositories.js
export default graphql(gql(`...`), {
  options: {
    client: 'github'
  }
})(Repos);

// App.js
<ApolloProvider clients={{ shopify, github }} defaultClient={shopify}>
  <App>
    <Store />
    <Repositories />
  </App>
</ApolloProvider>
```

I know I'm supposed to create PRs only if there's a consensus, but it was more of a pain setting up an environment to show my use-case than it was to just see if moving all clients to the provider context would fix it.  After I added all of this, it indeed fixed my issue where double fetching on the server & browser would happen.

This still needs a few more tests and error handling when clients with invalid names don't exist, etc.

### Checklist:

- [X] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)

https://github.com/apollographql/react-apollo/issues/1232

- [X] Make sure all of the significant new logic is covered by tests

I still need to add a few more tests and clean things up.

### DIFF
--- a/src/ApolloProvider.tsx
+++ b/src/ApolloProvider.tsx
@@ -8,7 +8,9 @@ import QueryRecyclerProvider from './QueryRecyclerProvider';
 const invariant = require('invariant');
 
 export interface ProviderProps<Cache> {
-  client: ApolloClient<Cache>;
+  client?: ApolloClient<Cache>;
+  clients?: Map<string, ApolloClient<Cache>>;
+  defaultClient?: ApolloClient<Cache>;
 }
 
 export default class ApolloProvider extends Component<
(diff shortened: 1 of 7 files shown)


jbaxleyiii:
@j I'm so so sorry this PR was never responded to! I'm curious, why do you create multiple clients?

j:
@jbaxleyiii hey, I was using Shopify API + Graphcool and the route that was implemented (client option).  In my case, I'd probably just end up using schema stitching, but I've become too uncomfortable with the current state of stitching, so ideally the front-end could just simply call two APIs. :)

jbaxleyiii:
@j would being able to send the request to two different endpoints (not two full clients) work for you? That is possible with the 2.0 and keeps this integration simpler

sh0umik:
Is anyone using this PR till date ? This commit seems fine but also 3 years old. Any feedbacks ?

cansin:
@jbaxleyiii I couldn't quite understand your comment re: "the recommended approach is a single client which can use multiple endpoints". Can you elaborate on how one can re-use a single client instance in order to connect both, let's say, GitHub and Shopify GraphQL APIs?

cansin:
Btw, I am guessing, we should be using `useMutation({ client: anyExtraClient })` instead of trying to provide multiple client instances. But I was intrigued by that comment and wanted to make sure I am not missing a feature that I could rely on. Thanks

PR #1300, opened 2017-10-28
ensure state is null (not undefined) for components during walkTree

mattfysh:
### Checklist:

- [X] (N/A) If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [X] Make sure all of the significant new logic is covered by tests
- [X] (N/A) If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

PR #1310, opened 2017-10-31
initialize state to null if not when not set

### LINKED ISSUE #1249: During SSR, default state of a component should be null (not undefined)
**Intended outcome:**

I'm trying to use react-apollo, with SSR via `getDataFromTree`, together with the `react-emotion` library. 

**Actual outcome:**

An error is throwing during SSR at `getDataFromTree` because `walkTree` fakes a component render, but state is `undefined` instead of `null`

In a component that is part of the tree during SSR, do a console.log to see the value of `this.state` in the `render()` function, it is `undefined` when it should be `null`

**Version**
- apollo-client@1.9.2
- react-apollo@1.4.16

mattfysh:
Fixes #1249

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,10 +2,11 @@
 
 ### vNext
 - BREAKING: removed cleanupApolloState as it is no longer needed!
-- Exported getDataFromTree on the client 
+- Exported getDataFromTree on the client
 - Removed `redux` from peer dependencies. [Issue #1223](https://github.com/apollographql/react-apollo/issues/1223) [PR #1224](https://github.com/apollographql/react-apollo/pull/1224)
 - Support arrays being returned from render in SSR [#1158](https://github.com/apollographql/react-apollo/pull/1158)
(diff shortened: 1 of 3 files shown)


mattfysh:
This fixes an issue with `react-emotion` due to an explicit check in their library for null:

https://github.com/emotion-js/emotion/blob/master/packages/react-emotion/src/index.js#L82

The `react-apollo` SSR walk needs to be updated to match React behaviour, i.e. `isNull(this.state)`

mattfysh:
@jbaxleyiii - this is a rework of #1300 applied to latest master, however tests in an unrelated area are failing. Does this mean they are also failing in master? I can't see any changes I've made that could affect the unit tests that are breaking. Thanks! 👍

jbaxleyiii:
@mattfysh yep not you! That's a flaky test I need to rewrite haha

PR #1337, opened 2017-11-13
Move README.md closer to the 2.0. setup / usage

petetnt:
This PR updates README.md from 1.0. usage closer to the 2.0. setup & usage from https://www.apollographql.com/docs/react/basics/setup.html.

The texts might need some rework though.


petetnt:
Thanks @peggyrayzis!

PR #1340, opened 2017-11-15
Include link to react-apollo 2.x docs in README.

peteratticusberg:
I see we have [this PR](https://github.com/apollographql/react-apollo/pull/1337) up to update the README, which is great, but I noticed that a) it doesn't address setup for users of react-apollo 1.x  and b) it recreates documentation that exists on apollographql.com (specifically, [these](https://www.apollographql.com/docs/react/basics/setup.html)). 

This "solution" preserves the 1.x docs while also directing users of 2.x to the appropriate place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ For an amazing developer experience you may also install the [Apollo Client Deve
 
 ## Usage
 
+*Note: The following pertains to react-apollo 1.x. Docs on setting up 2.x can be found [here](https://www.apollographql.com/docs/react/basics/setup.html)*
+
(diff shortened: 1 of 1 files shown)


petetnt:
#1337 was merged just now, and IHMO the README should be kept up-to-date with the current versions (as I did in #1337).

That said, linking to the 1.x documentation (at https://s3.amazonaws.com/apollo-docs-1.x/index.html) would be great as those that haven't migrated to 2.0 (or don't plan to) do need them.

peteratticusberg:
Awesome! So long as something is up there.

I put this up in part because I thought that maybe the maintainers had an issue with that PR but now I realize you are a maintainer :) 

I agree linking to the 1.x docs would be a good idea at this point. I can put up a PR this -- do you know if can we depend on the reliability of the link you provided for the next 12 months or so?

petetnt:
Not a maintainer myself, just a contributor like you :)

I think that currently that `aws` link is the only source of truth for the 1.x docs and it's commonly shared in the `Apollo` Slack channel too so I think it would be safe to put up.

PR #1355, opened 2017-11-25
Skip null array items in `getDataFromTree`

dan-lee:
I have the case that occasionally a component may return an array with a `null` value in it. Unfortunately this breaks the tree walker and eventually our server side rendering:
> Cannot read property 'type' of undefined
>     at walkTree (/.../node_modules/react-apollo/react-apollo.umd.js:674:34)

From the React standpoint this should be totally fine: https://codesandbox.io/s/y0pl8pkwnx.

So in this PR I added a null check and tests to cover this case.

### DIFF
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -36,6 +36,9 @@ export function walkTree<Cache>(
 
     return;
   }
+
+  if (element === null) return;
+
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@dan-lee this is great! Thanks so much!

PR #1358, opened 2017-11-27
fix readme for displaying todos

kubido:
since http request is asynchronous.
todos will undefined when render called for the first time.
adding handler for undefined todos


jbaxleyiii:
Thanks @kubido!

PR #1362, opened 2017-11-29
Added component name in Unhandled error

dozoisch:
- it's really hard to know from which container the error comes from, this should help it a bit.

---

Let me know what you think! Thanks.


dozoisch:
If you think it's valuable, I'll update the test.

PR #1259, opened 2017-10-19
feat(graphql): Added TGraphQLVariables type

dyst5422:
Added a type generic to the graphql function named TGraphQLVariables.
It allows for specifying types that are required for the query that
should NOT be passed to the underlying component but are just used
for the GraphQL operation. The type is joined to the component's TProps
but not required on the passed in component.

Per apollographql/apollo-client#2198

This should not effect any current implementations as it defaults to `{}`.

I need a sanity check on type inference in this case. I know that typescript does not currently do partial type inference, but I'm pretty sure that it will use default types when type inference is not possible. Someone might want to confirm that as it would be pretty unfortunate if someone were depending on type inference and this change somehow let their type unspecifyable (I don't know how that would occur, but would rather not cause someone an afternoon of type bugs).

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests
- [x] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

There were some failing checks when I pulled the master branch. I took note of both the compile and test results before and after my changes and am including them here in case Travis complains.

Compile state prior to change:
```
jdstewar@LMC-056682 react-apollo (master) $ yarn compile
yarn run v1.1.0
$ tsc
$ npm run bundle

> react-apollo@1.4.16 bundle /Users/jdstewar/gitRepo/react-apollo
> rollup -c && rollup -c rollup.browser.config.js && rollup -c rollup.test-utils.config.js

The following options have been renamed — please update your config: options.entry -> options.input, options.moduleName -> options.name, options.sourceMap -> options.sourcemap, options.dest -> options.output.file, options.format -> options.output.format

lib/index.js → lib/react-apollo.umd.js...
No name was provided for external module 'react' in options.globals – guessing 'React'
No name was provided for external module 'prop-types' in options.globals – guessing 'PropTypes'
No name was provided for external module 'redux' in options.globals – guessing 'redux'
No name was provided for external module 'apollo-client' in options.globals – guessing 'apolloClient'
No name was provided for external module 'graphql-tag' in options.globals – guessing 'graphqlTag'
No name was provided for external module 'react-dom/server' in options.globals – guessing 'ReactDOM'
created lib/react-apollo.umd.js in 119ms
[...]

PR #1365, opened 2017-11-29
feat(graphql): Added Variable generic type

dyst5422:
Originally opened as PR #1259, rebased by hand when Apollo-client 2.0 dropped.

Added a type generic to the graphql function named TGraphQLVariables.
It allows for specifying types that are required for the query that
should NOT be passed to the underlying component but are just used
for the GraphQL operation. The type is joined to the component's TProps
but not required on the passed in component.

Per apollographql/apollo-client#2198

This should not effect any current implementations as it defaults to {}.

Checklist:

- [x]  If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x]  Make sure all of the significant new logic is covered by tests
- [x]  If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -67,11 +67,12 @@ let nextVersion = 0;
 export default function graphql<
   TResult = {},
   TProps = {},
-  TChildProps = ChildProps<TProps, TResult>
+  TChildProps = ChildProps<TProps, TResult>,
+  TGraphQLVariables = {}
 >(
   document: DocumentNode,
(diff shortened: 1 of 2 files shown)


dyst5422:
Not sure how you like to organize the changelog.

jbaxleyiii:
@dyst5422 I can update the changelog! Thanks for these changes!

PR #1302, opened 2017-10-29
Created an example using flow.

excitement-engineer:
I created an example using flow. Hopefully it will be useful for people to have a reference on how to implement flow in react-apollo.

I ran into a couple of issues trying to create this example:

- Using flow 0.57.3 causes there to be 50 type errors. These errors are mostly in graphql-js. Perhaps this library isn’t compatible with this flow version yet.
- I then downgraded to flow 0.53.0 and then the errors disappeared. However, I still can’t get it to work. It doesn’t seem to use the `OperationComponent` type definition from react-apollo. Therefore, the wrapped component’s props aren’t typed properly.

Anyone have any ideas on a solution to the issues I mentioned?

I was also wondering whether the type definitions in react-apollo  can also be used for mutations. I haven’t found any docs for this. I would be glad to help document these definitions and include them in this example repo.

PR #1372, opened 2017-12-02
Add Flow example

excitement-engineer:
This PR is a second attempt (following #1302) at creating an example repo that illustrates the usage of flow with apollo. 

Unfortunately, the flow typing don’t work yet. I’m importing the types`OperationComponent` and `QueryProps` but flow does not seem to register them causing the props `loading` and `error` not to be typed in the component. 

Note, that if I import the types as follows:

```
import type {
  OperationComponent,
  QueryProps,
} from "react-apollo/src/index.js.flow";
```

then loading and error props are typed. However, importing the types like this causes all kinds of flow errors for react-apollo’s internal graphql dependency. 

Anyone have any idea how to import the types directly from `react-apollo`

I decided to use apollo-client 2.0 in this repo. This however causes the `MockedProvider` in the tests to break with the error: 

```
In order to initialize Apollo Client, you must specify link & cache properties on the config object.
            This is part of the required upgrade when migrating from Apollo Client 1.0 to Apollo Client 2.0.
```
Is this a bug in react-apollo or am I setting the test up wrong? 

Does anyone know how to fix the flow/Mocked Provider issues. I have been trying for quite a while but I can’t seem to get it to work.


excitement-engineer:
cc @jbaxleyiii

jbaxleyiii:
@excitement-engineer this is fantastic! I'll figure out whats messing it up today and get that fixed! I'll probably merge this with it failing so I can fix all of the flow related issues on master

PR #1374, opened 2017-12-03
Include flow check in the CI process.

excitement-engineer:
In ran into issues trying to integrate flow into an example project in #1372. I decided to investigate further and found out that the flow tests included in this repo contain all kinds of flow errors.

To highlight these errors I added the flow and typescript type check to the CI script. I think that it is valuable to include these checks in CI to prevent issues from occurring.

I fixed the errors by adding a React import statement to the `test/flow.js` file and updating the GraphQL dev dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "enzyme-adapter-react-16": "^1.0.0",
     "enzyme-to-json": "1.6.0",
     "flow-bin": "0.59.0",
-    "graphql": "^0.11.7",
+    "graphql": "0.11.7",
     "graphql-tag": "^2.0.0",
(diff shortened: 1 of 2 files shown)


excitement-engineer:
It seems that there is an issue in travis when installing the dependencies. Is this a cache issue?

jbaxleyiii:
@excitement-engineer they used to be run on the CI until flow 0.43? came out and broke all of the typings! I think with the latest work on fixing them, we can turn them back on though!

I'm going to close this PR and fix the CI today before I cut the release. Thank you so much for bringing it up though!

PR #1377, opened 2017-12-03
Added test for the MockedProvider component.

excitement-engineer:
I noticed that there were not tests yet for the MockedProvider so I took the liberty of writing some:) I would love to hear feedback!

There are also no docs available on testing in the [react-apollo-documentation](https://www.apollographql.com/docs/react/). Is there anyone working on this at the moment, if not then I would be up for adding this to the docs:)


jbaxleyiii:
@excitement-engineer these look great! There are no docs but I would SO LOVE for their to be some. I'll send you some swag if you write them ;)

PR #1381, opened 2017-12-04
Observed query can return wrong data when variables change

staylor:
`this.queryObservable` can contain wrong `data` values if `updateQuery()` is called without unsubscribing from the original query when variables change.

`data.variables` can contain the wrong value here: https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L526

This specifically happens at the end of `componentWillReceiveProps()` when `updateQuery()` is called. The side effect here is that a unit test now fails: `networkStatus` is expected to be `2` when variables change, but the value is `1` if a new query subscription is created.


staylor:
is this resolved by a commit?

staylor:
this code seems to have changed in a nice direction - will test later tonight and perhaps close this

jbaxleyiii:
@staylor do you have a reproduction I can test with? Curious if the 3.0 fixes this

staylor:
I will fire up my env right now and try to reproduce with this latest beautiful code. If I can still repro, I will get a CodeSandbox together.

staylor:
here is a code sandbox: https://6w9mjxzkn.codesandbox.io/admin/video

staylor:
ok - found the issue

for a connection, you HAVE to use `@connection`, and it HAS to contain the pagination args in `filter` like so `@connection(key: "videos", filter: ["year", "search", "first", "after"])`
if you remove the directive, does not work
if you take “first” and “after” out of there, the data never changes
if you remove the directive completely, and page back to one, blows up
if you remove the directive completely, and apply my patch, it works

jbaxleyiii:
This is an issue with apollo-client and its spreading of variables. The current workaround is to use `{ varName: undefined }` when omitting variables that were previously there. Will get a fix in AC shortly

PR #1399, opened 2017-12-10
Implemented the ApolloConsumer component.

excitement-engineer:
The [roadmap](https://github.com/apollographql/react-apollo/blob/master/ROADMAP.md) for v2.1 describes moving to a component based API (`Query`, `Mutation` and `Subscription`) in addition to the HOCs.

I implemented the component associated with the existing HOC `withApollo` called `ApolloConsumer`. This will allow users access to the apollo-client instance using a render prop as follows: 

```js
<ApolloConsumer render={client => {
    return <div>data</div>
}} />
```

I got my inspiration from react-router's API. They have a [Route](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Route.js) component and a [withRouter](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/withRouter.js) HOC. The implementation of the HOC is simply a wrapper around the `Route` component. The same could be done for `withApollo`, it can simply wrap the `ApolloConsumer`.

Curious to hear how people feel about such a component and the API proposed above.

> Note, this is a WIP. Tests still need to be added.


rosskevin:
Style-wise you could use `children` as the render call back, no need to add a `render` prop. Adding a render prop is more verbose but achieves the same thing. react-i18next is a good example of this.

excitement-engineer:
Hey @rosskevin, this is indeed a style issue. I know that react router uses the "render" prop pattern for example. I will leave it the way it is for now and wait for feedback from others:)

excitement-engineer:
I updated the component to use `children` for rendering in favor of a `render` prop.

src/ApolloConsumer.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import ApolloClient from 'apollo-client';
+const invariant = require('invariant');
+
+export interface ApolloConsumerProps {
+  children: (client: ApolloClient<any>) => React.ReactElement<any>;
```
rosskevin: I think the return should `React.ReactNode` - it is looser than `ReactElement` and we have no real concerns about the return type.

excitement-engineer: Agreed, as soon as I update the type to `React.ReactNode` I get the typescript error: 

```
Type '(props: ApolloConsumerProps & { children?: ReactNode; }, context: any) => ReactNode' is not assignable to type 'StatelessComponent<ApolloConsumerProps>'.
  Type 'ReactNode' is not assignable to type 'ReactElement<any>'.
    Type 'string' is not assignable to type 'ReactElement<any>'.
```

It seems that StatelessComponent not accept a `ReactNode` as the return type. Any ideas on how to fix this?

rosskevin: That's mistake then, it should match render's return type. I was thinking it would be node because 16 allows arrays.

rosskevin: Should say my mistake - I'm mobile

PR #1395, opened 2017-12-09
Fully type `ChildProps` and reorder `graphql()` types to reduce verbosity

rosskevin:
### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ ] Make sure all of the significant new logic is covered by tests
- [ ] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

---

## Context
I have a fully typescript codebase, and am using `apollo-codegen` to generate types for mutations and queries.  I found using a fully typed `react-apollo` was not possible with the current type `ChildProps` helper, and very verbose with the `graphql` parameterized type ordering.

This changes/fixes parameterized types only and does not affect any runtime code.

## Problems
- `ChildProps` - is not fully parameterized and does not allow passing through of variables to `MutationFunc`
- `graphql` parameterized type ordering places a derived type `ChildProps` in the third position, making a really verbose fully typed usage
- `graphql` parameterized type did not pass `TVariables` to derived type `ChildProps`
- `graphql` parameterized type ordering now matches `ChildProps` for consistency - `props, operation, operation variables`

## Before example usage

```js
class MyDialog extends React.Component<
  ChildProps<MyDialogProps, UpdateMutation>, // <= Mutation variables untyped
  State
  > {
}
export default graphql<
  UpdateMutation, 
  MyDialogProps, 
  ChildProps<MyDialogProps, UpdateMutation>, // <= Mutation variables untyped and derived type ordering causes more verbose specification of parameterized types 
  UpdateMutationVariables>(
  UPDATE,
)(MyDialog)
```

## After example usage

```js
class MyDialog extends React.Component<
  ChildProps<MyDialogProps, UpdateMutation, UpdateMutationVariables>,
  State
  > {
}

export default graphql<MyDialogProps, UpdateMutation, UpdateMutationVariables>(
  UPDATE,
)(MyDialog)
```

PR #1402, opened 2017-12-11
Dependencies update, typescript check tests, fix hoc inference

### LINKED ISSUE #1404: test sources have typescript errors and are not checked in `yarn type-check`
**Intended outcome:**
`test` sources are asserted as valid typescript

**Actual outcome:**
test sources are not asserted with the exception of `test/typescript.ts`

Some `test/**/*.ts` files have errors (noticed while working on PR #1402), but you may
- add an obvious ts error not in `test/typescript.ts`
- `yarn type-check`

**Version**
- react-apollo@2.0.4 (master)

I'll take a shot at this after #1402

test-ts-errors

rosskevin:
**testing privately (all good so far), but feel free if you want to ping me to confirm before squash+merge**

## Summary
This PR is largely housekeeping and the fallout from dependency updates.  There are no runtime changes included in this PR with the exception of removing the one deprecated `skip` prop.  
- Jest 20->21 required some test assertion changes for which I created a custom matcher.
- Tests are all now valid typescript and pass tslint
- Remove deprecated `operationOptions.options.skip`, use `operationOptions.skip` instead

## Typescript summary
Typescript types have been refactored for consistency and a focus on typescript user experience.  This means consistent parameterized type names, types have been refactored and exported individually, and changes made to the signature (exported type names were not changed).
- Fix graphql hoc inference type signature
- Make all test sources type check (closes #1404)
- Greater reuse of types `types.ts`
- Fix `shouldResubscribe` props on `OperationOption`
- `graphql` parameterized types streamlined for a) full typing; and b) ease of use; and c) consistency.  New parameterized usage is: `graphql<TProps, TData, TGraphQLVariables, TChildProps>` where none are required and full typing only requires the first three params (`TChildProps` can be derived).  The first three params will be most typical when used in conjunction with `apollo-codegen`. **Note** the ordering of graphql parameters is different to be consistent with `ChildProps` and variants and to minimize verbosity (supersedes #1395)
- Add `ChildDataProps` and `ChildMutateProps` for optional stronger typed usage version of `ChildProps`, and can be used as the fourth parameter to `graphql`.  This is in preparation for a potential split query vs mutation usage of the `graphql` component.

## Misc
- Update all dependencies 
- Update script usage - remove any deprecations
- Update jest setup
- Update prettier setup
- Update typescript setup
- Run `lint-fix` (prettier + tslint --fix)
- Remove integration tests/samples for external libraries (out of date and not necessary for unit testing)


rosskevin:
The big PR warning is because I found the scripts were not running prettier.  I have configured and run that so it is probably best to review this PR commit-by-commit, skipping the `run lint-fix (prettier + tslint fix) 824c5b2` commit.

rosskevin:
My last push is broken, it is failing with errors such as:

```
Expected value to equal:
  {"people": [{"name": "Luke Skywalker"}]}
Received:
  {"people": [{"name": "Luke Skywalker", Symbol(id): "$ROOT_QUERY.allPeople({\"first\":1}).people.0"}], Symbol(id): "$ROOT_QUERY.allPeople({\"first\":1})"}
```

I found this somewhat related [SO post](https://stackoverflow.com/questions/44734124/access-the-value-of-symbolid-property-on-an-object), except ironically we aren't trying to get the `Symbol(id)`, it is just showing up.  I moved to typescript recently from flow, so maybe this is just something common I haven't bumped into.

@jbaxleyiii any thoughts/guidance here?  I'd like to get this PR to the finish line.

rosskevin:
The latest jest `isEqual` is aware of symbols, so several tests are failing.

It seems `expect(props.data.allPeople).toEqual(data1.allPeople)` will need to be the equivalent of 

```js
expect(JSON.parse(JSON.stringify(props.data.allPeople))).toEqual(
  data1.allPeople,
);
```

I'm making a utility method to serialize/deserialize to wash out the symbols for comparison to the hardcoded fixtures.

rosskevin:
I'm going to publish this privately and start integrating with our codebase to ensure no errors in typing that I didn't already uncover with the tests.

orta:
@rosskevin - I'd recommend updating Danger to 2.x, it stops relying on Jest internals then 👍

rosskevin:
@orta I updated to 2.1.4 - so perhaps that is the problem - any guide for 2.x and TS?  I saw a bunch of TS related docs removed from the repo.

orta:
Should "just work" will clone and look now.

orta:
Interesting, so the results of running [the ts transpiler](https://github.com/danger/danger-js/blob/master/source/runner/runners/utils/transpiler.ts#L34-L39) with [your tsconfig.json](https://github.com/apollographql/react-apollo/blob/master/tsconfig.json) seems to be that it produces `import`s instead of `require`s.

e.g. with the tsconfig, and with an empty config.

screen shot 2017-12-14 at 2 17 21 pm

Digging into see if there's something I'm doing wrong in Danger

---
 
I think in this case Danger needs to override/remove the [module](https://www.typescriptlang.org/docs/handbook/compiler-options.html) option in a user's tsconfig. Can't see a use-case where it should be needed.

orta:
OK, want to try updating to 2.1.5 please? 

When I update & run `node --inspect  node_modules/danger/distribution/commands/danger-pr.js https://github.com/apollographql/apollo-client/pull/2476` with that in your PR - it passes

rosskevin:
Thanks @orta!  danger ran without a problem.

orta:
Thanks for poking me - I had a PR on our RN repo which exhibited the same issue https://github.com/artsy/emission/pull/804

rosskevin:
@jbaxleyiii all green, ready for review.  In the meantime I'm working with the package in our codebase to find any inconsistencies with the types.

rosskevin:
Amazing, committed a script change some other test fails. We need a better way than timing, because clearly travis is inconsistent.

rosskevin:
I'm never committing again for fear of breaking the time-dependent tests on travis :flushed:

jbaxleyiii:
@rosskevin this is absolutely INCREDIBLE WORK. Thank you SO SO much for showing the imitative and commitment here!

I'm good to merge this!

PR #1413, opened 2017-12-18
Fix polling.test

### LINKED ISSUE #1316: TS2688:Cannot find type definition file for 'react'
I am using typescript version 2.3.4 in my app . I am getting  the above error  in server.d.ts file in react-apollo node modules.kindly   patch this  in your package.As I am unable to use it.I am following versions for my App
 typescript version 2.3.4
"react-apollo": "^1.4.15",

rosskevin:
Was going to change types to `devDependencies` but reverted, leaving the test change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "apollo-client": "^2.0.3",
     "react": "0.14.x || 15.* || ^15.0.0 || ^16.0.0"
   },
-  "devDependencies": {
+  "dependencies": {
     "@types/enzyme": "3.1.5",
     "@types/graphql": "0.11.7",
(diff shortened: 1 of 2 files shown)


rosskevin:
Flaky polling.test again.

leoasis:
Won't this make all users of the library have to download the types even if they're not using Typescript? I don't think this is a good idea, even being a Typescript user myself. Also, some of those types added in dependencies are related to the test libraries being used for internally for development.

rosskevin:
Good points, definitely test types need to go back. I'm going to roll back the types change for now and just keep the test fix.

PR #1438, opened 2017-12-20
add notifyOnLoadingStatusChange to fix issue #1347

### LINKED ISSUE #1347: Refetched query not updating "loading" prop
Hey there,

I've got a component that is wrapped with an apollo query. I use react-apollo's "refetchQueries" to cause that query to update itself (after a mutation is run by a separate component). The new data is getting fetched, but unfortunately, the "loading" prop for that mutation does not appear to be getting changed. Thus I am not able to get a spinner to show over the component while it's updated data is getting loaded.

Is this how refetch queries is set up to work?

Is there a way to get the components interested in the query to be made aware that a fresh query is being made?

Hope that all makes sense.
Thanks!

tonyjiang88215:
Closes #1347 

Redefine function refetch() on the graphql hoc, and we do forceUpdate in this function and call the original refetch function.The graphql re-render can pass the right props through calling dataForChild();


tonyjiang88215:
At least version, i set up a notifyOnLoadingStatusChange in OperationOption, when notifyOnLoadingStatusChange is true, calling refetch function will re-render and pass the new loading status to the  wrappedComponent.
If this is acceptable maybe such as fetchMore can do this either. ^ _ ^

rosskevin:
@tonyjiang88215 please note that I am not neglecting this PR, I simply am not familiar enough to make a call on it.  @jbaxleyiii will be back soon, so I'll let him review.

@excitement-engineer any change we adopt here we should probably see how it impacts `Query`

tonyjiang88215:
@rosskevin thanks for feeding back.

In my option, i think the `react-apollo` is HOC that makes the React UI Component has the ability to use `apollo-client` and get the query(mutation) result in react-way, so i think its more convenient if the loading status or other status can pass to wrapped component reactively.

It`s all right if this PR is not accept , i can still do it in my project. : )

Thanks for the time again.

jbaxleyiii:
@tonyjiang88215 I think this is already supported by using `notifyOnNetworkStatusChange` in the options function? [docs link here](https://www.apollographql.com/docs/react/basics/queries.html#graphql-config-options-notifyOnNetworkStatusChange)

PR #1448, opened 2017-12-23
Included tests by adding .test extension

excitement-engineer:
Included Query.test.js and test-utils.test.js into the test suite by adding the .test extension.


rosskevin:
looks good but tests aren't running again due to what looks like a conflict again with prettier/tslint.  I've never used that package you added so not sure if it actually solves it here.

excitement-engineer:
cc @rosskevin, I finally fixed the tests:)

excitement-engineer:
I agree with you. I was hesitant to update to snapshots as well but I found no other clean way to run the assertions. Note, other test files use a `toEqualJson()` assertion (which is implemented in this lib), however this is broken causing errors as soon as the assertion fails. Something we should correct in the future.

rosskevin:
I had to add `toEqualJson` for the jest upgrade because the comparison of our hardcoded data did not have a `Symbol(id)` in it, and the json comparison is what we desired.

PR #1449, opened 2017-12-23
Made prettier responsible for all formatting

excitement-engineer:
This PR makes prettier responsible for all formatting. This is something that is needed in PR #1448 where the conflicting formatting rules between prettier and tslint is causing issues.


excitement-engineer:
cc @rosskevin, you want to verify this?

PR #1451, opened 2017-12-23
Removed the react-native test.

excitement-engineer:
This PR removes the only react-native test in the repo. I don’t really see the value of including this test in the repo, in addition it adds complexity with the jest configurations and adds some other devDependencies which can now be removed.


excitement-engineer:
Note, this is WIP. I am completing this as soon as I find some more time.

excitement-engineer:
cc @rosskevin @jbaxleyiii

rosskevin:
@excitement-engineer can you rebase - I'd like to get this one merged.

test/test-utils/shim.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * React requires the requestAnimationFrame to be defined for tests.
+ */
+// Prevent typescript from giving error.
+const globalAny: any = global;
+
+globalAny.requestAnimationFrame = callback => {
+  setTimeout(callback, 0);
```
rosskevin: This is the long way to type it, you could just do `(global as any).requestAnimationFrame`

excitement-engineer:
It seems like the ci is not starting, maybe travis is down. The ci will not pass because of the broken test. Given that it seems to be testing something internal in apollo-client. Perhaps we should simply delete the test

rosskevin:
If you resolve conflicts and push a change, it will run.

PR #1452, opened 2017-12-23
Updated lint rule to make prettier responsible for formatting.

excitement-engineer:
Following #1449, attempt 2 at making prettier responsible for formatting.

I used a tool to remove all the lint rules that conflicted with prettier in the tslint.json. I also tested it on the file `Query.tsx` that is causing issues in #1448. 

This allowed me to remove `tslint-config-prettier` as a dependency.

Finally, I also updated the `pretty` script and the lint-staged script in the `package.json`.


excitement-engineer:
cc @rosskevin

PR #1454, opened 2017-12-24
Remove flow-typed from the repo

excitement-engineer:
This PR removes `flow-typed` from the project.

The project was setup so that the `flow-typed` definitions were not included in source control, instead they were generated locally using a npm script. This goes against flow-typed’s [FAQ](https://github.com/flowtype/flow-typed/wiki/FAQs#why-do-i-need-to-commit-the-libdefs-that-flow-typed-installs-for-my-project), they suggest to include it in version control.

I don’t really see any use of flow-typed in the project therefore I have decided to simply remove it completely from the project.


rosskevin:
@excitement-engineer I did a merge, but it looks like you started with the old graphql test for `polling interval using options`.  Check out the updates I did to `polling.test.tsx` `allows a polling query to be created`.  What I found over many attempts is that these polling tests are unreliable on travis - regardless of how lenient you make the timing.

Perhaps you can find something else, but we either need to solve that problem or make your test more lenient.

PR #1457, opened 2017-12-26
Fixed the typings for the tests

excitement-engineer:
This PR fixes the typescript issues in the project. 

I removed the `@types/jasmine`, this allowed me to remove`toMatchSnapshot.d.ts`.

The `toEqualJson` matcher is now also recognised in the matchers. This matchers also did not work, crashing when the assertion failed. I have tried to fix this but it turns out that defining a matchers in terms of another matcher is not supported by Jest, see [issue](https://github.com/facebook/jest/issues/2547). This means that writing `expect(received).not.toEqualJson(expected)` does not work. Instead I think I have found a much simpler solution, a `stripSymbol` function that simply removes all the symbols from an object, this can then be used to run assertions using `toEqual`. I have converted one test to use this method in order to get feedback. One thing I like about `stripSymbols` is that it makes the purpose of the function very clear compared to `toEqualJson` which may leave readers a bit confused as to the reason of this matcher.

I also noticed that we are not running type checks in the CI, this should be turned on in a follow-up PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "@types/graphql": "0.11.7",
     "@types/invariant": "2.2.29",
     "@types/isomorphic-fetch": "0.0.34",
-    "@types/jasmine": "2.8.2",
     "@types/jest": "21.1.9",
     "@types/lodash": "4.14.91",
     "@types/object-assign": "4.0.30",
--- a/test/react-web/client/graphql/queries/api.test.tsx
+++ b/test/react-web/client/graphql/queries/api.test.tsx
@@ -11,6 +11,7 @@ import {
   OptionProps,
 } from '../../../../../src';
 import '../../../../test-utils/toEqualJson';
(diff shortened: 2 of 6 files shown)


excitement-engineer:
cc @rosskevin, what do you think of using `stripSymbols`?

rosskevin:
I'm well aware of https://github.com/facebook/jest/issues/2547

I had quite the trouble getting the custom matcher added.  I chatted with @jbaxleyiii on slack and we figured that would be better than adding another method to pre-process the data, but a custom matched delegating to another matcher hasn't proven to be as easy as it might seem.  If chaining is also broken with the current implementation, then we should just use the utility method that you have prototyped here.  It is explicit and easy to see what is going on.

Thanks for digging into this, I know how much time it can take on a seemingly trivial bit of code.

rosskevin:
I'm going to fold this work into #1462 - I need it there to complete it.  I ran into the same thing.

PR #1460, opened 2017-12-26
Attempt at speeding up the tests

excitement-engineer:
I noticed the tests are running really slowly on CI, so I decided to increase their speed. See [jest docs](http://facebook.github.io/jest/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server) for more info.


excitement-engineer:
This decreases the jest execution time from 148 seconds in #1459 to 35 seconds:)

PR #1461, opened 2017-12-26
Fix for flaky polling tests

jdmunro:
After pulling master I observed that some tests related to polling functionality were failing intermittently. After a discussion with @rosskevin he pointed me in the direction of a timing workaround in `polling.test.tsx` which needed firming up.

Rather than relying on magic numbers for timing intervals it seems to be more sensible to take full control of advancing the timers ourselves. Fortunately [Jest provides an API for doing this](http://facebook.github.io/jest/docs/en/timer-mocks.html) and we can manually advance time as required. This should make the tests deterministic.

Note that `jest.runTimersToTime` has been renamed to `jest.advanceTimersByTime` in v22.0.0.

I think I may have another pass at this to rename some of the test variables as its actually the render count we are asserting on and not the number of polls that happened - although there's plenty of cohesion between the two.

### DIFF
--- a/test/react-web/client/Query.test.js
+++ b/test/react-web/client/Query.test.js
@@ -35,6 +35,10 @@ const catchAsyncError = (done, cb) => {
 };
 
 describe('Query component', () => {
+  beforeEach(() => {
+    jest.useRealTimers();
+  });
+
   it('calls the children prop', done => {
     const Component = () => (
       <Query query={query}>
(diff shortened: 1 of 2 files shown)


test/react-web/client/Query.test.js
```diff
@@ -35,6 +35,10 @@ const catchAsyncError = (done, cb) => {
 };
 
 describe('Query component', () => {
+  beforeEach(() => {
+    jest.useRealTimers();
```
jdmunro: If we don't reset this and one of our test fails this will cause other tests to incorrectly fail due to timeouts. This seems to be the safest way to ensure that other tests are unaffected.

rosskevin:
Fantastic work - thank you.

/cc @excitement-engineer

excitement-engineer:
This is great! Thanks so much for this elegant solution @jdmunro!

PR #1462, opened 2017-12-26
Migrate Query.test to tsx and parameterize Query.tsx

rosskevin:
- `Query.test` was js - migrated to `tsx` and fixed errors by parameterizing `Query.tsx` with `TData`
- Removed custom jest matcher prototyped in #1457 by @excitement-engineer - switched to more explicit/flexible/working utility method


.travis.yml
```diff
@@ -10,7 +10,7 @@ cache:
 before_install:
   - npm install -g npm@^3 --cache-min 999999999
   - npm install -g coveralls --cache-min 999999999
-  - time npm i -g yarn --cache-min 999999999
```
excitement-engineer: Why did you remove this in the PR?

test/test-utils/catchAsyncError.ts
```diff
@@ -0,0 +1,9 @@
+const catchAsyncError = (done, cb) => {
```
excitement-engineer: There is also a `wrap` in the test-utils that does something very similar. Could we merge these into one single method? Perhaps an idea for a follow-up PR?

package.json
```diff
@@ -24,10 +24,10 @@
       "rollup -c && rollup -c rollup.browser.config.js && rollup -c rollup.test-utils.config.js",
     "watch": "tsc -w",
     "lint": "tslint --project tsconfig.json --config tslint.json",
+    "lint:fix":
+      "npm run prettier && tslint 'src/*.ts*' --project tsconfig.json --fix",
     "lint-staged": "lint-staged",
-    "lint-fix":
-      "npm run pretty && tslint 'src/*.ts*' --project tsconfig.json --fix",
-    "pretty":
```
excitement-engineer: Just a small tip. But could you try and keep your PR as condensed as possible so that you are fixing a single thing in a PR. So I would prefer to see unrelated changes like this in a separate PR. I prefer to keep the PRs as small as possible in order to maximize the quality of the review.

.travis.yml
rosskevin: We don't use yarn at all. Why install it alongside npm?

test/test-utils/catchAsyncError.ts
rosskevin: We could merge - I just extracted because it wasn't necessary here.

excitement-engineer: Great. Maybe an idea to do it in a follow-up PR to keep the code clean and prevent duplication of methods from appearing.

.travis.yml
excitement-engineer: Good point! I think that the travis.yml could be cleaned up a bit as well, lot of unused stuff in there.

excitement-engineer: good for a follow-up PR

rosskevin: BTW - I use yarn, I just didn't want to unnecessarily change here.  But I don't mind switching if others want to.

PR #1465, opened 2017-12-26
Remove copied shallowEqual code - use fbjs

rosskevin:
- Remove copied code `shallowEqual` and delegate to `fbjs`
- change tsconfig.json to allow for direct `import = require` of commonjs modules


evocateur:
Adding a dependency on fbjs for _one_ function seems a rather Pyrrhic victory. Now anyone bundling this library gets a giant blob of unused, non-tree-shakeable code. :/

PR #1471, opened 2017-12-27
Typescript example update

rosskevin:
Working on upgrading the typescript example to use the lates apollo-codegen with typescript-modern and the Query render callback component.

I have tried to use `src/` direct from the example and have fixed various type usage issues found in the source, though should be no runtime changes.  This is going to need to be handled bit by bit to get it updated and running.


rosskevin:
I'm actually not sure what is left here - as the relative imports don't work and we need them.  The goal here is to get this dir to work with https://codesandbox.io

rohmanhm:
Hi, thanks for the great example.
Is it already done?

I found this line trying to import `Query` from `react-apollo`.
https://github.com/apollographql/react-apollo/blob/3e3dfc5b34fc242e0d0ed6efe69dd6f05a33c5fd/examples/typescript/src/Character.tsx#L6

But, on typescript definition `Query` properties do not exist. Can you explain?

rosskevin:
The example is not done and those components are only on `master`

Issue #1445, opened 2017-12-21
2.0.4 requires unstated dependency @types/lodash.flowright


jonaskello:
Error description is already in #1286 but it is closed so I open another issue because this error is also in 2.0.4 (which has been confirmed several times in #1286). 

Basically, if you just install 2.0.4 it is broken from the start with this error. The easiest work-around is to add a file `fix-react-apollo.d.ts` (the name does not matter) which contains one line `declare module "lodash.flowright";` as described in [this comment](https://github.com/apollographql/react-apollo/issues/1286#issuecomment-350591666).

rosskevin:
Watchers please check and discuss my comment https://github.com/apollographql/react-apollo/pull/1478#issuecomment-354132340 in the PR.

PR #1474, opened 2017-12-27
Fix issue #1445: Add typings for lodash.flowright

excitement-engineer:
Attempt at fixing #1445. I am  not too familiar with typescript yet so could someone verify for me that this is indeed the solution to the issue.


rosskevin:
This won't fix it because that `.d.ts` file does not make it into the release.  I'll take a shot at it.

PR #1476, opened 2017-12-27
Cleaned up travis.yml and use yarn for dependencies

excitement-engineer:
Cleaned up travis.yml and use yarn for managing dependencies. I also simplified the scripts in the package.json and added type-check to the CI.


excitement-engineer:
@rosskevin, you updated the type name from `QueryProps` to `GraphqlQueryControls` however the flow type has not been updated yet causing an issue in the type-check.

PR #1478, opened 2017-12-27
Fix lodash.* packages - remove re-export `compose`

### LINKED ISSUE #1445: 2.0.4 requires unstated dependency @types/lodash.flowright

rosskevin:
Closes #1445.  Supercedes #1474 

## Updated summary
We were re-exporting `flowright` a.k.a. `compose` which was implicitly or explicitly using the `any` type - which is not a good idea.  Instead, remove this re-export and allow users to choose their composition method on their own.

## Migration
User may migrate by:
1. `yarn add lodash @types/lodash` && using with `const compose = require('lodash/flowRight)`; or
2. `yarn add recompose @types/recompose` && using with `const compose = require('recompose/compose')`

## Other changes
- lock down bundle max size to 4.7 KB prior to change (4.67KB actual)
- remove `lodash.pick` and `lodash.flowright` packages
- use `lodash/pick` which has types in `@types/lodash`
- re-compile and check file size
- fix inconsistency with flow types `QueryProps` -> `GraphqlQueryControls`


rosskevin:
I see in the `lib/index.d.ts` that compose is exported as `any`.  So while **our** usage is fully typed, we are re-exporting something as an explicit `any`.  While some might be good with it, it seems like we are dumbing down types.

rosskevin:
@jbaxleyiii @excitement-engineer I propose we remove the re-export of `flowright` because it is dumbing down the type-checking, and we are assuming it is how users will want to compose.

excitement-engineer:
I agree. I think the docs can point out some the various alternatives and let the user decide which to use.

excitement-engineer:
Can you add the breaking change in the flow type name to the changelog as well? Otherwise it looks good

rosskevin:
Already in the changelog but I'll add the second link to this issue.

rosskevin:
@excitement-engineer tests are failing on travis only - something with imports.  I tweaked the jest config a bit but I'm at a loss.  Wiped out node_modules locally and still all good.  Wiped out cache on travis and it still fails.  Makes no sense.

develomark:
+1 on this approach, however, I simply replaced:

```
import * as compose from 'lodash.flowright';
export { compose };
```

with

```
import { flowRight as compose} from "lodash";
export { compose };
```

in `browser.d.ts` to fix TS compile errors.

ryanflowers:
Hello @rosskevin I see you fixed this issue and merged it to master. Thank you! I'm just wondering when it will be published in npm for consumption..

kaanozcan:
I am also wondering when will this be released?

PR #1479, opened 2017-12-27
Fix TypeScript test coverage

jdmunro:
I was looking at the test coverage to see where any weaknesses lie in terms of missing test cases - I noticed that the reported missed lines weren't mapped to the actual locations.

If we add the [`mapCoverage` Jest option](http://facebook.github.io/jest/docs/en/configuration.html#mapcoverage-boolean) the coverage output now matches up to the correct lines so that we can start to identify any missing test cases.

**Before**

screen shot 2017-12-27 at 17 56 11

**After**

screen shot 2017-12-27 at 17 54 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "danger": "danger run --verbose",
     "deploy": "./scripts/prepare-package.sh && cd npm && npm publish",
     "test": "npm run lint && npm run type-check && npm run testonly",
-    "testonly": "jest --maxWorkers=4",
+    "testonly": "jest --maxWorkers=4 --coverage",
     "test-watch": "jest --watch",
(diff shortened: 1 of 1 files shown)


jdmunro:
@rosskevin We are not currently generating coverage reports on Travis - is this something we want to setup?

> /home/travis/.travis/job_stages: line 57: ./coverage/lcov.info: No such file or directory

rosskevin:
Sure, coverage is a good thing

jdmunro:
@rosskevin Here we go, working now: https://coveralls.io/builds/14817755
Not sure why this wasn't already enabled - hopefully doesn't cause any issues 😄

PR #1482, opened 2017-12-28
Removed snapshots from test

excitement-engineer:
Removed snapshots from the Query test in favour of direct assertions


rosskevin:
Nice! this makes me feel safer in the future that no-one will miss a bad snapshot update

PR #1483, opened 2017-12-28
v2.1 Implementation of Subscription component

excitement-engineer:
Implementation of the Subscription component for the next version of react-apollo.


rosskevin:
Noun or verb?
```html
<Subscription subscription={s}>
```
or
```html
<Subscribe subscription={s}>
```

excitement-engineer:
I'm in favour of sticking to the exact language used by graphql for the component: `Query`, `Mutation` and `Subscription` to prevent any confusion, what do you think?

A `node` does not intuitively signal a `DocumentNode` to me, should we simply call the prop `query`?

```
<Query query={...} />
<Subscription query={...} />
```

rosskevin:
I'm not sure. Using query is a length improvement, and I _think_ an accurate name (not positive though). Either way I want to reduce length without reducing explicitness or clarity.

src/Subscriptions.tsx
```diff
@@ -0,0 +1,134 @@
...
+  error: ApolloError;
+}
+
+export interface SubscriptionProps {
+  query: DocumentNode;
+  variables?: OperationVariables;
+  children: (result: any) => React.ReactNode;
+}
+
+export interface SubscriptionState<TData = any> {
```
rosskevin: Remove `export`.  I just call these `State` since they are essentially private/not exported.

src/Subscriptions.tsx
```diff
@@ -0,0 +1,134 @@
...
+  SubscriptionProps,
+  SubscriptionState<TData>
+> {
+  static contextTypes = {
+    client: PropTypes.object.isRequired,
+  };
+
+  private client: ApolloClient<any>;
+  //TODO: What is the type here?
+  private queryObservable: any;
```
rosskevin: From `ApolloClient.d.ts` (your IDE should allow you to follow the declaration of `client.subscribe`
```js
import {Observable} from 'apollo-client/util/Observable'
Observable<any>
```

src/Subscriptions.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    if (this.client !== nextContext.client) {
+      this.client = nextContext.client;
+    }
+
+    this.endSubscription();
+    this.initialize(nextProps);
+    this.startSubscription();
+    this.setState(this.getInitialState(nextProps));
+  }
```
rosskevin: My preference in this section is that it is unnecessarily broken up.  `initialize` and `getInitialState` are not reused, so i would simply bring their implementation up here.  It increases comprehension without increasing complexity.

I could even say the same about start/end subscription - but I would have to see it before making a judgement.

My go-to in this case would probably be all that code right here.  All init code, not reused, fairly simple, and if needed, a one-line comment to break it up.

Reasoning: when I see code in a method that is one-line I think - _this must be reused, where is it reused and why?_

src/Subscriptions.tsx
excitement-engineer: I tried this but typescript doesn't allow it with error: 

```
'extends' clause of exported class 'Subscription' has or is using private name 'SubscriptionState'.
```

src/Subscriptions.tsx
excitement-engineer: I'm not sure I understand? All of these methods are being reused in the component

src/Subscriptions.tsx
excitement-engineer: Great thanks!

src/Subscriptions.tsx
rosskevin: ah, ok good.  It would have been easier for me to spot in my IDE, but you are right, these are needed.

src/Subscriptions.tsx
rosskevin: ok, if TS is going to complain and require the export (an obscure open issue in TS), then it is best to leave it as-is.

PR #1488, opened 2017-12-28
Add `client` to `Query`'s `QueryResult`

rosskevin:
- Add `client` to `Query`'s `QueryResult` to prevent the need for query + mutation components to need to require either `ApolloConsumer` or `withApollo`
- Add test for `client` in `QueryResult`
- Rename script `testonly` to `jest`
- Regenerate snapshot as they didn't include `client`
- Reorganize `Query.test` based on grouped describe contexts

grouping


rosskevin:
`bundlesize` is currently failing, will probably be back up soon - https://github.com/siddharthkp/bundlesize/issues/190

rosskevin:
Someone needs to hit rebuild on travis later.  This is good to go.

excitement-engineer:
I'm wondering if it is really necessary to add the `client` to the render prop, is it really that awkward to add an `withApollo` or `ApolloConsumer` for those cases where someone wants to do a query + mutation?

The improvements to the test are great! Thanks so much, it is organized much better now:)

rosskevin:
Yes, it is a big improvement for me (my code is already converted). I have several render call backs so just another adds to the nesting, and with 30 components that are mutations + queries, it's just more cruft when I know the client is there. 

The beauty of this pattern is that the user can take or leave it. They can choose to destructure it here or from ApolloConsumer.

PR #1495, opened 2017-12-29
Disregard falsy elements when walking tree

bostrom:
This PR fixes a problem with server side rendering when rendering an array and one or more of the items in the array is falsy (but other than `null`). This would throw an `cannot read property type of null` error. 

An example case when this can happen is when you do a inline conditional render:

```
const renderSmth = false;

render() {
  return [
    renderSmth && (<Smth key="smth" />),
    <MoreContent key="morecontent" />
  ];
}
```

React handles this case gracefully by disregarding falsy values, not rendering them. I would like to assume that Apollo would do the same, but that's not the case, currently it only checks if an element is `null`. 

This PR replaces the `element == null` with `!element`, effectively disregarding all falsy values altogether, more closely mimicking the behaviour of React.


rosskevin:
Thanks @bostrom

Omniroot:
@rosskevin any idea when the next apollo release will be, so we can take advantage of this fix?

rosskevin:
@Omniroot We don't have a timeline yet, but if you are up for using `master`, feel free to grab my package for the time being (you'll have to either alias your imports or search/replace for the time being).  https://www.npmjs.com/package/@alienfast/react-apollo

We will be discussing strategy for the next release on Friday 1/5/18 with the committers.

bostrom:
@Omniroot as a workaround, you can filter the array:

```
render() {
  return [
    renderSmth && (<Smth key="smth" />),
    <MoreContent key="morecontent" />
  ].filter(Boolean);
}
```

Omniroot:
@rosskevin any idea what the plan is for the next release?

rosskevin:
@Omniroot we are discussing on a call tomorrow, call was moved from Friday.

Omniroot:
@rosskevin any updates?

PR #1500, opened 2017-12-30
Adjust Flow types to (recent) changes

hendrikniemann:
Hey everyone,

I did some improvements to the flow types:

#### withApollo takes an additional options argument

I diverged from the [Typescript definition](https://github.com/apollographql/react-apollo/blob/master/src/withApollo.tsx#L15) here. `withApollo` seems to take all available `OperationOption`s but only uses `withRef` to determine if a reference should be passed into the child component. I created a type for the single option and used it in the type definition. I can also use the same type as in the TS definition and this can be changed in both in a separate discussion/PR, but this would introduce another type parameter to the function.

#### withApollo now correctly injects the `client` property into the wrapped component

This is basically a fix for the definition to represent the actual behaviour. I also added a test to the `flow-usage.js` file.

#### ~compose is now exported again using $Compose utility type introduced in Flow 0.55~

~The compose export was removed but has now been reintroduced using `lodash.flowright`. Flow 0.55 has the $Compose utility type that natively represents the compose function type signature.~

#### Network status is now an number literal union type representing the actual 7 network states available

The [network status](https://www.apollographql.com/docs/react/basics/queries.html#graphql-query-data-networkStatus) can be expressed more precisely using an enumeration.

#### gql is no longer exported

I removed the export since gql from graphql-tag does not seem to be exported anymore by this package and the docs also now always import it directly from graphql-tag.

PS: Sorry for everything in one commit but your precommit-hook breaks partially staged files :(


src/index.js.flow
```diff
@@ -13,11 +13,10 @@ import type {
...
 } from 'apollo-client';
 import type { DocumentNode, VariableDefinitionNode } from 'graphql';
-import gql from 'graphql-tag';
 
-// export everything from apollo-client
-export * from 'apollo-client';
-export { gql };
+export type NetworkStatus = 1 | 2 | 3 | 4 | 6 | 7 | 8;
+
+export var compose: $Compose;
```
rosskevin: We will need to remove this.  `master` no longer exports `compose`.  See the current changelog.

hendrikniemann: Thanks! I removed this from the commit.

rosskevin:
Thanks for the help with flow @hendrikniemann - we can definitely use more of that!

PR #1501, opened 2017-12-31
The mocks now take into account the specific variables used. Fixes #1412

### LINKED ISSUE #1412: MockedProvider ignores variables when doing matching
**Intended outcome:**

`MockedProvider` should serialise `query` as well as `variables` passed into it and mock requests accordingly.

**Actual outcome:**
`MockedProvider` only serialises `query` but not `variables` passed into it, causing any request with arbitrary `variables` to be matched with the mocked request with the same `query`.

The buggy code is here:

https://github.com/apollographql/react-apollo/blob/47336df2bc7f481ada6a1990a9e023677738c2c0/src/test-links.ts#L138

```
const requestKey = {
    variables: request.variables || {},
    query: queryString,
  };
JSON.stringify(requestKey, Object.keys(requestKey).sort());
```

When serialising the `requestKey`, `variables` field would get lost. 

Sample code to reproduce the issue:

```
var requestKey = {
  query: 'something',
  variables: {
    a: 'else',
  }
}

console.log(JSON.stringify(requestKey, Object.keys(requestKey).sort()));
// {"query":"something","variables":{}}
// variables empty, because a is not in the array passed to JSON.stringify as second parameter

console.log(JSON.stringify(requestKey, ['query', 'variables', 'a']));
// {"query":"something","variables":{"a":"else"}}
// variables present, because a is in the array passed to JSON.stringify as second parameter

console.log(JSON.stringify(requestKey));
// {"query":"something","variables":{"a":"else"}}
// variables present, because no filtering

```

**Version**
- apollo-client@1.9.1
- react-apollo@1.4.14

excitement-engineer:
Fixes #1412.

Note, after updating the code to fix the bug some of the tests broke. I tried to fix them as I saw fit but could someone verify that I have updated these tests correct.

### DIFF
--- a/src/test-links.ts
+++ b/src/test-links.ts
@@ -135,7 +135,8 @@ function requestToKey(request: GraphQLRequest): string {
     variables: request.variables || {},
     query: queryString,
   };
-  return JSON.stringify(requestKey, Object.keys(requestKey).sort());
+
+  return JSON.stringify(requestKey);
 }
 
 // Pass in multiple mocked responses, so that you can test flows that end up
--- a/test/__snapshots__/test-utils.test.tsx.snap
+++ b/test/__snapshots__/test-utils.test.tsx.snap
@@ -8,6 +8,16 @@ Object {
 }
 `;
 
+exports[`errors if the variables in the mock and component do not match 1`] = `
+[Error: Network error: No more mocked responses for the query: query GetUser($username: String!) {
+  user(username: $username) {
+    id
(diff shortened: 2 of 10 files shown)


rosskevin:
@excitement-engineer some `tsc` errors now - not sure if due to other PRs merging or a mistake.

excitement-engineer:
That was a mistake:) I fixed it, thanks for taking a look!

PR #1505, opened 2018-01-02
fix(index.js.flow): fix options and errorPolicy types

jedwards1211:
fixes two problems:
* def for `errorPolicy` option is missing
* the flow types required `options` to be a function returning the options,
  but it appears `withApollo` can take a plain options object as well.

### Checklist:


jbaxleyiii:
@jedwards1211 thank you so much!

PR #1519, opened 2018-01-05
Render callback should be typed with TData

rosskevin:
Missing parameterized type propagation in `Query` component


src/Query.tsx
```diff
@@ -70,7 +70,7 @@ export interface QueryState<TData = any> {
 }
 
 class Query<TData = any> extends React.Component<
-  QueryProps,
+  QueryProps<TData>,
```
excitement-engineer: Can you also update the type definition of the props in the constructor.

rosskevin: Good catch

PR #1520, opened 2018-01-05
Mutation component

excitement-engineer:
This is an implementation of the `<Mutation />` component.

Note, this is still a WIP. I will be improving it over the next few days and adding tests.


lifeiscontent:
@excitement-engineer first thing I see is being able to pass variables to runMutation. If you can take variables from both the function and props it would make for a much more powerful API because then you can partially apply the function from props and pass the rest of the variables into the function itself.

src/Mutation.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const result = notCalled
+      ? undefined
+      : {
+          loading,
+          data,
+          error,
+        };
+
+    return children(this.runMutation, result);
```
lifeiscontent: What if we add props instead of result here? That way people can still use polling functions from the component as well.

src/Mutation.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+class Mutation extends React.Component<MutationProps, MutationState> {
+  private client: ApolloClient<any>;
+
+  static contextTypes = {
+    client: PropTypes.object.isRequired,
+  };
+
+  constructor(props: any, context: any) {
```
rosskevin: `props: MutationProps<TData, TVariables>`

src/Mutation.tsx
```diff
@@ -0,0 +1,93 @@
...
+const invariant = require('invariant');
+import { DocumentNode } from 'graphql';
+
+import { OperationVariables } from './types';
+
+// refetchQueries?: string[] | PureQueryOptions[];
+// update?: MutationUpdaterFn;
+// notifyOnNetworkStatusChange?: boolean;
+
+export interface MutationProps {
```
rosskevin: `MutationProps<TData = any, TVariables = OperationVariables>`

src/Mutation.tsx
```diff
@@ -0,0 +1,93 @@
...
+  mutation: DocumentNode;
+  variables?: OperationVariables;
```
rosskevin: `: TVariables`

src/Mutation.tsx
```diff
@@ -0,0 +1,93 @@
...
+  optimisticResponse?: Object;
+  children: (mutateFn: () => void, result?: any) => React.ReactNode;
```
rosskevin: This result is the from `apollo-client` `mutate<T>(options: MutationOptions<T>): Promise<FetchResult<T>>`

Note there is a typing issue on it so it cannot yet be fully typed results: https://github.com/apollographql/apollo-client/issues/2795
  
Alternately, we could define our own `MutationResult` like we did in `Query` and solve the type problem ourselves.

src/Mutation.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    this.setState({
+      loading: true,
+      error: undefined,
+      data: undefined,
+      notCalled: false,
+    });
+
+    try {
+      const response = await this.client.mutate({
```
rosskevin: This will ultimately need parameterized types but https://github.com/apollographql/apollo-client/issues/2795 needs fixed first to use them.

lifeiscontent:
@rosskevin @excitement-engineer what do you guys think about the following?

1. use the `render` as the prop for render props.
2. use `React.children` to pass the props from `Mutate` into the `children` of `Mutate`.
3. use this pattern for the `Subscriptions` and `Query` components as well.

I see this API being a lot more flexible, because 1. you get the render prop pattern, and 2. if you need to compose React Components together, it's really straight forward.

rosskevin:
> 1. use render as the prop for render props.
2. use React.children to pass the props from Mutate into the children of Mutate.
3. use this pattern for the Subscriptions and Query components as well.

No, this will be far more confusing.  I've already made my opinion known - `children` are for rendering.  If `render` was used for all rendering in react, then yes, but that is not the case. 

Along these lines, I have discussed with @excitement-engineer and I'm not even sure there is a great use case for a `Mutation` component at all.  Given mutations aren't simply fired when they are rendered, and are usually fired in in a callback which will live in the class somewhere, I think the primary use case for mutation usage is with `withApollo` or `ApolloConsumer` to get the `client` and call `client.mutate` directly.  I've done this with our entire codebase in conjunction with using the `Query` component.

lifeiscontent:
@rosskevin if I create a demo of what I discussed in a real use case would you consider otherwise? Or should I just create my own node module for the api I am thinking about?

The pattern I’ve discussed would compose incredibly well with apollo-link-state when providing local state which will be used in turn as variables for a remote query/mutation

rosskevin:
@lifeiscontent I am not the ultimate authority on the decision, but my thoughts are centered around ability for _many_ to understand.  What is clear is that _many_ would not comprehend passing parameters as children, therefore it would be a docs/education problem and would create more issues.

I suggest wrapping it as you desire.  I have a wrapper on `Query` to tweak a few things to suit our opinions that aren't suitable for maney as well.

kandros:
Might be worth to take a look at https://github.com/apollographql/react-apollo/pull/1550#issuecomment-360309478

Currently  what is the state of this?

I would lile to help

excitement-engineer:
@kandros I have been on holidays for the past few weeks. But I will be picking this PR up again in a couple of days when I get back:)

lifeiscontent:
@excitement-engineer when can I test the latest release? e.g. `Mutate`, `Query`, etc.

kandros:
You can use Query by installing react-apollo@beta Mutation is WIP

kandros:
@excitement-engineer totally personal feedback: I think `onComplete` and`onError` are against the purpose of using a function renderer, I find them redundant and not clear

excitement-engineer:
@kandros I think there is a use case for performing some custom operation as soon as the mutation is done or there is an error (e.g. updating the state or calling a callback in the props). This cannot be done using the render prop AFAIK. Therefore the `onError` and `onComplete` serve a different function from the render prop.

kandros:
Since onComplete just pass data and onError pass the error, can’t we estimate the completion state by the loading prop and error state by state prop in the renderer? Am I missing some specific usecases?

kandros:
I re read your answer, I get it now 👍🏻

src/Mutation.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  componentWillReceiveProps(nextProps, nextContext) {
+    if (
+      shallowEqual(this.props, nextProps) &&
+      this.client === nextContext.client
+    ) {
+      return;
+    }
+
+    this.verifyDocumentIsMutation(nextProps.mutation);
```
leoasis: Do we want to do this only if the mutation is different from the previous one? It involves calling `parser` which I think does a couple things that may add unnecessary overhead if the mutation doesn't change, which should be most of the time

excitement-engineer: Nice! good tip

src/Mutation.tsx
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private runMutation = async () => {
+    this.onStartMutation();
+    
+    this.mostRecentMutationId = this.mostRecentMutationId + 1;
+    const mutationId = this.mostRecentMutationId;
+    
+    try {
+      const response = await this.mutate();
```
kiurchv: Could we return the result when `this.mutate` was resolved?

src/Mutation.tsx
```diff
@@ -0,0 +1,237 @@
...
+      this.onCompletedMutation(response, mutationId);
```
leoasis: I think we should run this outside of the try/catch. Otherwise, if there's an error inside this method (which is likely since it calls user code), it would trigger the mutation error path, which would be incorrect.

src/Mutation.tsx
```diff
@@ -0,0 +1,237 @@
...
+  data: TData;
+  error?: ApolloError;
+  loading: boolean;
+}
+
+export interface MutationProps<TData = any, TVariables = OperationVariables> {
+  mutation: DocumentNode;
+  variables?: TVariables;
+  optimisticResponse?: Object;
+  refetchQueries?: string[] | PureQueryOptions[];
```
leoasis: @jbaxleyiii if we want to get rid of the query recycler, should we also get rid of the string array in this option? This would look up the currently watched queries by name, which seems to be the same thing that the deprecated `updateQueries` does.

src/Mutation.tsx
```diff
@@ -0,0 +1,237 @@
...
+  update?: MutationUpdaterFn;
```
leoasis: I'm wondering if we should improve the type of the updater function so that it is parametric on `TData`, so that the second argument provides the result in the shape you'd expect, instead of an object of any keys of any value.
We could do this here in react apollo, the same way we did it to improve some of the `Query` types, and then we can send some PRs to correctly fix this in apollo client.

src/Mutation.tsx
```diff
@@ -0,0 +1,237 @@
...
+    
+    const {
+      mutation,
+      variables,
+      optimisticResponse,
+      refetchQueries,
+      update,
+    } = this.props;
+
+    const response = await this.client.mutate({
```
lifeiscontent: why not return the await'd mutate function instead of creating a variable?

src/Mutation.tsx
excitement-engineer: Good point. Thanks, updated it.

src/Mutation.tsx
excitement-engineer: I tried to improve on the type and made it parametric on TData, the second argument is now fully typed. The issue with this is that it is not so easy to to fix this issue. I had to override the type definition for `ExecutionResult` which is contained in graphql-js. So we will have to update the type in that repo to create a long-term sustainable solution I think,

thomasdashney:
I agree with @lifeiscontent - I think it's important for the `mutate` function to take mutation variables for the following reasons:
* It's very possible that the variables come from a child component (e.g. using the [Formik](https://github.com/jaredpalmer/formik) library):
```js
<Mutation mutation={mutation}>
  {({ mutate }) => (
    <Formik onSubmit={(formValues) => }
      {/* etc */}
    </Formik>
  }}
</Mutation>
```
* The existing HOC version's `mutate` function also takes variables.

jbaxleyiii:
@excitement-engineer I think the only question remaining here is the usage of variables. Is it better done as a prop or passed to the function or both (i.e. assign). This is def a case where I think the prop remapping of the HOC gets it right and would be nice to replicate the ideas of that here in some manner?

JonathanZWhite:
@excitement-engineer I'm really excited about your work on the Mutation Component! Will this be included in the non-beta 2.1 release? 🎉

TheGrimSilence:
@JonathanZWhite I'm hoping

excitement-engineer:
Hey @JonathanZWhite, nice to hear that you like the component!  As far as I'm aware this will not be included in the 2.1 release. We will however do our best to get it in your hands as quickly as possible:)

JonathanZWhite:
@excitement-engineer Awesome! Thanks for letting me know 🤘

excitement-engineer:
I updated the PR! It is now possible to pass variables as options in the mutation function.

```js
<Mutation mutation={mutation}>
      {(createTodo, result) => {
          setTimeout(() => {
            createTodo({ variables });
          });
        return <div />;
      }}
    </Mutation>
```

src/Mutation.tsx
excitement-engineer: Good point! I updated the PR:) Thanks

jbaxleyiii:
@excitement-engineer I've merged this into a branch on the upstream so I can finish it out! :yay:

JonathanZWhite:
@jbaxleyiii @excitement-engineer Do you see mutation components replacing the HOC mutation pattern completely? Or are there still going to be use-cases for both? In the latter case, what are some good guidelines to figure out when a component should be used vs a HOC?

Issue #1385, opened 2017-12-06
Loading state stuck on true and network request is never made


nlucero:
I'm having an issue using the `graphql` HOC. I'm receiving via props on my wrapped component an object with the variables that are used on the graphql query (let's call it the `filters` object). When an error occurs on the API, the `data { error }` prop is injected correctly into my wrapped component and contains the error that comes on the response. But, after this, when the query is re-run with different `variables` (due to the `filters` object being updated), `graphql` makes NO NETWORK REQUEST at all, and the `data { loading }` prop is stuck on `true` forever.

I've even tried setting the `errorPolicy` to `all`. But in that case, the `data { error }` prop always come as `undefined` (even when there are errors on the API response). Not sure if this is intended or another bug :stuck_out_tongue:

Here is my wrapped component (I'm using [recompose](https://github.com/acdlite/recompose)):

```
const withData = graphql(caseInfoQuery, {
  options: ({ filters }) => ({
    variables: filters
  })
});

const enhance = compose(
  withData,
  defaultProps({ data: {} }),
  withProps(({ filters, data: { cases, loading, error } }) => ({
    cases,
    loading
  }))
);

export default enhance(MyDataTable);
```

I'm using these package versions:

```
"dependencies": {
    "apollo-cache-inmemory": "^1.1.3",
    "apollo-client": "^2.1.0",
    "apollo-link": "^1.0.5",
    "apollo-link-error": "^1.0.2",
    "apollo-link-http": "^1.3.0",
    ...
}
```

nlucero:
I created the [same issue](https://github.com/apollographql/apollo-client/issues/2687) on the apollo-client repo, but I think it belongs here.

nfantone:
@jbaxleyiii I think this is related to the `queryObservable` object on the HOC. On initial load, [when requesting the `currentResult()`](https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L539), errors are returned as expected. However, after any error (network or graphQL originated), that method *always* returns:

```json
"{"data":{},"loading":true,"networkStatus":1,"partial":true}"
```

nfantone:
I can also confirm that this is still happening on `master` code from both this component and `apollo-client`.

mfurniss:
I'm having exactly this issue. After a data error, if the same query is reissued then the data loading flag never clears. apollo-client 2.1.0 with react-apollo 2.0.4

nfantone:
[...]

PR #1531, opened 2018-01-09
Resubscribe after error -- apollo-client issue 2513

wdimiceli:
This is a fairly minimal fix for React components ceasing to get updates after an error occurs in a query.  It happens due to the way observables work.. they will terminate a subscription after the error handler is called.  Without going through and reconsidering all of the subscription logic between react-apollo and the client, simply resubscribing on error seems like the simplest way to get it working again.  A unit test was written to cover this scenario.

Here is the most relevant issue:
https://github.com/apollographql/apollo-client/issues/2513
I have seen several others which might be resolved with this fix.


rosskevin:
@excitement-engineer if this looks good, we need to mirror the tests and changes to `Query`.

@wdimiceli if you want to increase bundle maxSize in the package.json and push it that will solve the build error.  4.8 will be fine.

edorivai:
Is there a chance this will be released on NPM any time soon? Pretty, pretty please :smile:

rosskevin:
Please check the releases tab before posting.  We are going to try and release more frequently and this most certainly has been released (though strangely danger did not require a changelog entry for it).

Try the 2.1 beta: https://github.com/apollographql/react-apollo/releases

edorivai:
@rosskevin Sorry, wasn't aware of the beta releases, just checked npm stable. Thanks!

nfantone:
Unfortunately, this didn't resolve https://github.com/apollographql/react-apollo/issues/1385 for me 😞 . However, installing the beta release of `react-apollo` + latest `apollo-client@2.2.2` + `apollo-cache-in-memory@1.1.7` make requests work again after an error and the `loading` was being set correctly.

AndrewHenderson:
@nfantone I have upgraded to all of those versions with `react-apollo@2.1.0-beta.2`.

I'm finding that when their is an error in a component, the loading prop is false, however none of the props are passed to the component.

I'm using a sub-resolver and testing when there is an error in that single function.

Everything works as expected in GraphQL – when there is an error, that single property is `null` and the GraphQL error is returned on the response.

However, because there is an error, none of the parent props, the partial result that returned, gets passed to the component 😕

edorivai:
@AndrewHenderson I might be misunderstanding, but partial results sound like a job for `errorPolicy`: https://www.apollographql.com/docs/react/features/error-handling.html#policies

AndrewHenderson:
@edorivai That’s exactly the setting I needed. I completely missed that. Thank you! 😁

nfantone:
@AndrewHenderson Actually, what I'm seeing is that now the `loading` and `error` properties are working fine, but - on error, `data` continues to hold the _latest correctly fetched_ data, instead of what was actually returned by the `/graphql` response.

@edorivai Would you have any input on this? Would this be the intended functionality?

AndrewHenderson:
@edorivai I tried using errorPolicy, but I’m still experiencing what @nfantone has described.

I thought it might work using the new Query component, but don’t see an errorPolicy propType for that new component.

edorivai:
Hey guys, sorry to disappoint you, but I'm no authority on this lib 😅,
just another user.

@AndrewHenderson I didn't try it with the Query component yet. But I would
expect it should be possible to implement if it isn't yet.

tmeasday:
I could be misunderstanding but I suspect this does not work with polling queries (I haven't been able to make it work), because `apollo-client` does not [resume polling after a network error](https://github.com/apollographql/apollo-client/issues/1099#issuecomment-299603974)

Perhaps if the query has `pollInterval` set `react-apollo` should also re-create the `observableQuery` on error also?

404sand808s:
Hi guys, checking in to see if anybody knows when this might be released — desperately need this to re-establish connection after an error (to give the user a way to reload). Thanks, and sorry for the bump.

iamrommel:
This is really breaking, we are are using react-native and apollo, and whenever i go to elevator (which most of the time has not data connection), my app gets error, and even when i get back to wifi connection, my connection cannot be recovered. Correct data are being pull from the server, but the component are not updated

Issue #1567, opened 2018-01-22
Error in Firefox are incomplete


nchevobbe:
Hello, Firefox DevTools' Console maintainer here.
I've been warned on [twitter](https://twitter.com/mxstbr/status/954306693009166336) that the Firefox console is mostly useless when working with `react-apollo`.

I dug a bit and think I found the root of the problem: Apollo is using `error.stack` to log errors:
https://github.com/apollographql/react-apollo/blob/e3f98a8273fab29d39f98f6bdd5fe8b2e5fb8bc0/src/graphql.tsx#L567-L570

The issue is that `error.stack` is not standard (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack), and differs in different browsers ("pure" stack only in FF, error message + stack in Chrome, something totally different in Safari, …).

It would be really nice if we could show the same message in every browser.
I guess we could check  that `error.message` is included in `error.stack`, and if not, append it to the output.

nchevobbe:
Chrome:
<img width="730" alt="chrome" src="https://user-images.githubusercontent.com/578107/35223441-abeeb1da-ff81-11e7-9315-7f27a96d6192.png">

Safari:
<img width="624" alt="safari" src="https://user-images.githubusercontent.com/578107/35223447-b0b5f868-ff81-11e7-8841-e4a750fd8063.png">

Firefox:
<img width="799" alt="ff" src="https://user-images.githubusercontent.com/578107/35223452-b526e146-ff81-11e7-949a-21a79aef60e5.png">

jbaxleyiii:
❤️ thanks so much! Merged! It'll be out in the next release!

mxstbr:
Yess, that is awesome! Thank you @nchevobbe for digging into this :heart:

hwillson:
Fixed in https://github.com/apollographql/react-apollo/pull/1568 - thanks!

PR #1568, opened 2018-01-22
Fix error messages for non-Chrome browsers.

nchevobbe:
Check if error.stack contains error.message, and if not, append it to the error message
printed to the console.
This should fix error messages in Firefox and Safari.


nchevobbe:
Not sure about this change since I am not familiar with the codebase.
Don't hesitate to tell me what's wrong with this so I can fix it.

nchevobbe:
This fixes https://github.com/apollographql/react-apollo/issues/1567

jbaxleyiii:
@nchevobbe this is fantastic! Thanks so so much!

PR #1576, opened 2018-01-25
Package now consists of ES2015 and commonjs

rosskevin:
As discussed in `apollo-maintainers` target package creation to support `main` as commonjs and `module` as ES2015.

/cc @jaydenseric


package.json
```diff
@@ -90,28 +104,26 @@
     "babel-core": "6.26.0",
     "babel-jest": "22.1.0",
     "babel-preset-env": "1.6.1",
-    "bundlesize": "0.15.3",
```
jbaxleyiii: @rosskevin why are we removing bundlesize?

package.json
```diff
@@ -90,28 +104,26 @@
...
     "preact": "^8.2.7",
     "preact-compat": "^3.17.0",
     "prettier": "1.10.2",
     "react": "16.2.0",
     "react-dom": "16.2.0",
     "react-test-renderer": "16.2.0",
     "recompose": "0.26.0",
+    "recursive-rename": "^2.0.0",
     "rimraf": "2.6.2",
-    "rollup": "0.54.1",
```
jbaxleyiii: do we not need this anymore? Does typescript handle all of the bundling?

rosskevin: tsc does 100% of the transpilation - no more bundles (discussed with @jaydenseric in the thread)

package.json
rosskevin: We aren't creating a UMD bundle anymore, not sure if we can size a non-compressed folder of files.  I'll check.

package.json
```diff
@@ -5,37 +5,40 @@
   "module": "src/index.ts",
```
jaydenseric: How does this work, does `main` and `module` get added to the package.json that actually gets published?

rosskevin: ugh, that's no good

rosskevin: wait, no, this is ok - it's there for codesandbox.  Pay no attention to the /package.json, just to lib/package.json

package.json
rosskevin: Added back to test rollup-produced CJS file

package.json
rosskevin: rollup added back for bundlesize only

jbaxleyiii:
L:ooks like the bundlesize calculation is wrong but I'll fix that on master

rosskevin:
I confirmed this PR locally with a package build using ts-loader

jaydenseric:
I'm confused, I'm looking at what is published in `react-apollo@^2.1.0-beta.0` and it is nothing like discussed in Slack:

screen shot 2018-01-29 at 3 51 04 pm

It should be:

```json
{
  "main": "index",
  "module": "index.mjs"
}
```

We discussed minification and bundling, and agreed that it would only happen to test bundle size but no UMD or AMD would be distributed. Currently the `main` points to a minified `react-apollo-umd.js`!

Also, many of the .mjs files are invalid as they have `require` in them, see https://github.com/apollographql/react-apollo/issues/1589.

I'm trying to work out what has been going on looking at the release descriptions, and they don't make sense. [v2.1.0-alpha.1](https://github.com/apollographql/react-apollo/releases/tag/v2.1.0-alpha.1) and [v2.1.0-alpha.2](https://github.com/apollographql/react-apollo/releases/tag/v2.1.0-alpha.2) have the exact same bullet list of changes.

No release mentions `.mjs` anywhere, which it should because it's a pretty big deal.

rosskevin:
@jaydenseric @jbaxleyiii changed the `main` in https://github.com/apollographql/react-apollo/pull/1578.

You should look at this changeset, it is exactly as discussed.

I'll place the rest of my comments in #1589.

PR #1577, opened 2018-01-25
Only call fetchData on Apollo components

### LINKED ISSUE #1571: SSR should check that instance belongs to Apollo before calling `fetchData`
**Intended outcome:**

`fetchData` should not be called on components that do not belong to `react-apollo` during server render.

**Actual outcome:**

If the component is not expecting `fetchData` to be called, SSR will fail. Right now, when walking the tree, `react-apollo` only checks if an instance exists and that it has a function called `fetchData`:

https://github.com/apollographql/react-apollo/blob/9222ab3c003467bd0b0e1bf47b3ba50dc6e32840/src/getDataFromTree.ts#L139-L154

That is problematic if the instance has a function called `fetchData` that should not be called during server rendering.

Add something along the lines of the following to a component being server rendered:

```javascript
class MyComponent extends React.Component {
  fetchData() {
    throw new Error('fetchData should not be called');
  }

  render() { return null; }
}
```

**Version**
- apollo-client@2.0.3
- react-apollo@1.4.15

asmockler:
Resolves #1571 .

When walking the tree on the server, `react-apollo` was calling `fetchData` on any component that had a function called `fetchData`. For me, this was problematic as I have other components with functions called `fetchData`. During SSR, those functions would be called erroneously and would often throw errors.

This PR adds a check to ensure a component is an apollo `GraphQL` component before calling `fetchData`. The check isn't 100% bulletproof since it depends on the class name which is not guaranteed to be used only by `react-apollo`. If anyone has a better suggestion, I would love to hear it :)

### Checklist:

### DIFF
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -138,7 +138,12 @@ function getQueriesFromTree<Cache>(
 
   walkTree(rootElement, rootContext, (element, instance, context) => {
     const skipRoot = !fetchRoot && element === rootElement;
-    if (instance && typeof instance.fetchData === 'function' && !skipRoot) {
+    if (
+      instance &&
+      typeof instance.fetchData === 'function' &&
+      instance.constructor.name === 'GraphQL' &&
+      !skipRoot
+    ) {
       const query = instance.fetchData();
       if (query) {
         queries.push({ query, element, context });
--- a/test/server/getDataFromTree.test.tsx
+++ b/test/server/getDataFromTree.test.tsx
@@ -883,6 +883,43 @@ describe('SSR', () => {
       });
     });
 
+    it("shouldn't call `fetchData` on components that are not apollo components", () => {
+      const query = gql`
+        {
+          currentUser {
+            firstName
+          }
+        }
+      `;
+      const link = mockSingleLink({
+        request: { query },
+        result: { data: { currentUser: { firstName: 'James' } } },
+        delay: 50,
+      });
+      const apolloClient = new ApolloClient({
+        link,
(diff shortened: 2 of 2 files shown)


jbaxleyiii:
@asmockler haha that is actually by design! The `getDataFromTree` is a generic util function bringing async two pass rendering to React

asmockler:
Hm interesting. What would you suggest here? It seems a little odd (and was very hard to debug) that `react-apollo` would be reaching inside my components, and as far as I can tell this behavior isn't documented anywhere. I definitely see why it would be useful, though. I could add a note to the docs or would be happy to help pull `getDataFromTree` into its own package where that behavior could be highlighted better. 

We fixed this in our app by renaming the method, so if this doesn't fit with all things Apollo, I'm also happy for this to be closed and move on with other things :)

leoasis:
@jbaxleyiii I think maybe the name that we look for is too common maybe? If we would call it `apolloFetchData` or something, people could opt in into this mechanism intentionally rather than by mistake, because of the name. The problem with this is that it'd be a breaking change.

jbaxleyiii:
@leoasis @asmockler I'd prefer to actually extract the two pass API and make it its own package. Its one of the cooler things about this lib and I'd really like to keep it as `fetchData` as there are usages outside of just apollo.

So I guess I'd probably say `wont-fix` on this one? Sorry for any problems this causes!

asmockler:
Yeah I agree! Thanks for much for taking the time to look into this :) feel free to ping me if you want help extracting that stuff – I would be happy to help.

PR #1578, opened 2018-01-25
Browser

jbaxleyiii:
@rosskevin in my testing the inclusion of react-dom blew up ReactNative and some older build systems. Also, without a single bundle, we can't support usage of unpkg.org which I'd like to keep for people using it!

This also fixes the precommit hooks with husky which weren't working previously and removes using `yarn` since its not a required part of the node ecosystem


rosskevin:
@jbaxleyiii it looks like you removed husky - which is the `precommit` hook, and added a script called `precommit`, which is not using git hooks.  So we have no more git hook.  What am I missing?

package.json
```diff
@@ -3,48 +3,45 @@
   "version": "2.1.0-alpha.0",
   "description": "React data container for Apollo Client",
   "module": "src/index.ts",
+  "browser": "lib/react-apollo.browser.umd.js",
```
FredyC: @jbaxleyiii I don't understand why this change was necessary. Doesn't it mean that eg. Webpack cannot do tree shaking? Besides it's broken altogether :) #1591 and this is from debugger

image

PR #1579, opened 2018-01-25
Add SSR support for Query component

leoasis:
This PR adds support for `Query` component to be prefetched via `getDataFromTree`. Also added an optional `ssr` prop (same as the option in the `graphql` HoC) so that we're able to skip the component during the prefetching process.


jbaxleyiii:
@leoasis haha amazing! I just took a stab at it too but this is better! :yay:

PR #1586, opened 2018-01-27
Update typescript example for 2.1

kandros:
`npm start` now does work

`react-scripts-ts` no longer bundle typescript so it needs to get installed manually (maybe is a problem in their init script,  investigating...)

the example was named typescript in package.json so i could not install typescript (npm limitation warning), thus the package is not named typescript-example

the typings generations with target typescript-modern using apollo-codegen is broken due to one one his dependencies, i changed it to just typescript


examples/typescript/src/Character.tsx
```diff
@@ -1,26 +1,30 @@
...
+  GetCharacterQuery,
+  GetCharacterQueryVariables
+> {}
 
 export interface CharacterProps {
-  episode: string;
+  episode: Episode;
 }
 
 export const Character = (props: CharacterProps) => {
```
magicmaaaaan: `Character` should be explicitly typed as `React.StatelessComponent`.

examples/typescript/package.json
```diff
@@ -30,6 +32,6 @@
     "schema":
       "apollo-codegen introspect-schema https://mpjk0plp9.lp.gql.zone/graphql --output ./src/schema.json",
     "types":
-      "apollo-codegen generate --schema schema.json --target typescript-modern `find ./src -name *.graphql`"
+      "apollo-codegen generate ./src/**/queries.ts --addTypename --schema ./src/schema.json --target typescript --output ./src/__generated__/types.ts"
```
rosskevin: We'd rather be using `typescript-modern` here - as `typescript` is going to be deprecated and all efforts are in the `modern` variants.  I know it doesn't say that anywhere but I discovered it in chat.

kandros: it is brocker right now, bacause of a dependency on a beta version on babel types

examples/typescript/src/Character.tsx
kandros: good catch

examples/typescript/package.json
rosskevin: Ah, got it.  Yeah I'm on a private release.

jbaxleyiii:
@kandros this is great!

@rosskevin a move to lerna would make it possible for these to use the symlinked version and building it / testing it could be part of the CI process. That would be much better than keeping track of versions / letting these become stale.

PR #1587, opened 2018-01-27
Added prop-types validation to unchecked components.

excitement-engineer:
Added `prop-types` validation to the `<Query />`, `<Subscription />` and `<ApolloConsumer />` component


jbaxleyiii:
ugh, I hate prop-types in libraries but since we use context I guess we have them already? With the move to react's new context though I wonder if depending on `prop-types` will be necessary? If not, I'd like to not ship them anymore?

Thoughts? Are they still used / valuable?

excitement-engineer:
I personally don't use them as I use flow on all my react projects. However, for those that don't use any typings they could be valuable. I know that react-router uses them for example.

PR #1594, opened 2018-01-30
Fix queryErrors in getDataFromTree

silverlyra:
The code in `getDataFromTree` which tries to build a composite Error of all encountered query errors does not work correctly. The final error thrown from the Promise will have the right message, but `queryErrors` will only have one item – that points to the composite error itself:

    { Error: 3 errors were thrown when executing your GraphQL queries.
        at react-apollo.umd.js:763:19
        queryErrors: [ [Circular] ] }

Upon investigating, the `catch` within `queries.map` was catching the composite error thrown within the recursive call to `getDataFromTree`. It's dutifully pushed onto the local `errors`, and then its `queryErrors` property is overwritten within the `Promise.all` `then` handler.

Using the two-argument form of `then` instead of a `then`/`catch` pair solves the issue.

### DIFF
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -169,8 +169,10 @@ export default function getDataFromTree(
   const mappedQueries = queries.map(({ query, element, context }) => {
     // we've just grabbed the query for element, so don't try and get it again
     return query
-      .then(_ => getDataFromTree(element, context, false))
-      .catch(e => errors.push(e));
+      .then(
+        _ => getDataFromTree(element, context, false),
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@silverlyra amazing! Can you sign the CLA for this?

silverlyra:
yep! so I figured out how to add a test that would get multiple `queryErrors`, but the test passes on master too 😐

any ideas how to write a better test case?

jbaxleyiii:
@silverlyra hmmm, I'm not sure. I'm surprised this passed on master

Issue #1595, opened 2018-01-30
Problem in getDataFromTree, React Component detection


edorivai:
**Intended outcome:**

`getDataFromTree` should not crash when passed a valid React tree.

**Actual outcome:**

`getDataFromTree` crashes when the `<Link>` component from `react-scroll` is rendered.
Error: `error: Cannot call a class as a function TypeError: Cannot call a class as a function`

@leoasis was quick to respond in the Slack channel, and we narrowed it down to the following commit:
https://github.com/apollographql/react-apollo/commit/1dff95968d1dad99ee91665c7c0a5c35274e0d22
in combination with the following line from `react-scroll`:
https://github.com/fisshy/react-scroll/blob/master/modules/components/Link.js#L7

The commit checks for `Component.prototype.render`, however, in `react-scroll` the render function is defined on the instance, instead of the prototype.

Something along the lines of this works in the case of `react-scroll`:
```js
if (Component.prototype && (Component.prototype.render || Component.prototype.isReactComponent)) {
```

But I have no experience with preact, so not sure if this covers all of the possible cases.

**Version**

* react-apollo@2.1.0-beta.0

jbaxleyiii:
crazy! Seems like a good change. Would you be open to a PR for it?

developit:
The suggestion would cover all cases.

edorivai:
Should be fixed in #1604

PR #1604, opened 2018-01-31
Fix React Component detection

edorivai:
Solves https://github.com/apollographql/react-apollo/issues/1595

@jbaxleyiii @leoasis

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ### vNext
 
+* Fix React Component detection bug in `getDataFromTree` [#1604](https://github.com/apollographql/react-apollo/pull/1604)
+
 ### 2.1.0-beta.0
 
 * Beta release of all 2.1 features!
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -59,7 +59,7 @@ export function walkTree<Cache>(
 
     // Are we are a react class?
     //   https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L66
-    if (Component.prototype && Component.prototype.render) {
+    if (Component.prototype && (Component.prototype.render || Component.prototype.isReactComponent)) {
       // typescript force casting since typescript doesn't have definitions for class
       // methods
       const _component = Component as any;
--- a/test/server/getDataFromTree.test.tsx
+++ b/test/server/getDataFromTree.test.tsx
@@ -268,6 +268,20 @@ describe('SSR', () => {
         );
         expect(elementCount).toEqual(9);
       });
+
+      it('basic classes with render on instance', () => {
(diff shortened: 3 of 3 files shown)


leoasis:
@edorivai mind adding a test for this? It would be super useful so that we don't cause a regression accidentally.

Also, add an entry in the changelog that mentions this fix.

edorivai:
@leoasis 2 questions:

First, where in the changelog should I add the entry? Under `2.1.0-beta.0`, or rather somewhere else?

Second, I'm having a bit of trouble with writing the test, I tried just copying over the `basic classes` test, and changing the `render` to be on the instance, instead of on the prototype:

```typescript
      it('basic classes with render on instance', () => {
        let elementCount = 0;
        class MyComponent extends React.Component<any, any> {
          render = () => {
            return <div>{_.times(this.props.n, i => <span key={i} />)}</div>;
          }
        }
        walkTree(<MyComponent n={5} />, {}, element => {
          elementCount += 1;
        });
        expect(elementCount).toEqual(7);
      });
```

However, Typescript is too smart, and points out the "mistake": 
```
TS2424: Class 'Component<any, any>' defines instance member function 'render', but extended class 'MyComponent' defines it as instance member property.
```

I tried just rendering the `<Link/>` element from `react-scroll`, which does the job. But I can imagine you'd prefer not to add `react-scroll` as a dev dependency.

Do you have any idea how to neatly test this?

leoasis:
> First, where in the changelog should I add the entry? Under 2.1.0-beta.0, or rather somewhere else?

Under **vNext**

> However, Typescript is too smart, and points out the "mistake":

In this case I think it's fair to bypass Typescript since this is a use case for plain JS users:

```ts
class MyComponent extends (React.Component as any) {
  render = () => {
    return <div>{_.times(this.props.n, i => <span key={i} />)}</div>;
  }
}
```

leoasis:
@edorivai awesome work, thanks!

edorivai:
:tada:

@leoasis final question, is there a plan for when a new release will be cut? We've currently disabled SSR on a couple pages until this patch lands. But if it will take a couple days before a release, we might have to look at alternative temporary solutions.

Thanks again :smile:

leoasis:
The idea would be to release in the next couple days, so yeah it should be soon.

Regarding a workaround, why not try defining the render method in the prototype for now? You can define it as an empty function, it won't matter since the other one is defined in the instance and will be looked at first:

```js
import { Link } from 'react-scroll';

Link.prototype.render = () => {};
```

I think that should be enough

edorivai:
@leoasis Heh, that would have been such a nice hack. Next couple days should be fine.

leoasis:
@edorivai this is now released in 2.1.0-beta.1! https://github.com/apollographql/react-apollo/blob/master/Changelog.md#210-beta1

PR #1620, opened 2018-02-02
Provide the last mapping results to the result `props` mapping function

jamesreggio:
This allows the `props` mapping function to examine and use its prior results which, when appropriate, enables it to preserve referential equality.

An alternative approach would be to allow `operationOptions` to be an _object or function_, the latter which could be used to create a closure per instantiation of the HoC for the housing of state. This is the approach embraced by Redux, as explained in the _Notes_ for `mapState` and `mapDispatchToProps`, [here](https://github.com/reactjs/react-redux/blob/master/docs/api.md#arguments).

Instead, I chose to just record and supply the prior value to the `props` mapping function. It seemed easier to explain in docs, and the `props` function is likely the be the only beneficiary from preserving cross-render state.


jbaxleyiii:
@jamesreggio very interesting! I'm assuming this is critical for react native? Would this also be useful in the new Query component?

jamesreggio:
Critical is a strong word — as you know by now, I'm a little crazy for cheap-and-accurate referential equality checks.

I don't think it's useful in the new Query component since the component doesn't have the same notion of a prop mapper. (Instead, you just receive the data in your callback and it's up to you to cache the outputs of whatever transformations you perform, if any.)

jamesreggio:
Thanks for merging, though!

Issue #1589, opened 2018-01-29
`require` in .mjs files


jaydenseric:
It is invalid for `require` to appear in .mjs files. This is currently happening in distributed files:

<img width="503" alt="screen shot 2018-01-29 at 3 39 59 pm" src="https://user-images.githubusercontent.com/1754873/35493895-be7a1d92-050a-11e8-8f53-3efb8435fb27.png">

<img width="533" alt="screen shot 2018-01-29 at 3 35 59 pm" src="https://user-images.githubusercontent.com/1754873/35493797-2782c43e-050a-11e8-8753-145fd5a20f08.png">

**Version**

* react-apollo@^2.1.0-beta.0

rosskevin:
Regarding `require`, there are a few in the codebase for legacy imports, where that is the only way to typescript-appropriately import, other than that the rest are `import` syntax.

`import x = require()` syntax won't work based on our tsconfig `module`, but `import * as x` or `import x` also won't work because these are commonjs modules that use `module.exports = x`.

If you can figure out how to import these and have `tsc` be happy - I'm happy to know that solution.

rosskevin:
It looks like #1617 will solve the `require`s.

jaydenseric:
Kind of related: https://github.com/babel/babel/issues/7294. There is [a workaround](https://github.com/jaydenseric/apollo-upload-server/commit/7e7ff92236b97a56294d2e5fb2440368264cca1a#diff-ad63aa4868d57aa47036f9e332c67c0fR1) though I have shipped in `apollo-upload-server`, to get native ESM and fallback CJS build targets to work from a common ES6 source that imports a third party CJS module.

leoasis:
Does .mjs allow importing common js .js files? If the answer is yes, then maybe we could wrap the packages and provide a default (or named) export, something like this:

```js
// some_file.mjs
import invariant from './invariant';
...
```

```js
// invariant.js
exports.default = require('invariant');
```

hwillson:
`.mjs` support was removed. Thanks!

AgronKabashi:
This is unfortunately still an issue when bundling with Rollup.
I don't see why the change was reverted as apollo-client was suffering from the same issue but they got it working here: https://github.com/apollographql/apollo-client/pull/4158/files.

rosskevin:
We have a new esm build already merged and will be released next.

PR #1621, opened 2018-02-02
Rollback importing non esm packages

leoasis:
This PR rolls back the imports of non esm packages. This breaks the bundles as they try to import those and access their `default` property.


rosskevin:
Fixes broken release, but reopens #1589

Issue #1389, opened 2017-12-08
errorPolicy not working


Torsten85:
`graphql` not passing the errors from `errorPolicy: 'all'` to the props.

As described here: https://www.apollographql.com/docs/react/features/error-handling.html#policies one would assume that there is a `errors` field next to `data` when a graphql error occures and the errorPolicy is set to all.

**Intended outcome:**
There is a error field next to data like this:
```
const withDataAndErrors = graphql(MY_MIXED_QUERY, {
  options: {
    errorPolicy: 'all'
  },
  props: ({ data, errors }) => {
     // we get data and errors (when they occure) 
  }
});
```

**Actual outcome:**
There is no query data and also no errors 

That there is no data at all is related to https://github.com/apollographql/apollo-client/issues/2703

**Version**
- apollo-client@2.1.0
- react-apollo@2.0.1

AndrewGibson27:
Replicable via [this repository](https://github.com/AndrewGibson27/react-apollo-refetch-error-example).

nfantone:
Agree with above ☝️ . Either the wording on the docs is confusing of this is a bug.

rosskevin:
PRs are welcome everyone - all help is appreciated.

nfantone:
@rosskevin Believe there needs to be, at least, some kind of triage before someone puts effort into building/fixing something. Is this really a bug? Is it expected behaviour? Is it documented anywhere?
 Why is it occurring? Who can give support?

rosskevin:
@nfantone I understand the frustration, but this is open source.  I'm a volunteer as are others, so we have to fix our own problems.  I'm just now getting involved and trying to help with maintenance.  Given this has a reproduction case and the `intended outcome` does not match the `actual outcome` - it seems like it may be a bug - but unfortunately I don't have time to dive into that on your behalf.

nfantone:
I know where you're coming from. I understand. I'm lucky enough to be the maintainer of a couple of open source projects myself. And I know it's not always an easy task and it demands time.

However, it's not on _my_ behalf (even though I'm not even the one who opened the issue). It's on behalf of the _project_. This discussion has been around forever. I honestly believe your mental model should stay away from "this is open source, send a PR, I don't have time". You cannot expect people to help you out if you don't pour that same effort yourself. It's a matter of synergy and feedback loop.

I appreciate you taking the time to answer here, since no one else did.

afonsodelgado:
+1

amok:
[...]

PR #1630, opened 2018-02-05
TASK: Provide errors to component when error Policy is "all"

Torsten85:
The docs say that "any errors reported will come under an error prop along side the data returned from the cache or server."  (https://www.apollographql.com/docs/react/features/error-handling.html)

This PR will provide those errors to the component.

To make it work, another PR is mandatory: https://github.com/apollographql/apollo-client/pull/2956

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -369,7 +369,7 @@ export default function graphql<
 
         const resultProps = { [name]: defaultMapResultToProps(result) };
         if (hasErrors) {
-          resultProps.errors = errors;
+          resultProps.errors = newResult.errors;
         }
 
         return resultProps;
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import ApolloClient, {
+  ApolloError,
   ObservableQuery,
   ApolloQueryResult,
 } from 'apollo-client';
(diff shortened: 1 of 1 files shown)


Torsten85:
Resolves #1389

twelve17:
@Torsten85 does [this line](https://github.com/Torsten85/react-apollo/blob/2c2e542e81a042961e7f3c1993bdc39d391c238f/src/graphql.tsx#L600) also need to be changed to look for `errors` (plural)?

```
const { loading, error, networkStatus } = currentResult;
```

Torsten85:
@twelve17 No, I'm getting the `errors` directly from the currentResult inside `calculateResultProps` because `errors` and `data` are along side each other.

Thats how I assume the docs would want it ;) 

If `errors` should be within `data`, then the line you mentioned would be the point to add it. But this would result in `data.error` and `data.errors` which seems wrong.

twelve17:
@Torsten85 I don't believe `currentResult` is `data`, but the result form the server, which would have `data` and `errors`, thus that code is not trying to get `errors` from inside `data`.  At least that is not what I observed when debugging that code (not sure if you saw the screen shot in the issue thread).

Torsten85:
The tests show that I've missed quite a lot of cases there so I have to invest more time in this. But before I start I need an "official statement" on how the result should look like. The docs are not very clear: "Any errors reported will come under an error prop along side the data returned from the cache or server."

twelve17:
@Torsten85 One important distinction is what the apollo *client* itself will return to the component, vs. what the apollo *server* returns to the client.  The bug, as I understand it at the moment, is that the client is not searching in the proper place for the server's response (which exists in the `errors` (plural) property).

Since now we have a divergence in property naming, I no longer know what the *client* should return to the component.  I believe this is what you are referring to here (as well as in a previous comment).  

In the case of server `errors`, should the client provide a `errors` (plural) component, which is also an array, similar to what the server returns?  Or should it return that same array in the `error` property, which is not always an array (in the case of network errors, for example, where there is no server response).  

Indeed, we need an official statement.

jbaxleyiii:
@Torsten85 thanks for the work on this one! The fix for this actually was merged in (but not released on the beta channel yet). Should be out this week!

twelve17:
@jbaxleyiii Can you let us know what release the fix is in? I'm on react-apollo 2.4 and am still seeing the behavior [in this screenshot|https://github.com/apollographql/react-apollo/issues/1389#issuecomment-363018389], and not seeing the error being set to the `error` property.

menkveldj:
June 21, still seeing the issue...

PR #1631, opened 2018-02-05
currentResult: use "errors" (plural) instead of "error"

### LINKED ISSUE #1389: errorPolicy not working

twelve17:
Fixes #1389.

**Note: this needs assistance before accepting.**

According to the [Apollo Docs](http://graphql.org/learn/serving-over-http/#response), the `response` object contains an `errors` (plural) attribute with an array of errors, whereas the client is looking for a single `error` attribute.

Apologies for not being able to correct the unit tests thus far, as I am not super well versed in typescript, and it seems that I've managed to run into a problem refactoring the tests anyway:

[The `MockedResponse` type in the `test-links` file](https://github.com/apollographql/react-apollo/blob/master/src/test-links.ts#L12) has an  `error` attribute, and all tests are configured to use it. However, I see tests that are also testing network errors.  In such cases, it doesn't make sense to be using a `MockedResponse` since, in theory, there was not a response, and it seems like a different code path for resolution on the client side.  

Alternatively, I wasn't sure if I should have made the `MockedResponse` have _both_  `error` and `errors` properties, so both network errors and graphql error use cases could be handled in the tests.

Assistance would be greatly appreciated.

### Checklist:

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -583,7 +583,12 @@ export default function graphql<
         } else {
           // fetch the current result (if any) from the store
           const currentResult = this.queryObservable.currentResult();
-          const { loading, error, networkStatus } = currentResult;
+
+          // GraphQL's server response returns a "errors" (plural) property,
+          // whereas the Apollo client reports them under a "error" (singular)
+          // property.  We rename it here.
+          const { loading, errors: error, networkStatus } = currentResult;
+
           assign(data, { loading, networkStatus });
 
           // Define the error property on the data object. If the user does


twelve17:
@Torsten85 created a PR right before I did, and that one looks better already.  Closing this one as a dupe of #1630 .

PR #1639, opened 2018-02-07
fix(Query): track previousData for showing while loading

### LINKED ISSUE #1635: [2.1.0-beta.2] Query always alternates returning undefined, and response for data
**Intended outcome:**
Query should call `children` with the previous result while loading

**Actual outcome:**
Query calls children with `{ data: undefined }`, then again with `{ data: <populatedData> }` once the response finishes.

Log response in `children` while modifying `variables`

```javascript
<Query query={query} variables={variables}>
  {({ data }) => { 
    // undefined, then populated even if there is content available in cache
    console.log(data) 
  }}
</Query>
```

**Version**

* apollo-client@2.2.0 
* react-apollo@2.1.0-beta.2

arahansen:
fixes #1635

this ports functionality that was [in graphql.ts](https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L621-L630) over to `Query.tsx`

### Checklist:

* [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes) **(N/A)**
* [ ] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion. **(N/A)**

## Note:
I wasn't able to get pre-commit hooks to run, I kept getting a "could not find directory ./node_modules/pre-commit"


arahansen:
Thanks for merging this! Any chance we could get a new cut of the beta branch? Would love to continue playing with the Query component.

derek-duncan:
@jbaxleyiii Can we get a new beta with this fix? I'm up for helping out with anything we need to get there too

clayne11:
This is incredibly undesirable behavior IMO. If the variables change we shouldn't be returning data from the previous variables as they don't match the query we're performing. It's possible to add a cache like this in userland code, however it's impossible to revert to the opposite behavior once this is in place.

PR #1641, opened 2018-02-07
Flow types for the Query component

excitement-engineer:
This PR adds flow typings for the Query component. 

This PR also removes the existing types for the `graphql` HOC, these type definitions were wrong.


jbaxleyiii:
@excitement-engineer per our convo in slack, lets drop flow and push for flow-typed / community support

PR #1652, opened 2018-02-09
Run tests for compiled bundles.

leoasis:
This PR adds an extra check when compiling the bundles that runs the tests against each bundle type. It currently runs tests for the browser umd bundle, the server umd bundle and the common js files. .mjs files are not added since Jest doesn't yet have support for them.

Currently the server umd bundle is causing some tests to fail, this seems to be an issue related to how rollup deals with the re-exports we do in both `browser.ts` and `index.ts`. I managed to fix some of those by removing one level of re-exports, by avoiding re-exporting browser.ts from index.ts and instead copy-paste those imports. But still that doesn't seem to work well with non default imports from the leaf modules. We'll need to find a way to fix this, but since this is still broken in master, we can maybe do this in a separate PR (unless the issue is spotted and we find a fix before merging this).

The tests against the umd bundles verify we're testing against the public API of the package, and will fail otherwise with this error:

image

I also removed some lodash methods that we were using that are now trivial to implement with new JS features (some are already there). This should allow us to ship a bit fewer bytes.

Should we add the compilation and testing against the bundles as a PR check? If so, this should be added when we find a solution to the failing tests for the server umd bundle.


jbaxleyiii:
Fixed via #1684

PR #1653, opened 2018-02-09
Add variables definition to OperationComponent Flow interface

igorgladkoborodov:
Current OperationComponent Flow interface does not include the query variable definition. It means that the code `options: (props) => { variables: { ... } }` in grapqhl HOC is not fully covered by Flow.

I have added the fourth parameter to the OperationComponent interface where you can define variables.

Here is the example:

```
const HERO_QUERY = gql`
  query GetCharacter($episode: String!) {
    hero(episode: $episode) {
      name
      id
      friends {
        name
        id
        appearsIn
      }
    }
  }
`;

type Hero = {
  name: string,
  id: string,
  appearsIn: string[],
  friends: Hero[],
};

type Response = {
  hero: Hero,
};

export type InputProps = {
  episode: string,
};

type Variables = {
  episode: string,
};

type Props = GraphqlData<Response, Variables> & {
  someProp: string,
};

const withCharacter: OperationComponent<
  Response,
  InputProps,
  Props,
  Variables,
> = graphql(HERO_QUERY, {
  options: ({ episode }) => {
    // $ExpectError [string] The operand of an arithmetic operation must be a number
    episode * 10;
    return {
      // $ExpectError [number] This type is incompatible with string
      variables: { episode: 10 },
    };
  },
  props: ({ data, ownProps }) => ({
    ...data,
    // $ExpectError [string] This type cannot be compared to number
    episode: ownProps.episode > 1,
    // $ExpectError property `isHero`. Property not found on object type
    isHero: data && data.hero && data.hero.isHero,
    // $ExpectError Property `someProp`. This type is incompatible with string
    someProp: 1,
  }),
});
```

These change are not fully backward compatible. Old code without variables definition will raise Flow warnings.

Combined with query and variables typed generated by apollo-codegen it will provide the full Flow coverage of everything inside graphql HOC.


chollier:
It looks like flow isn't exported to the npm package anymore — is this example supposed to still work?

PR #1661, opened 2018-02-10
Refactor to make the project compatible with the new context API.

excitement-engineer:
This pull request aims to update react-apollo with the new context API being introduced in react@16.3. 

I introduced `create-react-context` as a polyfill for people using older versions of react. We first check if the `createContext` is available on the `React` object and if not then we fallback to the polyfill.

API that has been converted:
- [x] `ApolloProvider`
- [x] `ApolloConsumer`
- [x] `Query`
- [ ] `graphql` HOC
- [ ] `QueryRecyclerProvider`
- [x] `Mutation`
- [x] `Subscription`


jaydenseric:
Are you sure SSR (`getDataFromTree`) is working when using the `Consumer` component from the new React context API? I'm [having issues](https://github.com/ctrlplusb/react-tree-walker/issues/22#issuecomment-369803759) trying to walk into it with `react-tree-walker`, which has fairly similar logic.

jbaxleyiii:
@excitement-engineer this is done (including SSR @jaydenseric ) on #1742

PR #1678, opened 2018-02-15
Skip updateQuery for Mutations when client is specified

MilosMosovsky:
This PR fixes the error behaviour when you try to specify `client` options on the mutation wrapped HoC. Skip the updateQuery and subscribeToQuery inside a function
when client option is passed to the mutation. It fixes initial query execution and failure that variables are not passed to the mutation query.

## Usecase:
Usage of different client for the HoC mutation component. 

```jsx
const CustomClient =  new ApolloClient({
  link,
  cache
})

export default graphql(gql`
  mutation addUser($userId: ID!) {
    addUser(userId: $userId) {
  }
`, {
  options: () => ({
    client: CustomClient
  })
})(Component)
```

## Actual Behaviour
`browser.js?7a6c:40 Uncaught Error: The operation 'addUser' wrapping 'Component' is expecting a variable: 'userId' but it was not found in the props passed to 'Apollo(Component)'
`

## Expected Behaviour
`It should mount a component without any error as mutation is not executed yet`
### Checklist:

### DIFF
--- a/src/graphql.tsx
+++ b/src/graphql.tsx
@@ -189,6 +189,10 @@ export default function graphql<
           this.queryObservable = null;
           this.previousData = {};
 
+          if (this.type === DocumentType.Mutation) {
+            return;
+          }
+
           this.updateQuery(nextProps);
           if (!this.shouldSkip(nextProps)) {
             this.subscribeToQuery();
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@MilosMosovsky could you add a changelog for this and maybe an inline comment detailing the change?

jbaxleyiii:
This shouldn't be needed anymore now that mutations use their own component

Nedomas:
@MilosMosovsky ran into the same issue, thanks. @jbaxleyiii if I understand correctly, we cannot specify `client` in the new betas at all?

PR #1687, opened 2018-02-17
Pass reject from mutation function

excitement-engineer:
The promise returned by the mutation function passed in the render prop of the <Mutation /> component now rejects in case an error occurs.

Note, that some tests are failing due to an internal error in apollo-client. Do you know what is wrong @jbaxleyiii?


jbaxleyiii:
Done via #1715 🎉

PR #1692, opened 2018-02-19
Remove unused dependency graphql-tag

excitement-engineer:
This PR removes graphql-tag from the dependencies. I think it was accidentally introduced in #1650.


jbaxleyiii:
@excitement-engineer I think we will need to move it back to dev. I'll do that in my cleanup PRs today so I'm going to close this one out. Good catch!

Issue #1616, opened 2018-02-01
[2.1] Prop to skip execution in <Query/> component


scarmuega:
*Use case*:
* you have an optional query, which should only be executed under certain scenarios
* you also have children inside the <Query/> component which should be rendered statically, for every scenario.

*Workaround*: I currently choose to avoid rendering the <Query/> component altogether, but this changes the tree structure, which makes children unmount and re-mount again. Also thought of passing a null gql to the query prop, but that throws an exception.

*Proposal*: add a `skip` prop to the <Query/> component. When flagged, it should bypass the graphql call, but still execute the children render prop, to maintain the same virtual dom tree structure.

abenhamdine:
https://www.apollographql.com/docs/react/basics/queries.html#graphql-skip ?

scarmuega:
@abenhamdine I'm referring to the new `<Query/>` component that uses render props, not the `graphql` container HOC (maybe I should make that clearer in the issue description).

AFAIK, the link you provided details the "skip" option for the later.

abenhamdine:
>  I'm referring to the new <Query/> component that uses render props, not the graphql container HOC (maybe I should make that clearer in the issue description).

Oh yes, sorry, I realized after having posted that you were referring to the Query component.
Nethertheless, perhaps it would be useful to explain if your proposal is the equivalent (or not equivalent, and so why) of the HOC `skip` property, to be clearer for everyone.

arahansen:
I would also really appreciate a `skip` property that is equivalent to the HOC's skip property. ie, we should still render the underlying component (children here), but not execute the query.

arahansen:
I'm currently working around this by setting `fetchPolicy={skip() ? 'cache-only' : undefined}`

f2net:
I think this property would be useful, too.
In real world web apps you often need to skip queries for not-logged users to avoid useless not-authenticated errors.

Thank you,
Matteo

arahansen:
This feature should be available on 2.1.0-beta.3. Check out this PR for more info: https://github.com/apollographql/react-apollo/pull/1715

PR #1694, opened 2018-02-20
[Query]: Implement skip prop

arahansen:
### Checklist:

* [ ] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion. **(N/A)**

Prop is currently being discussed on https://github.com/apollographql/react-apollo/issues/1616

## Open Questions:

Currently, `children` render prop will be called with `undefined` if skip resolves to true. The result is that all usage of `Query` had to be updated to account for the type signature update. It didn't feel great to have to do this which made me think this isn't the best API approach.

Perhaps, alternatively we should set `data` to undefined and force `loading` to false? Would that be reliable/predictable for any potential edge cases?


excitement-engineer:
Hey @arahansen! Thanks for contributing:) 

When creating the `Query` component initially @leoasis made a good point about why the `skip` was not needed. See [comment](https://github.com/apollographql/react-apollo/pull/1398#issuecomment-353340397) for the reasoning behind this decision

arahansen:
Ah yeah. That's an interesting point. 

However, I still want to render my children, I just don't want the query to execute. This is behavior supported by the hoc.

With the hoc, If skip resolves to true, the query will not be executed, but the wrapped component will still be rendered, but no data.

test/client/Query.test.tsx
```diff
@@ -163,12 +171,12 @@ describe('Query component', () => {
       const Component = () => (
         <Query query={allPeopleQuery}>
           {result => {
-            if (result.loading) {
+            if (!result || result.loading) {
```
jbaxleyiii: I think this is the behavior that most concerns me. I'd like to see this still be an object with thinks like `client`, the observable query methods, and loading set to false (since its skipped its not loading). The only thing that would be missing on a skip would be data / error

test/client/Query.test.tsx
```diff
@@ -1174,7 +1200,161 @@ describe('Query component', () => {
...
+        skip: true,
+      };
+
+      componentDidMount() {
+        this.setState({ skip: false });
+      }
+
+      render() {
+        return (
+          <Query query={query} skip={() => this.state.skip}>
```
jbaxleyiii: I don't think skip should be a function, since you have props in scope, it should just be a boolean

src/Query.tsx
```diff
@@ -201,9 +215,14 @@ class Query<
   render() {
     const { children } = this.props;
     const queryResult = this.getQueryResult();
-    return children(queryResult);
+    return children(this.shouldSkip() ? undefined : queryResult);
   }
 
+  private shouldSkip = (props = this.props) => {
```
jbaxleyiii: instead of a function, this we should take `skip` as a boolean prop which makes it easy to check against previous props and lessen the internal function calls.

src/Query.tsx
```diff
@@ -201,9 +215,14 @@ class Query<
   render() {
     const { children } = this.props;
     const queryResult = this.getQueryResult();
-    return children(queryResult);
+    return children(this.shouldSkip() ? undefined : queryResult);
```
jbaxleyiii: This is noted below, but we shouldn't return `undefined` here, but rather return everything we can (including `client`) so that delayed actions can be done. It also lowers the risk of error for people checking loading / data / states on an undefined vs {}

test/client/Query.test.tsx
arahansen: yeah, this is what I was referring to in my original comment. Should we be hard-coding `data` to `{}` and `loading` to `false`? or remove the `data` field entirely?

test/client/Query.test.tsx
arahansen: This makes sense to me! Will update

test/client/Query.test.tsx
```diff
@@ -50,7 +51,7 @@ describe('Query component', () => {
     }
   });
 
-  it('calls the children prop', done => {
+  fit('calls the children prop', done => {
```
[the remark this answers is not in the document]
arahansen: 🙈

arahansen:
hey @jbaxleyiii thanks for the feedback! I think `skip` being a boolean feels much nicer. Let me know if there's anything else I can take a look at 🙂

leoasis:
Yeah, in the issue linked by OP there is a good point in favor of `skip`, which I haven't thought about at that time, which is that you may want to render some content that sometimes doesn't need to fetch data, but in all cases it has some part of the tree that is the same. So to avoid React unmounting the tree and remounting, adding a `skip` prop that would keep the same component tree but not run the query would be useful in those scenarios. Most of the time you'd just conditionally render the `Query` component, but for cases where you need this optimization, this prop would come in handy.

jbaxleyiii:
@arahansen thank you for this PR! I ended up rewritting a huge portion of the library this weekend and pulled in a lot of this PR into it so its on master now! Thank you again! Done via #1715 🎉

arahansen:
Sounds good, really excited to try out the new beta!!

muhaimincs:
Hi I'm using **2.1.0-rc.3**. How do I implement skip because I still get ```loading``` is true

PR #1715, opened 2018-02-23
Moving

jbaxleyiii:
🌮 this is a big one, so take seat and grab a taco.

### Death of the Recycler
At long last, we have removed the query recycler, the most common point of loading pain for people. It is replaced with a small shim to backwards support refetchQueries which use a string (which will be removed in future Apollo Client versions)

### So long dear friend
The much beloved (by me at least) original `graphql` hoc is now gone, It is replaced by a small switch statement that chooses the correct named hoc (query, mutation, subscriptions) which wraps the named components. The *GOOD* news here, is if you tree shake, this thing oughta shake well. 

### New kids in town
```
> A wild group of HOCs have appeared!! Do you engage or run away?
> Engage!
> `query`, `mutation`, and `subscription` walk up, they all look like `graphql`, but let you know what you are doing. Throw pokeballs?
> YOU KNOW IT! GOTTA WRAP EM ALL
```

This PR should maintain full backwards compat. There are three tests that need more assertions around query recycling, but this is good enough to get out into a beta for people to try out.

@rosskevin @excitement-engineer @leoasis I tried to keep the types tight, but please let me know if I screwed something up?

🎉


src/ApolloProvider.tsx
```diff
@@ -29,17 +32,23 @@ export default class ApolloProvider<TCache> extends Component<ApolloProviderProp
...
       client: this.props.client,
+      operations: (this.props.client as any).__operations_cache__,
     };
   }
 
   render() {
-    return (
-      <QueryRecyclerProvider>{React.Children.only(this.props.children)}</QueryRecyclerProvider>
-    );
+    return React.Children.only(this.props.children);
```
rosskevin: I'm not sure we care about children being one and only one here.  Any reason (with react 16) that we shouldn't just `return this.props.children`?  If that complains, then just `return <React.Fragment>{this.props.children}</React.Fragment`

jbaxleyiii: ah yeah that sounds great to me

src/Mutation.tsx
```diff
@@ -157,7 +161,15 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
   };
 
   private mutate = (options: MutationOptions<TVariables>) => {
-    const { mutation, variables, optimisticResponse, refetchQueries, update } = this.props;
+    const { mutation, variables, optimisticResponse, update } = this.props;
+    let refetchQueries = options.refetchQueries || this.props.refetchQueries;
+    if (refetchQueries && refetchQueries.length && Array.isArray(refetchQueries)) {
+      refetchQueries = (refetchQueries as any).map((x: string | PureQueryOptions) => {
+        if (typeof x === 'string' && this.context.operations) return this.context.operations.get(x);
```
excitement-engineer: Will this be removed in the 3.0 as well? Perhaps we can add a not of that here in the code.

src/Query.tsx
```diff
@@ -37,8 +36,9 @@ export interface FetchMoreQueryOptions<TVariables, K extends keyof TVariables> {
 // Improved ObservableQuery field types, need to port them back to Apollo Client
 export type ObservableQueryFields<TData, TVariables> = Pick<
   ObservableQuery<TData>,
-  'startPolling' | 'stopPolling'
+  'startPolling' | 'stopPolling' | 'subscribeToMore'
```
excitement-engineer: SubscribeToMore is used for subscriptions right? Should it be exposed in the Query component or is it being used for backwards compatibility?

src/Subscriptions.tsx
```diff
@@ -19,6 +19,7 @@ export interface SubscriptionResult<TData = any> {
 export interface SubscriptionProps<TData = any, TVariables = OperationVariables> {
   query: DocumentNode;
   variables?: TVariables;
+  shouldResubscribe?: boolean;
```
excitement-engineer: We need to add a test for this in the Subscription component test suite in a follow-up

src/Query.tsx
excitement-engineer: If we do decide to keep it then I think we should include a test in the `Query.test.tsx` file in a follow-up

src/Mutation.tsx
```diff
@@ -41,12 +42,13 @@ export declare type FetchResult<C = Record<string, any>, E = Record<string, any>
...
   variables?: TVariables;
   optimisticResponse?: Object;
-  refetchQueries?: string[] | PureQueryOptions[];
+  refetchQueries?: string[] | PureQueryOptions[] | RefetchQueriesProviderFn;
   update?: MutationUpdaterFn<TData>;
 };
 
 export interface MutationProps<TData = any, TVariables = OperationVariables> {
   mutation: DocumentNode;
+  ignoreResults?: boolean;
```
excitement-engineer: Need to add a test in the `Mutation.test.tsx` file as a follow-up

src/Query.tsx
```diff
@@ -90,22 +88,22 @@ export interface QueryProps<TData = any, TVariables = OperationVariables> {
   query: DocumentNode;
   variables?: TVariables;
   ssr?: boolean;
-}
-
-export interface QueryState<TData = any> {
-  result: ApolloCurrentResult<TData>;
+  displayName?: string;
+  skip?: boolean;
+  client?: ApolloClient<Object>;
```
excitement-engineer: Need to add tests for these props in a follow-up PR in `Query.test.tsx`.

src/subscription-hoc.tsx
```diff
@@ -0,0 +1,116 @@
...
+          resubscribe: shouldResubscribe(this.props, nextProps),
+        });
+      }
+
+      render() {
+        let props = this.props;
+        const shouldSkip = mapPropsToSkip(props);
+        const opts = shouldSkip ? Object.create(null) : mapPropsToOptions(props);
+
+        if (!shouldSkip && !Boolean(opts.variables || !operation.variables.length)) {
```
excitement-engineer: This if statement is a little difficult to read with all the negations, it can be simplified to: 

```js
if (!shouldSkip && !opts.variables && operation.variables.length > 0) {}
```

test/client/Query.test.tsx
```diff
@@ -67,10 +67,10 @@ describe('Query component', () => {
             const { client: clientResult, ...rest } = result;
 
             if (result.loading) {
-              expect(rest).toMatchSnapshot('result in render prop while loading');
+              expect(stripSymbols(rest)).toMatchSnapshot('result in render prop while loading');
```
excitement-engineer: `stripSymbols` is not only removing all the symbols but also the checks for all the render props (see the updated snapshots in the diff). Should we remove the stripSymbols from this test to make sure that the snapshot is complete?

src/Query.tsx
```diff
@@ -56,26 +56,24 @@ function observableQueryFields<TData, TVariables>(
   observable: ObservableQuery<TData>,
 ): ObservableQueryFields<TData, TVariables> {
   const fields = {
+    variables: observable.variables,
     refetch: observable.refetch.bind(observable),
     fetchMore: observable.fetchMore.bind(observable),
     updateQuery: observable.updateQuery.bind(observable),
     startPolling: observable.startPolling.bind(observable),
     stopPolling: observable.stopPolling.bind(observable),
+    subscribeToMore: observable.subscribeToMore.bind(observable),
```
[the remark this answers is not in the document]
jbaxleyiii: @excitement-engineer `subscribeToMore` is how you can take a regular query, and attached a subscription to it to live update it. So its part of the `Query` feature set

src/Mutation.tsx
```diff
@@ -56,6 +58,7 @@ export interface MutationProps<TData = any, TVariables = OperationVariables> {
     result?: MutationResult<TData>,
   ) => React.ReactNode;
   onCompleted?: (data: TData) => void;
+  displayName?: string;
```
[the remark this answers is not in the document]
jbaxleyiii: Yep, that was a carry over from the Query additions

src/query-hoc.tsx
```diff
@@ -0,0 +1,141 @@
...
+                }, 10);
+                Object.defineProperty(r, 'error', {
+                  configurable: true,
+                  enumerable: true,
+                  get: () => {
+                    clearTimeout(logErrorTimeoutId);
+                    return error;
+                  },
+                });
+              }
```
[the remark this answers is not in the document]
jbaxleyiii: @excitement-engineer yep! I'm going to make it configurable too and off since its one of the major causes of issues on this repo

src/Mutation.tsx
jbaxleyiii: yep for sure!

src/Query.tsx
jbaxleyiii: @excitement-engineer for sure! It isn't really tested in the current test suite either, just a smoke test to make sure its found and callable

src/Mutation.tsx
jbaxleyiii: I added a simple test of this (could be expanded to more cases) on the next commit

src/Query.tsx
jbaxleyiii: @excitement-engineer this kind of stuff does warrant an overall discussion about how we want the test suite to be adjusted / grow with both HOCs and components

PR #1725, opened 2018-02-27
Silence errors by default

jbaxleyiii:
@timbotnik just for you ;)

### DIFF
--- a/src/query-hoc.tsx
+++ b/src/query-hoc.tsx
@@ -58,6 +58,7 @@ export function query<
 >(
   document: DocumentNode,
   operationOptions: OperationOption<TProps, TData, TGraphQLVariables, TChildProps> = {},
+  logUnhandled: boolean = false,
 ) {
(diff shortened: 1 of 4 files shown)


src/query-hoc.tsx
```diff
@@ -58,6 +58,7 @@ export function query<
 >(
   document: DocumentNode,
   operationOptions: OperationOption<TProps, TData, TGraphQLVariables, TChildProps> = {},
+  logUnhandled: boolean = false,
```
timbotnik: should this be like `processingOptions.logUnhandled`?

timbotnik: Just suggesting to use named parameters in a new options key instead of positional.

jbaxleyiii: @timbotnik ah! Well the plan is for this to be removed in the very near term with a 3.0 of react-apollo (no API change, just removal of deprecated options). So I think having it an internal positional argument is good because it prohibits doing more things in this position

PR #1742, opened 2018-03-01
[WIP] 3.0 alpha

jbaxleyiii:
This isn't really intended to be merged anytime soon as the React team is still working out this API. However, this PR implements strict mode requirements, new context for client and server rendering, and allows async mode for the query component as a first pass!

```bash
$ npm i --save react-apollo@canary
```

```js
<Query query={QUERY_DOC} asyncMode>
 {({ data }) => // all you need is data or error because react will delay this!}
</Query>
```

### TODO
The React ecosystem will need to catch up for things like enzyme support, and we will have to drop preact in this release for the time being.

- finalize all currently skipped tests
- cleanup quick type overrides
- usage patterns and testing
- drop anything we want to deprecate
  - ~prop-types~
  - ~compose~
  - ?

If you try this out *please* ping me in slack or comment here! I'd love to hear about how its going!


src/Context.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const ApolloProvider: React.StatelessComponent<any> = ({
+  client,
+  children,
+}: ApolloProviderProps<any>) => <Provider value={client}>{children}</Provider>;
+
+export interface ApolloConsumerProps {
+  children: (client: ApolloClient<any>) => React.ReactElement<any> | null;
+}
+
+export const ApolloConsumer: React.StatelessComponent<ApolloConsumerProps> = Consumer;
```
excitement-engineer: I think that there is still a use case for the old ApolloConsumer. It verified that a client had been set by the provider and if not it threw an error for the user. Should we keep the old `ApolloConsumer` and make it a wrapper around the consumer defined here?

src/Query.tsx
```diff
@@ -370,8 +365,16 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
 
-    data.client = this.client;
+    data.client = this.props.client;
 
     return data;
   };
 }
+
+export default class ApolloQuery<TData = any, TVariables = any> extends React.Component<
+  QueryProps<TData, any>
```
excitement-engineer: Shouldn't this be `QueryProps<TData, TVariables>`?

src/Query.tsx
```diff
@@ -370,8 +365,16 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
     }
 
-    data.client = this.client;
+    data.client = this.props.client;
 
     return data;
   };
 }
+
+export default class ApolloQuery<TData = any, TVariables = any> extends React.Component<
```
excitement-engineer: We can type `TVariables` as `OperationVariables` instead of any to make it a bit more strict by default.

src/Query.tsx
```diff
@@ -351,11 +340,17 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
 
     Object.assign(data, { loading, networkStatus, error });
 
+    if (evictData) this.previousData = {};
     if (loading) {
+      // only throw if we are still in a loading state
+      if (this.props.asyncMode) {
+        // async mode wants to throw a promise that waits for the data to be loaded
+        throw this.state.queryObservable!.result();
```
excitement-engineer: Wow cool!!!

src/Query.tsx
```diff
@@ -147,86 +184,92 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
+    nextProps: InnerQueryProps<any, any>,
+    prevState: QueryState<any>,
+  ) {
+    // if we aren't working from a live query, we can just ignore props changes
+    if (!prevState.queryObservable) return null;
 
-  constructor(props: QueryProps<TData, TVariables>, context: QueryContext) {
-    super(props, context);
+    // if there are no changes to the props, don't do anything state wise
+    if (shallowEqual(nextProps, prevState.props)) return null;
```
jamesreggio: I'm surprised there wasn't more of a firestorm over the lack of a `prevProps` argument to the `getDerivedStateFromProps` function. (Maybe I missed the firestorm, or maybe people haven't realized yet how ugly it's going to be to have to store props in state strictly for the sake of comparison.)

Either way, stashing the _entire_ prior `props` in `state` already has a code smell. If possible, it'd be good to evaluate the prop-change dependencies of each state key, as opposed to these wholesale comparison short-circuits.

jamesreggio: More importantly, it looks like this class never updates `state.props` to the new value of `props`. I think every `return` within the method after this line needs to include `{ props: nextProps }`.

src/Query.tsx
```diff
@@ -147,86 +184,92 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
 
-    const { client } = nextProps;
-    if (
-      shallowEqual(this.props, nextProps) &&
-      (this.client === client || this.client === nextContext.client)
-    ) {
-      return;
-    }
+    // if there are no changes to the props, don't do anything state wise
+    if (shallowEqual(this.props, prevProps)) return null;
```
jamesreggio: Again, this short-circuit is unnecessary, given that everything below is guarded with more specific conditionals.

Plus, in the world of render-props (where the prevailing trend is to use inline arrow functions), the `children` prop is _never_ going to equal its prior value. Might as well save some cycles and cognitive overhead.

src/Query.tsx
```diff
@@ -147,86 +184,92 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
-    if (this.props.query !== nextProps.query) {
       this.removeQuerySubscription();
     }
 
-    this.updateQuery(nextProps);
-    if (nextProps.skip) return;
-    this.startQuerySubscription();
+    if (this.props.skip) return;
+    // start a new subscription if we don't have one already
+    this.startQuerySubscription(this.state.queryObservable);
```
jamesreggio: Nit: in the interest of triggering each side-effect off of a specific, easy-to-understand change in props, I'd rewrite this as:

```js
if (!this.props.skip) {
  this.startQuerySubscription(this.state.queryObservable);
}
```

That way, you don't have to think about the accumulation of prior conditions from which we may have hit an early `return`.

src/Query.tsx
```diff
@@ -320,28 +305,32 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
-    this.startQuerySubscription();
-    Object.assign(this.queryObservable!, { lastError, lastResult });
+    queryObservable!.resetLastResults();
+    this.startQuerySubscription(queryObservable);
+    Object.assign(queryObservable!, { lastError, lastResult });
   }
 
   private updateCurrentData = () => {
-    // force a rerender that goes through shouldComponentUpdate
     if (this.hasMounted) this.forceUpdate();
```
jamesreggio: I'm not particularly excited about throwing promises, but I _am_ excited about the day where you can change this line to do something like this:

```js
if (this.hasMounted) {
  if (this.props.priority === 'low') {
    ReactDOM.unstable_deferredUpdate(() => this.forceUpdate());
  } else {
    this.forceUpdate();
  }
}
```

Naturally, I also hope that the React team decouples this scheduling primitive from `ReactDOM` and provides further granularity (e.g., explicit scheduling timeouts), but hey, this is _the_ value prop for me.

FredyC:
I am curious, do you have some example for typing queries/mutations written like this? I mean I am using https://github.com/dotansimha/graphql-code-generator to generate TypeScript annotations that look like this.

```ts
export namespace GLogoutWidget {
  export type Variables = {
    id: string
  }

  export type Query = {
    user: User
  }

  export type User = {
    id: string
    firstName?: string | null
    lastName?: string | null
  }
}
```

Currently I have my own wrapper around `graphql` HoC that allows me to do this...
```
export const LogoutWidgetQuery = buildQuery<
  GLogoutWidget.Query,
  GLogoutWidget.Variables
>(gql`
  query GLogoutWidget($id: ID!) {
    user(id: $id) {
      id
      firstName
      lastName
    }
  }
`)
```

It works really well, but it seems that with render prop approach this is not possible or am I missing something here?

neoziro:
Is it possible to extract the visitor from getDataFromTree in a separated package? If not I will do it because I need it for Loadable Components. Using the same package between libraries could improve perf if we use a common cache ;).

mike-marcacci:
Just wanted to touch base on the goals/needs of this PR. I have a few other changes I've started, which need to be implemented quite differently depending on which lifecycle methods we're using. I would prefer not making more work than necessary, so:

- Can we break out just the changes to new lifecycle methods and context API?
- Would you be interested in a PR for that, which could be merged more quickly?

hwillson:
A lot of the React internals have changed since this PR was created, so we'll close this off for now. We're still waiting for the dust to settle a bit more on the changes / proposed changes in the [React 16.x Roadmap](https://reactjs.org/blog/2018/11/27/react-16-roadmap.html), but we'll be working on whatever React changes / future proofing we can, as things are being finalized, in separate PR's.

PR #1769, opened 2018-03-08
Bundles

jbaxleyiii:
Sheesh, bundling is a mess in the js world today. This PR gives what I think is the best chance for future success? 

~This doesn't ship a flat bundle anymore, but looks like that may not be the best way forward. Instead we should work to get all Apollo projects to replicate the above so full tree shaking is possible. It will need to start with the Apollo Client repo for base utils, then include Apollo Link repo before this mjs will be particularly useful. This is part of the overall goals outlined here https://github.com/apollographql/apollo-link/issues/537~

EDIT: This didn't work. It breaks usage with CRA which is impacting graphql-js too https://github.com/facebook/create-react-app/pull/4085

This PR reverts any mjs inclusion for now


fabien0102:
mmmm, I've just try to update to `2.1.0-rc.1`, the `.mjs` are still there and the build is still broken on CRA…

tavelli:
just tested with 2.1.0-rc.3, working!

gaearon:
If your project is built with Create React App and not ejected, `react-scripts@1.1.2` should fix this.
Here is how to upgrade:

https://github.com/facebook/create-react-app/releases/tag/v1.1.2

Otherwise you can apply a [similar fix](https://github.com/facebook/create-react-app/pull/4085/files) yourself in your project config. Cheers!

PR #1774, opened 2018-03-10
Fixed inconsistency in naming of the prop for the subscription

excitement-engineer:
There is an inconsistency in the names of the props for the documents:

```
<Query query={} />
<Mutation mutation={} />
<Subscription query={} />
```

The name of the prop for the `Subscription` component is inconsistent with the others. Therefore, I propose that we call the prop `subscription` instead of `query`

```
<Subscription subscription={} />
```

### DIFF
--- a/src/Subscriptions.tsx
+++ b/src/Subscriptions.tsx
@@ -42,7 +42,7 @@ class Subscription<TData = any, TVariables = any> extends React.Component<
   };
 
   static propTypes = {
-    query: PropTypes.object.isRequired,
+    subscription: PropTypes.object.isRequired,
     variables: PropTypes.object,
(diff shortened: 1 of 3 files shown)


excitement-engineer:
Curious to hear your thoughts @jbaxleyiii

fbartho:
Counter-proposal: move the `mutation` prop to be `query={`. They're all GraphQL queries.

excitement-engineer:
This is also an option:) I'm happy as long as it is consistent!

jbaxleyiii:
I think we can either go `query` => `operation` or `query` => `subscription`. I think subscription will lead to the least confusion when looking back at old code, especially if you extend the component class and give it another name (like in TS)

Issue #1773, opened 2018-03-09
Why is the Mutation component's result `undefined` before the mutation is called?


rylanc:
This is more of a clarification than a bug report or feature request.

When using the `Mutation` component and the mutation hasn't been called, the component passes a result that is `undefined` (see https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L137).

This is clearly intentional, but it prevents the user from destructuring the result argument:
```js
<Mutation>
  {(mutate, { loading }) => (loading ? <Loading /> : <Button onClick={mutate />)}
</Mutation>
```

It's easy enough to work around:
```js
<Mutation>
  {(mutate, result) => (result && result.loading ? <Loading /> : <Button onClick={mutate />)}
</Mutation>
```

It seems to me that it would be less confusing if the result behavior more closely matched the `Query` component (the result is always an object). If the intention is to convey that the mutation hasn't been called yet, could a `called`/`notCalled` entry be added to the result?

**Version**

* apollo-client@any
* react-apollo@2.1.0-rc.3

excitement-engineer:
@rylanc, I was also running into this issue myself. The idea of the nullable prop is that it signifies that the mutation has not been called yet. However, this does complicate destructuring.

I will create a PR to investigate the possibilities to resolve this.

Thanks for creating this PR and providing your feedback:)

excitement-engineer:
I created a PR. Let me know if this API clears things up for you?

rylanc:
The PR is just what I was imagining. Thanks!

Gregoirevda:
I don't get the issue about destructuring above. The initial result always is: 
```
called: false
loading: false
data: undefined
error: undefined
```
`data` is `undefined`, but result is not. Which doesn't prevent you to 

```
<Mutation>
  {(mutate, { loading }) => (loading ? <Loading /> : <Button onClick={mutate />)}
</Mutation>
```

Is there a confusion between result and data?

PR #1775, opened 2018-03-10
Added "called" prop to the Mutation component

excitement-engineer:
Added a "called" field in the `MutationResult` prop returned in the render prop of the Mutation component. This field is true or false depending on whether the mutation has been called previously.

Adding this field ensures that the `MutationResult` object in the render prop is always defined which makes destructuring easier. In addition,  this will prevent confusion about the meaning of an undefined `MutationResult` in the render prop as mentioned by #1773.

Would love to hear some feedback regarding these changes. Does this improve the experience of using the `Mutation` component.

### DIFF
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -12,7 +12,8 @@ import { parser, DocumentType } from './parser';
 export interface MutationResult<TData = Record<string, any>> {
   data?: TData;
   error?: ApolloError;
-  loading?: boolean;
+  loading: boolean;
+  called: boolean;
 }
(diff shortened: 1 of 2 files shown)


fbartho:
This is a breaking API change, but it really makes sense to me. I support this!

janhartmann:
This is so much needed - I was upgrading our system to the beta version, but I ran into these `undefined` issues but could not understand why, after viewing the source code of the `Mutation` component. 

It was because this change is not yet released. I can not wait! Nice work, guys!

Issue #1716, opened 2018-02-24
Mutation component update when unmounted


dyst5422:
The new Mutation component tries to update after its been unmounted. An example of this is when I have a mutation which results in changing my view in such a way that the mutation is no longer mounted. A pretty common scenario.

rylanc:
I'm experiencing this bug as well. It manifests itself in a console warning:
![image](https://user-images.githubusercontent.com/4443768/37228623-1c9f5d20-2396-11e8-80a3-107f5e5ee3a3.png)

I've created a demo in CodeSandbox:
https://codesandbox.io/s/nn21w8k0m4
If you open the console and remove a post, you can see the warning.

excitement-engineer:
I created a PR that fixes this! Thanks so much for filing this issue! Let me know if this fixes your issue

emmahoggan:
I'm still seeing the 'Can't perform a React state update on an unmounted component' console error when setState is called during onMutationStart, but my component has already unmounted.  

Is there a reason we don't also check for `this.hasMounted` before calling `this.setState(...)` in `onMutationStart`?

thawsitt:
Still seeing this error on Apollo 3.0 beta. 

```
react_devtools_backend.js:6 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

My use case is a selection modal. Users can select options in the modal and click on "Confirm" button. When confirm button is clicked, I make a mutation request and close the modal.

```js
const handleConfirm = () => {
    setSelection(checkedItems); // mutation
    closeModal(); // unmounted
};
```

klarstrup:
@thawsitt I know that it should work either way, but do you also get the warning if you `await` the mutation?

PR #1792, opened 2018-03-15
Fixed issue 1716 where Mutation updated state after unmounting.

excitement-engineer:
Fixes issue #1716 where the `Mutation` component mutated state even after unmounting.

Thanks @dyst5422 for the issue and @rylanc for the reproduction!


excitement-engineer:
hmm something is wrong with the CI. Locally all the tests pass but not on CI. Anyone has an idea?

PR #1801, opened 2018-03-17
officially drop .mjs support [again]

jbaxleyiii:
Somehow I lost the previous commit that actually did this with the previous bundles PR?! So its back again!


src/index.js.flow
```diff
@@ -1,199 +0,0 @@
-// @flow
```
thymikee: @jbaxleyiii, what's the reason behind removing Flow defs?

masad-frost: What happened here :eyes:

Issue #1777, opened 2018-03-11
Apollo 2, useless re-render (twice render). Simple example, simple reproduction on codesandbox.


johnunclesam:
# Reproduction: https://codesandbox.io/s/72kllvom21

**Intended outcome:**

Simple example of react-apollo. I have an Home component and a Players component. The Players has a graphql query to retrieve data.

**Actual outcome:**

The first time I have the good operation: it loads and I have two rendering because of `data > loading: true` value. The second time I go to that page it render two times but the content is the same and the `loading: false` both times.

I don't think this is the right way. **Where am I wrong?** https://codesandbox.io/s/72kllvom21

**How to reproduce the issue:**

https://codesandbox.io/s/72kllvom21

![image](https://user-images.githubusercontent.com/21275217/37255365-d3f2bb00-254b-11e8-8323-d0aecef44142.png)

**Version**

* apollo-client@2.2.6
* react-apollo@2.0.4

ntomsic:
This looks like it might be a bug. The 1st time you hit the Players route it should rerender twice. If you click the Players link again, it only rerenders once (expected). However, if you click the Home route and then back to the Players route, it rerenders twice again with no changes in props.

johnunclesam:
Maybe the same issue? https://github.com/apollographql/apollo-link-state/issues/217

akiran:
I am trying to debug this issue and found that forceUpdate [here](https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L337) is responsible for second re-render

johnunclesam:
@akiran why that force? Do you understand why?

jbaxleyiii:
@johnunclesam thanks for the super clear reproduction! Fixed in the newest release and verified here https://codesandbox.io/s/o743y3yx25

johnunclesam:
@jbaxleyiii thanks to you.

But using your last codesandbox I added `why-did-you-update` and there is still some problem:

**https://codesandbox.io/s/kmork0rxkv**

![image](https://user-images.githubusercontent.com/21275217/38036761-bcb8380c-32a7-11e8-9bb2-fda7683077ae.png)

![image](https://user-images.githubusercontent.com/21275217/38036783-c93a49ee-32a7-11e8-9fd6-099949bbf438.png)

putuyoga:
bump! i've face kind similar issue @johnunclesam at first i guess it was my component. So i make it pure, still no luck.

[...]

PR #1809, opened 2018-03-19
Fix to avoid unnecessary re-render

akiran:
This PR is an attempt to fix [unnecessary rerender issue](https://github.com/apollographql/react-apollo/issues/1777)
When data is queried from local-state, data is available in the first render itself and loading is not set to true. So second re-render is not necessary.
I added a check to avoid forceUpdate if loading is false initially.
Let me know if this approach is correct.

4 tests are failing with this PR. I am debugging them.


jbaxleyiii:
@akiran thanks for taking a stab at this! I built off of it with #1849

PR #1819, opened 2018-03-21
Typing improvements for withApollo

### LINKED ISSUE #1816: Can WithApolloClient type be made a top-level export (TypeScript)?
Currently, the `WithApolloClient` type is not exported directly by `react-apollo`; the only way to import it is to do this:

    import { WithApolloClient } from 'react-apollo/withApollo'

Was this just an oversight or is there some reason it's not exported at the top level? It would be great to just be able to do this:

    import { withApollo, WithApolloClient } from 'react-apollo'

**Version**

* apollo-client@2.2.7
* react-apollo@2.1.0-rc.5

mbrowne:
This exports the `WithApolloClient` type as a top-level export. Fixes #1816.

It also adds a default value for `withApollo`'s `TResult` parameter - it now defaults to `any`, for ease of use (e.g. if you're only using a mutation and don't care about the response [or the mutation always returns null] then there's no benefit in having to manually specify `any`).

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -19,6 +19,7 @@ export * from './Subscriptions';
 export { graphql } from './graphql';
 
 export { default as withApollo } from './withApollo';
+export * from './withApollo'
 
 export * from './types';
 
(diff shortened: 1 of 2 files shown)


mbrowne:
Thanks!

PR #1830, opened 2018-03-23
Updated Types for Mutation Component

brettjurgens:
### Checklist:

/label another-label-name

### DIFF
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -47,6 +47,10 @@ export declare type MutationOptions<TData = any, TVariables = OperationVariables
   update?: MutationUpdaterFn<TData>;
 };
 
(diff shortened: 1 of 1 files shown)


src/Mutation.tsx
```diff
@@ -47,6 +47,10 @@ export declare type MutationOptions<TData = any, TVariables = OperationVariables
   update?: MutationUpdaterFn<TData>;
 };
 
+export declare type MutationFn<TData = any, TVariables = OperationVariables> = (
+  options?: MutationOptions<TData, TVariables>,
+) => Promise<void | FetchResult<TData>>;
```
brettjurgens: declaring this makes it easier to use as a prop in our components. Also added `FetchResult<TData>` which was missing prior

jbaxleyiii:
@brettjurgens yay! Thank you!

PR #1844, opened 2018-03-25
fix: allow access results in mutation-hoc

doniyor2109:
### Checklist:

### DIFF
--- a/src/mutation-hoc.tsx
+++ b/src/mutation-hoc.tsx
@@ -60,11 +60,12 @@ export function mutation<
           <Mutation {...opts} mutation={document} ignoreResults>
             {(mutate, _result) => {
               const name = operationOptions.name || 'mutate';
-              let childProps = { [name]: mutate };
+              let childProps = { [name]: mutate, ..._result };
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@doniyor2109 this is actually intentional to keep backwards compat 100%. I'm curious why you wouldn't use the Mutation component instead?

doniyor2109:
```Mutation``` component is great, however I had some problems using ```mutate``` function.  I prefer using HOC instead of wrapping component. Wrapping component creates huge nested components when there are more mutations and queries.

Here is the problem I had which I could not solve by using ```Mutation``` component.

```jsx
class Form extends Component {
  handleSubmit = () => {
    // I want to call createUser from here
  }
  render() {
    return (
      <Mutation mutation={createUser}>
        {(createUser) =>{/* components */} }
      </Mutation>
    );
  }
}
```
I could only solve this problem with the help of mutation-hoc.

ekfn:
I'd love to get results in `mutation-hoc`. Is there any chance to see this in near future? @jbaxleyiii

PR #1849, opened 2018-03-26
Rerender

### LINKED ISSUE #1777: Apollo 2, useless re-render (twice render). Simple example, simple reproduction on codesandbox.

jbaxleyiii:
This is a variation of #1809 to fix #1777. Thanks @akiran for the first stab at this!

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 ### vNext
 
+### 2.1.1
+
+* Fix uneccesary rerender on cache hit
+
 ### 2.1.0
 
 * Officially release new components!
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -307,8 +307,19 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
 
   private startQuerySubscription = () => {
     if (this.querySubscription) return;
+    // store the inital renders worth of result
+    let current = this.getQueryResult();
     this.querySubscription = this.queryObservable!.subscribe({
-      next: this.updateCurrentData,
+      next: () => {
+        // to prevent a quick second render from the subscriber
+        // we compare to see if the original started finished (from cache)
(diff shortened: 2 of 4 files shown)


johnunclesam:
Can you release please?

stenin-nikita:
@jbaxleyiii The problem can not be solved if an error occurs.

[Edit React Apollo 2, twice rendering on error.](https://codesandbox.io/s/1yzy409vq)

Perhaps if an error occurred, do the same?

```javascript
    this.querySubscription = this.queryObservable!.subscribe({
      next: () => {
        // to prevent a quick second render from the subscriber
        // we compare to see if the original started finished (from cache)
        if (current && current.networkStatus === 7) {
          // remove this for future rerenders (i.e. polling)
          current = undefined;
          return;
        }
        this.updateCurrentData();
      },
      error: error => {
        if (current && current.networkStatus === 7) {
          // remove this for future rerenders (i.e. polling)
          current = undefined;
          return;
        }
        this.resubscribeToQuery();
        // Quick fix for https://github.com/apollostack/react-apollo/issues/378
        if (!error.hasOwnProperty('graphQLErrors')) throw error;

        this.updateCurrentData();
      },
    });
```

Issue #1852, opened 2018-03-27
"SyntaxError: Unexpected token import" error when importing in create-react-app test


excitement-engineer:
I get the error:

```
import * as React from 'react';
    ^^^^^^

    SyntaxError: Unexpected token import
```

When running tests in create-react-app.

I think the solution is to compile the source code down to CommonJs so it will work with this tool.

zamiang:
Believe this is related to https://github.com/apollographql/react-apollo/issues/1829

lucasconstantino:
I've basically cloned the repo, built react-apollo myself, and added the resulting lib directory inside my app for test only purposes. Quite lame.

excitement-engineer:
A fix is on its way! See #1855. Will be released soon:)

excitement-engineer:
It is merged and will be included in the next release

hiramhibbard:
Is there a workaround until the next release?

lucasconstantino:
Mine is sad, but works.

PR #1855, opened 2018-03-27
Compile test-utils with rollup

excitement-engineer:
This fixes issue #1852. 

Note, I had to downgrade `graphql` to 0.13 because 0.13.2 exposes `.mjs` files which don't play well with CRA.


excitement-engineer:
@jbaxleyiii, I got it to work, it now compiles correctly. 

To prevent the examples from breaking in the future I added a test script (see `test-examples.sh`). This script should be included in the CircleCI config

excitement-engineer:
I think that we should create a new release once this bug has been fixed.

PR #1858, opened 2018-03-27
Bring flow types back

thymikee:
**Summary**

For reasons unknown to me this PR https://github.com/apollographql/react-apollo/pull/1801 removed Flow type defs from the repo, rendering Flow useless. 
This PR brings the file back.

**Test Plan**

Spawn a `flow-check` command on CI with Flow warnings enabled, so it errors when types are missing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,9 +19,9 @@ jobs:
 
   Node.js 9:
     docker: [ { image: 'circleci/node:9' } ]
-    steps: 
+    steps:
       - checkout
-      - run: npm i 
+      - run: npm i
       - run:
           name: Jest suite
           command: npm run jest -- --ci --testResultsProcessor="jest-junit"
@@ -48,6 +48,7 @@ jobs:
       - checkout
       - run: npm i
       - run: npm run type-check
+      - run: npm run flow-check
 
   Preact:
     docker: [ { image: 'circleci/node:8' } ]
@@ -131,7 +132,7 @@ workflows:
       - Node.js 8
       - Node.js 9
       - Linting
-      - Typecheck 
+      - Typecheck
       - Preact
       - Browser
       - Server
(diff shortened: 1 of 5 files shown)


rosskevin:
Thanks @thymikee - flow types were removed due to lack of support from maintainers and the community.  Adding them back would mean they would just fall out of date again, and that is worse than providing zero flow files.

The best path forward for flow support is to submit them to flow-types and send a PR here to update the readme indicating flow support.

chollier:
thanks for the explanation — I spent a couple of hours yesterday trying to understand

damn that's a bummer 😢

jbaxleyiii:
@thymikee I really appreciate the effort but @rosskevin is correct :( Can we PR these to flow-typed?

thymikee:
Yep, I'll land a PR there soonish 🙂

thymikee:
If anyone is interested in helping: https://github.com/flowtype/flow-typed/pull/2041

saadq:
Unfortunately, there's going to be a pretty significant loss by moving this to `flow-typed`. Currently, `flow-typed` doesn't correctly handle importing dependencies on other libdefs, so because you did:

```js
import type {
  ApolloClient,
  MutationQueryReducersMap,
  ApolloQueryResult,
  ApolloError,
  FetchPolicy,
  FetchMoreOptions,
  UpdateQueryOptions,
  FetchMoreQueryOptions,
  SubscribeToMoreOptions,
  PureQueryOptions,
  MutationUpdaterFn,
} from 'apollo-client';
import type { DocumentNode, VariableDefinitionNode } from 'graphql';
```

Every single one of these types that you imported will be given a type of `any`. Currently, you can only use `import type` and `import typeof` inside the `declare module` itself, and even inside there `flow-typed` only supports importing types that are native to Flow, like React or Node stuff. Trying to import stuff from other libdefs like `graphql` will fail.

thymikee:
@saadq Yea, that's unfortunate, but what can I do :<

saadq:
Ah just wanted to mention it, don't think there's anything you could do from your side :/

PR #1881, opened 2018-04-01
feat: allow access mutation result in mutation-hoc

doniyor2109:


doniyor2109:
This does not support when multiple hocs are applied

PR #1882, opened 2018-04-02
Simplified the MockedProvider

excitement-engineer:
This PR aims to simplify the usage of the `MockedProvider`.

**Note, that the changes that I am proposing are breaking!**

### 1. Remove `removeTypename` prop.

Whenever I am writing a test I want to simply be able to write 

```
import { Query } from “./some_component”;

const mocks = [{
    request: {
         query: Query
    },
    result: someData
}]

renderer.create(
<MockedProvider mocks={mocks}>
   <Component />
</MockedProvider>
);
```

However, by default the `MockedProvider` adds typenames to everything in the cache. This caused the example above to fail and forces the user to write: 

```js
import { Query } from “./some_component”;


const mocks = [{
    request: {
         query: Query
    },
    result: someData
}]

renderer.create(
<MockedProvider mocks={mocks} removeTypename>
   <Component />
</MockedProvider>
);
```

This PR removes the `removeTypename` prop and simply forces the cache to not add typenames to documents it receives.

I hope that this change will simplify the usage of the MockedProvider and prevent users from having to worry about whether to remove typenames or not.

Are there any use-cases that I am missing by removing this prop? Would love to hear them. I couldn’t come up with a reason for this prop being there.

### 2. Remove the `client` prop

The `MockedProvider` allows the user to pass a custom `client` prop. However this `client` is passed straight to the the `ApolloProvider`. So I removed this prop and instead of letting the user do this:

```
<MockedProvider client={client} />
```

The user can use the normal `ApolloProvider`: 

```
<ApolloProvider client={client} />
```


rylanc:
I'm having trouble using fragments with this change.

When testing a component that uses fragments in its query, I'm getting the following warning from the `HeuristicFragmentMatcher`:
```
You're using fragments in your queries, but either don't have the addTypename:
  true option set in Apollo Client, or you are trying to write a fragment to the store without the __typename.
   Please turn on the addTypename option and include __typename when writing fragments so that Apollo Client
   can accurately match fragments.
```

Even if I add `__typename`, I still get the warning (I'm guessing this is because the cache isn't storing the typename, since it wasn't requested). Am I missing something here?

PR #1896, opened 2018-04-05
Optimise resubscription checks and allow shouldResubscribe to be a function

oliverwoodings:
### Checklist:

I started investigating this when i noticed that my subscribed components were frequently missing messages that were definitely sent to the browser. On looking at the websocket frames, I noticed loads of resubscriptions occurring when no variables had actually changed:

screen shot 2018-04-05 at 12 06 19

If you look closely, you can see `stop` requests being sent immediately after data frames have been received. All these stops match up to UI re-renders. My theory is that upon every re-render, the subscription gets torn down and recreated. During this recreation, if a message arrives, it gets ignored because nothing is currently subscribed to it.

To verify i tried setting `shouldResubscribe` to false. This worked; I no longer missed any messages, and the crazy amount of resubscriptions disappeared (obviously):

screen shot 2018-04-05 at 12 04 48

As far as I can tell, when the old subscription HOC was ported to the new `Subscription` component, the logic of checking whether we could bail from resubscription based on prop equality was [simply copy-pasted across](https://github.com/apollographql/react-apollo/blob/d58d9b85865eb799a9407bd523331e86d493a7f5/src/graphql.tsx#L174). This `shallowEqual` comparison check makes sense in a HOC where the props you are receiving are actually the ones destined for the inner component.

However in the new `Subscription` component this shallow equality check makes no sense anymore. If you pass the variables prop as a new object on each render, the equality check will fail every time. Forcing the user to pass the same object instance on subsequent renders is cumbersome. This PR attempts to fix this by doing the shallow equality check directly on the props, and then additionally checking the subscription query for equality as well.

Something else I noticed while fixing this was that in the old HOC world, `shouldResubscribe` was a function that was given the current props and the next props. In the new world, the `Subscription` component only accepts `shouldResubscribe` as a boolean which kinda makes it useless to consumers unless you want to permanently block resubscription. I've added back support for it to be a function.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,7 +121,7 @@
       "dev": true,
       "requires": {
         "@types/cheerio": "0.22.7",
-        "@types/react": "16.0.35"
+        "@types/react": "16.1.0"
       }
     },
     "@types/enzyme-adapter-react-16": {
@@ -203,7 +203,7 @@
       "integrity": "sha512-kmNh8g67Ck/y/vp6KX+4JTJXiTGLZBylNhu+R7sm7zcvsrnIGVO6J1zew5inVg428j9f8yHpl68RcYOZXVborQ==",
       "dev": true,
       "requires": {
-        "@types/react": "16.0.35"
+        "@types/react": "16.1.0"
       }
     },
     "@types/recompose": {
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@oliverwoodings thank you so much for this! Would you be up for a PR to the docs (in the apollo-client repo) showing this addition?

oliverwoodings:
Sure! I'll do it now

oliverwoodings:
@jbaxleyiii here you go: https://github.com/apollographql/apollo-client/pull/3281

PR #1901, opened 2018-04-06
SSR enable asynchronous refetch call

evans:
Adds a failing test and provides the fix for when refetch is called asynchronously from render. The component mounts, which starts the observable query subscription. Normally this would fetch data, causing a render, and the instance of refetch would be overridden with the force network. In cases of SSR, this second render does not occur. The refetch passed in the original render call needs to take this into account.

### Checklist:


jakehasler:
Is it clear which released version of react-apollo this is contained in?

Issue #1899, opened 2018-04-05
Warning: Apollo(xxx): Did not properly initialize state during construction. Expected state to be an object, but it was undefined.


saostad:
I am upgrading my project to react 16.3.1 but in components with HOC generated with react-apollo I am getting this errors and my app crashed.

> "Apollo(SubmittedFormsList): Did not properly initialize state during construction. Expected state to be an object, but it was undefined."

this is the code I am expecting to work (Recommended By React): 

Error: "TypeError: prevState is null"
```
static getDerivedStateFromProps(nextProps, prevState) {
    if(nextProps.showModal !== prevState.showModal) {
      console.log('recived new prop');
      return { showModal: true };
    } else {
      return null
    }
  };
```
and this is the code actually works for me (NOT Recommended By React): 
```
componentWillReceiveProps(nextProps) {
    if(nextProps.showModal !== this.state.showModal) {
      console.log('recived new prop');
      this.setState({ showModal: true }) ;
    };
  };
```

DriesH:
We are having the exact same issue, and I think it is due to the fact that react apollo is using an outdated hoist-non-react-statics that doesn't support React 16.3, but I'm not 100% sure on that... In our case the app doesn't crash, but just throws that warning in the console. Im going to debug it a bit more and if I find the cause, I'll report back!

bogdansoare:
same issue here

DriesH:
This should normally be fixed. Be sure you have te latest version of react-apollo by running `yarn upgrade react-apollo`.

hh83917:
Seeing same warning here.

lorrainefox:
I'm pretty sure that you should use this.state.showModal in getDerivedStateFromProps, so the compare should be the same as it is in componentWillReceiveProps:
if(nextProps.showModal !== this.state.showModal)

oliviertassinari:
This issue should be closed now.

PR #1910, opened 2018-04-06
Bumped version of hoist-non-react-statics to 2.5.0

DriesH:
Bumped version of hoist-non-react-statics to 2.5.0 for compatibility with react 16.3.0

Fixes issue #1899

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,7 +121,7 @@
       "dev": true,
       "requires": {
         "@types/cheerio": "0.22.7",
-        "@types/react": "16.0.35"
+        "@types/react": "16.1.0"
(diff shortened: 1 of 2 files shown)


thealjey:
when will this be released?

DriesH:
Hmm yea, indeed, that's strange... Because on npm it said the package was last published 5 days ago around the same time this got merged, so I thought this fix would be in...

DriesH:
For some reason I have version 2.5.0 of hoist-non-react-statics inside my lock file, no clue how it got there, but it does fix the issue... Hmm only thing we can do is wait I guess..

PR #1914, opened 2018-04-08
Add resolutions to pkg.json in TS example

jaredpalmer:
This is a way to guarantee that users install graphql@13.0 instead of 13.2 (which is broken) 

https://github.com/graphql/graphql-js/issues/1248

### Checklist:

### DIFF
--- a/examples/typescript/package.json
+++ b/examples/typescript/package.json
@@ -11,8 +11,13 @@
     "react-apollo": "../../lib",
     "react-dom": "^16.2.0"
   },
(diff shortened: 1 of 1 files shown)


hwillson:
Thanks for submitting this PR @jaredpalmer and sorry for the delay in getting back to you. We're going to keep `@types/graphql` pinned at 0.12.7 for now, since we're not ready to make the jump to 13.x yet (especially since 13.x introduces a read only graphql AST). Thanks again!

Issue #1869, opened 2018-03-29
Loading true when skipped


lucasconstantino:
**Intended outcome:**
When I configure a query to be skiped, I expect the injected `loading` prop not to be `true` for the data is not acctually being loaded.

**Actual outcome:**
The injected `loading` prop is set to `true` when the query is skipped.

**How to reproduce the issue:**
```js
const Comp = () => (
  <Query query={ query } skip>
    { ({ loading }) => loading ? 'Loading!' : 'Not loading!' }
  </Query>
)
```

**Version**

* apollo-client@^2.2.5
* react-apollo@2.1.0

**Issue Labels**

- [x] has-reproduction
- [ ] feature
- [ ] docs
- [ ] blocking
- [ ] good first issue

ZwaarContrast:
Running into the same issue with:

- react-apollo@2.1.2
- apollo-client@2.2.1

AdrianZielonka:
Running into the same issue with:

react-apollo@2.1.3
apollo-boost@0.1.4 (apollo-client@2.2.8)

AdrianZielonka:
Still an issue with:

react-apollo@2.1.4
apollo-boost@0.1.6 (apollo-client@2.3.1)

lucasconstantino:
@jbaxleyiii any opinion on this?

FranciscoGileno:
Running into the same issue here.

react-apollo@2.1.4

We managed to workaround this relying on networkStatus too. Not sure if it works for everyone.

```
const Comp = () => (
  <Query query={ query } skip>
    { ({ loading, networkStatus }) => loading && networkStatus !== 1 ? 'Loading!' : 'Not loading!' }
  </Query>
)
```

Gotcha: The first time the query loads, networkStatus === 1, so the loading does not show.

Any suggestions?

andreylh:
Running into the same issue here.

react-apollo@2.1.9
apollo-boost@0.1.10 (apollo-client@2.3.5)

veryscarycary:
Still an issue.

react-apollo@2.1.8,
apollo-boost@0.1.10
apollo-client@2.3.5

andrewtpoe:
I'm running into the same issue. 

react-apollo@2.1.9
apollo-client@2.3.7

Can provide any other details as needed.

zapo:
We workaround it by using our own Query wrapper as suggested in #1909 

```
[...]

PR #1916, opened 2018-04-09
Fix query result when skip=true (#1869)

edorivai:
Should fix https://github.com/apollographql/react-apollo/issues/1869

I added a test case, and took the implementation [from the original PR](https://github.com/apollographql/react-apollo/pull/1694/files#diff-45a733ae4e94396f8fa0947302418680R219) that aimed to implement the `skip` prop on `<Query>`.

I suspect this piece of logic got lost in the [big rewrite](https://github.com/apollographql/react-apollo/pull/1715) mentioned [here](https://github.com/apollographql/react-apollo/pull/1694#issuecomment-368563365).

### Checklist:


ZwaarContrast:
Any progress on this?

caseywebdev:
Ran into this issue yesterday, can a maintainer chime in here? I really like the `Query` component but I also need the `skip` and `loading` values to work correctly 😉

src/Query.tsx
```diff
@@ -358,30 +358,46 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
-    if (loading) {
-      Object.assign(data.data, this.previousData, currentResult.data);
-    } else if (error) {
-      Object.assign(data, {
-        data: (this.queryObservable!.getLastResult() || {}).data,
-      });
+    // When skipping a query (ie. we're not querying for data but still want
+    // to render children), make sure the `data` is cleared out and
+    // `loading` is set to `false` (since we aren't loading anything).
+    if (this.props.skip) {
```
hwillson: Quick note here - I've re-arranged this a bit to short circuit un-necessary extra query logic, when the query is to be skipped. One thing to add though, is that this means we're leaving `networkStatus` as `undefined`. This is currently intentional, as setting this to `NetworkStatus.ready` can cause problems when using something like the  `graphql` hoc and calculating the `skip` value using a function (forcing a `NetworkStatus.ready` value can prematurely end rendering). Leaving `networkStatus` as `undefined` here seems to be okay though, and is better than the way it was before when we did the `skip` check at the end of `getQueryResult`. That left `networkStatus` in `NetworkStatus.loading` state, even though we weren't loading anything.

caseywebdev:
What's the status on this one? Seems like it's good to go (besides conflicts).

rosskevin:
Please resolve conflicts, otherwise I think this good to merge

edorivai:
@hwillson are you picking up the conflicts, or would you like me to do it?

hwillson:
@edorivai I'll take care of it shortly. Thanks!

stephane-ruhlmann:
Is anyone taking care of this?

Kjagd:
@hwillson What's the status on this? We've been greatly affected by this for quite some time now.

pascalstr:
@hwillson what is the status of resolving these conflicts and merging this PR?

edorivai:
Great to hear :tada:

sbogdaniuk:
SO you just fixed this issue for Component, and for HOC not, right?

PR #1945, opened 2018-04-24
Include ApolloClient in MutationResult

cooperka:
ApolloClient is already exposed in the `Query` component; this adds it to `Mutation` as well. For example, after this PR you can call `client.resetStore()` after logging in. 

Tested and verified locally.

/label feature


cooperka:
This change causes a lot of snapshots to update. I'm not informed enough to know if the updates are expected. Jest also seemingly isn't respecting my `--updateSnapshot` flag. I'll let a maintainer verify and commit those when they have time.

PR #1951, opened 2018-04-26
Allow minor releases of lodash

cherewaty:
## Purpose
On a project where I use `react-apollo` and `lodash`, I got a warning from `duplicate-package-checker` that I have two versions of `lodash`. It'd be nice to only have one.

## Approach
Allow minor releases of `lodash` above the specified in `package.json`, so projects consuming `react-apollo` will get the latest `lodash@4.x`.

## Prior work
https://github.com/apollographql/react-apollo/pull/1478 moved `lodash` from `devDependencies` (where being locked into a specific patch release was fine) to `dependencies`.


cherewaty:
https://github.com/apollographql/react-apollo/pull/2021 upgraded `lodash`, but did not unpin it. Any reason to continue pinning the `lodash` version?

cherewaty:
Thanks @abernix. I rebased to get a green light from Circle, so this should be ready to merge.

PR #1959, opened 2018-05-02
merge variables from props and invocation options

spypsy:
Before v2.0, variables added to the mutation declaration in the `graphql` wrapper, were merged with any `variables` that were added when the mutation was invoked. I believe this is a nice-to-have since a mutation may have multiple ways of being invoked in a component, but some variables may remain the same. For example we're using a `userId` from the props in the mutation declaration `options` and then add whatever variables necessary when calling the mutation in the component. 

We can of course add `userId` to all the calls but it's pretty nice to have it declared just once.

What does everyone think?


samginn:
Hi! This caught me up as well. I believe this was an undocumented (or at least I couldn't find it in the changelog) breaking change that occurred which we caught while upgrading react-apollo. Would love this functionality to be included again as we use this pattern quite frequently.

rosskevin:
All tests must be passing for a PR to be considered. Please address them or close the PR.

Issue #1873, opened 2018-03-30
Component created with graphql() does not update query variables when props change


tuananhtd:
**Intended outcome:**

When create react component using `GraphQLComponent = graphql(query, options)(MyComponent)` it is expected that we can pass query variable via `GraphQLComponent`'s props and but I found a bug that when that props change the graphql query isn't got updated. You can take a look at the `Query` component in the picture below.

![inspected](https://user-images.githubusercontent.com/14288059/38142838-2a4323e8-3468-11e8-9dfa-1caf6e59cdaa.gif)

**Actual outcome:**
The graphql query isn't updated according to the props change.

**How to reproduce the issue:**
I created a repo to reproduce a code [here](https://github.com/tuananhtd/apollo-react-bug-reproduce).

**Version**

* apollo-client@<2.1.0>
* react-apollo@<1.3.4>

I believe [this line](https://github.com/apollographql/react-apollo/blob/master/src/query-hoc.tsx#L57) cause the bug. It eventually sets `operationOptions.options.variables` and cause the subsequence render thing that we already pass the variables from the beginning. Because of this we have to pass a non empty option to the graphql function to reproduce the issue. In my case I used this:
```
const BookGraphQLComponent = graphql(QUERY, {options: {pollInterval: 10 * 1000}})(Book);

```

doomsower:
I'm also facing this bug. This works:
```ts
graphql(query, { options: () => ({ fetchPolicy: 'network-only' }), alias: `${formName}FormQuery` }),
```

However, this return stale date with previous variables:

```ts
graphql(query, { options: { fetchPolicy: 'network-only' }, alias: `${formName}FormQuery` }),
```

Something is wrong with `mapPropsToOptions`, on this screenshot it totally ignores sourceId from props and uses old sourceId value:
![image](https://user-images.githubusercontent.com/12031673/38468533-21380798-3b50-11e8-9d41-d3b1e4722542.png)

hyperborea:
Facing the same issue, wrapping options in function as suggested by @doomsower works.

PR #1968, opened 2018-05-07
Don't mutate options object when calculating variables from props

ksmth:
This small change fixes https://github.com/apollographql/react-apollo/issues/1873


src/query-hoc.tsx
```diff
@@ -51,7 +51,7 @@ export function query<
       render() {
         let props = this.props;
         const shouldSkip = mapPropsToSkip(props);
-        const opts = shouldSkip ? Object.create(null) : mapPropsToOptions(props);
+        const opts = shouldSkip ? Object.create(null) : { ...mapPropsToOptions(props) };
```
hwillson: `options` can be passed into the `graphql` hoc as either an object or a function. Spreading the result of `mapPropsToOptions` here can happen in both cases - when `options` is an object or a function. In the case of a function though, the return value of the `options` function is already going to be a new object (see https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-query-config-options). So we really only need to spread the result when we're using an `options` object. I'll adjust this a bit.

hwillson:
We should be good to go here @ksmth (left the spread as is). LGTM - thanks!

Issue #1977, opened 2018-05-12
If initial Query errors, data is undefined instead of empty object


TLadd:
**Intended outcome:**
The data prop passed to the Query component is never undefined. That's my understanding anyways, from reading https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L89

**Actual outcome:**
If your first request for a Query errors, data ends up being null. It ends up hitting this line, https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L378, which overwrites the default-created empty object and replaces it with `(this.queryObservable!.getLastResult() || {}).data,`, which will be undefined if the query doesn't have a result yet

**How to reproduce the issue:**
On any react-apollo application with a Query, force an error to occur and log the data passed to children. I personally encountered this when I had accidentally not had my graphql server running. As a result, the first Query request failed due to a network error, and I got an exception because I was assuming data would be an empty object and not undefined.

**Version**

* apollo-client@2.3.0
* react-apollo@2.1.4

alamothe:
The typings say `data` is either `TData` or `undefined` - never empty object: https://github.com/apollographql/react-apollo/blob/3985b47f086f38f891099732842ca2f16c7966c4/src/Query.tsx#L97

So which one is it?

PR #1983, opened 2018-05-15
Query: Fix data is undefined on error

TLadd:
If the query fails on the first result, the query result data was being returned as undefined
Now this case will result in data being an empty object

Fixes https://github.com/apollographql/react-apollo/issues/1977


hwillson:
Thanks for working on this @TLadd! I agree with your assessment that the `data` result should be an empty object. Your changes look great, but I'm a little concerned that this is a breaking change. Some people have likely noticed that `data` is `undefined` when there is an error, and are leveraging that fact in their error handling logic. Rolling this change out could affect that since their `!data` checks will now fail. Hmmm - we'll need to think about this a bit more ... 🤔.

TLadd:
Using !data to check for errors would be pretty brittle anyways, since it only happens if there isn't already a previous result. Certainly could be possible that there is code out there depending on it though.

alamothe:
The typings say `data` is either `TData` or `undefined` - never empty object: https://github.com/apollographql/react-apollo/blob/3985b47f086f38f891099732842ca2f16c7966c4/src/Query.tsx#L97

So which one is it?

TLadd:
Thanks for getting it across the finish line!

src/Query.tsx
```diff
@@ -94,7 +94,7 @@ export interface QueryResult<TData = any, TVariables = OperationVariables>
   // I'm aware of. So intead we enforce checking for data
   // like so result.data!.user. This tells TS to use TData
   // XXX is there a better way to do this?
-  data: TData | undefined;
+  data: TData | {};
```
derekclair: 🥇

danilobuerger:
@hwillson @TLadd this should be reverted, see #2424 . This PR promised:
```
instead of `data && data.user` you can just check `data.user`
```
instead it made things worse. Now a type guard is required to discriminate between `TData` and `{}`.

tomitrescak:
Yes please, revert this change. It makes this VERY difficult to use in Typescript ;(

hwillson:
@danilobuerger @tomitrescak Rolling back and we'll have a new release out this morning - thanks!

kallaspriit:
So what ise the solution? Also the TypeScript example does not seem to work (any more) and it uses the old method of:

examples/typescript/src/Character.tsx
```typescript
if (!data) return <div>no data</div>;
```

hwillson:
@kallaspriit For now we'll roll this back, then we'll come up with another solution afterwards (we really didn't want breaking changes in this release). The examples are in the process of being updated and will be available shortly. We're also going to wire the example app tests into CI, to help keep them healthy.

kallaspriit:
Wrote my comment seconds before seeing your reply. Sounds great :)

tomitrescak:
@fenok that is not true. Those are fundamentally different concepts and would validate very differently. Your example also allows partial objects, while the former allows only complete objects or empty objects

fenok:
Valid point, I was trapped in my partial results world. I also found that the idea of `Partial` has already been discussed and rejected. Sorry for the interruption 😅

evenfrost:
So, what's the current status of it?

derekclair:
I believe it was like Typescript@2.5 or something, they introduced a `unknown` type; which, I believe addresses this issue perfectly! If I could request, [a code-review please](https://github.com/derekclair/react-apollo/commit/b9f1e5b2993bb634d7e83cd96e4f5790dc860da3), or two, on this fix that implements that type. I could really use another set of eyes on it.

lucasconstantino:
@evenfrost it was changed, but it's changed back from the previous change, which actually changed to keep things the same they were before the change, so that we don't really change what was change before.

IanVS:
Thank you so much for clarifying, @lucasconstantino.  I was getting really lost in trying to understand where things stand.  💫

PR #1989, opened 2018-05-19
Implemented the ManualQuery component

excitement-engineer:
The react-apollo [docs](https://www.apollographql.com/docs/react/essentials/queries.html#manual-query) on manually firing a query mention that this can be done by directly grabbing the client using the `ApolloConsumer` and then calling the `query` method on it. However this does leaves some issues with the user w.r.t. state management of the `loading`, `error` and `hasBeenCalled` properties. This has to be implemented by the users of the library themselves.

I decided to implement the `ManualQuery` that abstracts all the state management away and makes it easy for the user to perform ad-hoc queries that are not tied to an observable (like the `Query` component is). 

It can be used as follows: 

```js
<ManualQuery query={SOME_QUERY}>
{(load, { loading, error, data, called}) => {
   if (!called) {
       return <button onClick={load}>Load some data</button>
   }

if (loading) {
 return “Loading…”
}
if (error) {
  return “Error”
}

return this._renderData(data)
}}
</ManualQuery>
```

The `load` function triggers the query and the second parameter of the component contains the state of the query.

I personally have had quite a few use cases where this component would come in handy and simplify the fetching logic.

Note, this is an initial proof of concept and may need to be expanded upon to meet more use-cases.

Would love to hear feedback.


hwillson:
Thanks for your work on this @excitement-engineer! Unfortunately, this isn't something we're looking to add at this time. As mentioned in https://github.com/apollographql/react-apollo/pull/1989#issuecomment-390477070, leveraging the `skip` property could help achieve a lot of the same functionality, and if not, using the `ApolloConsumer` approach is already available as an escape hatch. Between those 2 methods, a lot of the bases should be covered. While the `ManualQuery` approach could be useful, it's something that might be better handled through documentation/examples or as a 3rd party package. We definitely want to make sure `react-apollo` is useful and covers as many developer use cases as possible, but we'd also like to keep it fairly focused and easy to maintain.

We'll definitely keep this idea in mind though, and re-visit this PR if/as needed. In the future though you might want to open a new issue with an idea like this first, so it can be discussed as a feature request and gauge community interest, to save you some precious time.

Thanks again for all that you do @excitement-engineer! 🙇

Issue #1931, opened 2018-04-17
loading is always True in Query render prop


Jinnified:
I am using query render prop, which been wrapped by a redux render prop, which provides a new query for a new fetch. But when the new fetch been triggered, the loading is always true in the Query render prop.

expected: the loading state should be false once data has been returned
actual: the loading state is always true

It seems that the loading is always true if data is the same? but it's not consistant

apollo-client: 2.2.8
react-apollo: 2.1.3

Code:
```
<WithRedux id={id}>
          {
            ({ d }) => {
              return (
                <Query query={getQuery(d)}>
                  {({ loading, error, data, refetch, networkStatus }) => {
                    console.log('.....networkStatus', networkStatus);
                    console.log('.....loading', loading);
                    console.log('.....data', data);
                    if (error) return <CompPlaceholder cnt={1} template={0} />;
                    if (loading) return <CompPlaceholder cnt={1} template={1} />;
                    if (data && Object.keys(data).length > 0) {
                      if (data.flightInfo && data.flightInfo.length > 0) {
                        return (
                          <DetailFlightInfoCard id={id} data={data.flightInfo} />
                        );
                      }
                      return null;
                    }
                    return null;
                  }}
                </Query>
              );
            }
          }
        </WithRedux>
```

apollo client with batchLink and cache
```
new ApolloClient({
    link: batchLink,
    cache: new InMemoryCache(),
    connectToDevTools: true,
  });
```

lucasfeliciano:
Having the same issue here

ju5td0m7m1nd:
Facing same problem.
When response data exceeded certain size, it didn't catch the response with data.

caseywebdev:
I’m seeing `loading === true` And `networkStatus === 1` when using `skip={true}` which I certainly didn’t expect. No request is actually being sent (which is correct) but the props passed to the render function would suggest one is.

powerdefy:
same problems please fix it.

nfantone:
Experiencing the same thing here.

![image](https://user-images.githubusercontent.com/476069/49313959-d187f500-f4e0-11e8-9e53-9435f3277951.png)

Above you can see two consecutive queries, both returning an empty array as `data`. First one works fine, switching `loading` flag to `true` and then back to `false. Second one, freezes at `true` and never flips back.

![image](https://user-images.githubusercontent.com/476069/49314023-0eec8280-f4e1-11e8-8ef3-c7add21b81c2.png)

[...]

PR #1990, opened 2018-05-19
Created a reproduction of the issue with the skip property.

excitement-engineer:
Created a reproduction of an issue where setting the prop `skip` to true in the `Query` component caused the `loading` value to always be true. I have a feeling that this is an issue in `apollo-client` given that the `loading` value is retrieved by calling the `currentResult()` method which is part of `apollo-client`.

Reproduction of issues: #1931, #1869

### DIFF
--- a/test/client/Query.test.tsx
+++ b/test/client/Query.test.tsx
@@ -688,6 +688,27 @@ describe('Query component', () => {
         done();
       });
     });
+
+    it('skip property', done => {
+      const Component = () => (
+        <Query query={allPeopleQuery}>
+          {result => {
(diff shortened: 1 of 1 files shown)


hwillson:
Thanks for this @excitement-engineer. It looks like this is already covered by https://github.com/apollographql/react-apollo/pull/1916, which we're close to getting merged. Thanks again!

nfantone:
@hwillson I know this is old now, but since @excitement-engineer pointed to #1931, I wanted to make you aware that that's still an on-going issue and has not been resolved by #1916.

PR #1995, opened 2018-05-21
Default the `addTypename` prop in the MockedProvider to false

excitement-engineer:
This PR defaults the `addTypename` prop in the `MockedProvider` to false. 

Before, the default for the `addTypename` was true which forces users of the MockedProvider to add the `__typename` key to their queries in order to get the MockedProvider to work.

This default behaviour may cause confusion because if users want to user the MockedProvider for the following query: 

```graphql
query GetUser($username: String!) {
    user(username: $username) {
      id
    }
  }
```

Then the MockedProvider wouldn’t work because the query does not include the `__typename`. 

I think that users should just be able to inject any query that they are using (regardless of whether it includes `__typename`) and have the MockedProvider work without first setting `addTypename` to false. This will make the MockedProvider more user-friendly.


hwillson:
As discussed in https://github.com/apollographql/apollo/pull/164#discussion_r193488876, we're going to leave `addTypename` as is for now due to potential backwards compatibility issues. I've added this to a new `Release 3` milestone, and we'll be sure to re-visit closed release 3 PR's. Thanks @excitement-engineer!

Issue #1314, opened 2017-11-01
loading is false, but no data is present (and no error)


adamdonahue:
Just upgraded to Apollo 2.0, using react-apollo 2.0.0 -- it now appears that the graphql props option for a query is passed a data argument that can have loading set to false but no data.

That is, given a query, say:

```gql
query X($id: ID!) {
  myItem(id: $id) {
    id
    field
  }
}
```

and a `graphql` call:

```javascript
graphql(query, {
  props({ data: { loading, error, myItem } }) {
    console.log('loading =', loading);
    console.log('error =', error);
    console.log('myItem =', myItem);
    return { ... }
 }
})(MyComponent)
```

We're seeing the output:

```
loading = true
error = undefined
myItem = undefined
loading = false
error = undefined
myItem = undefined
loading = false
error = undefined
myItem = {<data>}
```

That is, there is an interim state between `loading` having been set to false but `myItem` yet available on the `data` argument.  

Is this expected, as it appears to differ from the behavior of Apollo 1.0.  And if so, what is the reliable way to check that data was loaded and _is ready for use_ in the component? 

We're now using `(loading || !myItem)` to check whether something is loading.

Panoplos:
I can confirm this issue exists.

![loading false](https://user-images.githubusercontent.com/2549985/32585706-169ffda6-c542-11e7-85cd-4ec15b60298b.png)

![no data](https://user-images.githubusercontent.com/2549985/32585712-1c9f0e2c-c542-11e7-96b7-8efce772dd31.png)

Offending code:

```es6
 graphql(
    gql`
      query localise($locale: LocaleType!, $resourceIds: [String]) {
        l10nSlugs(locale: $locale, resourceIds: $resourceIds) {
          id
          resourceId
          slug
        }
      }
    `,
    {
      name,
      options: ({ locale }) => ({
        variables: {
          locale,
          resourceIds
        },
      }),
      props: props => ({
        l10n: {
          /**
           * Whether or not the graphql resource is loading.
           */
          loading: props[name].loading,
          /**
           * Conveniently offer the locale.
           */
          locale: props.ownProps.locale,
          /**
           * And the BCP47 normalise version.
           */
          bcp47Locale: normaliseBCP47Tag(props.ownProps.locale),
          /**
           * Localise a slug with insertions.
           * @param {string} resourceId Resource ID of the slug to localise.
           * @param {string} slug Default slug in case loading.
[...]

PR #2003, opened 2018-05-23
Refetch queries on underfetch

steelbrain:
## Situation

When an object is mounted, and a mutation is executed that returns the same ID as the mounted object but has less fields, `loading` is returned as `false` but `data` is an empty object as returned from this line in [`apollo-client::QueryManager`](https://github.com/apollographql/apollo-client/blob/2a88ed6413c684d40ee0fc0edf25f6d83ae27f26/packages/apollo-client/src/core/QueryManager.ts#L984).

An internal error is generated of course with the message template `Cannot find field $FIELDS on Object` but that's swallowed and never shown to the user in the code referenced above.

A repro for this error can be found at [stevehollaar/react-apollo-error-template@`mutate-bug`](https://github.com/stevehollaar/react-apollo-error-template/tree/mutate-bug)

## Possible fixes

There are multiple fixes that were evaluated for this, each with it's own downside

- Automatically "fill in" the missing fields in outgoing mutations based on the types of results currently being watched, so in case we get a result from server that has ID that's already mounted, we will have all the fields we need. This would've caused a lot of over-fetching and issues in large codebases.
- Do not automatically update the mounted components, like relay. Let users do it the `...fragment` way (which fills in the fields automatically) or specify a manual store updater.
- After a mutation that causes issues is executed, detect the errors and refetch the query

The last one seemed to be the most straight forward and simple solutions with no side effects so that's what I've done here.

This PR fixes https://github.com/apollographql/apollo-client/issues/3267 https://github.com/apollographql/react-apollo/issues/1314

## Credits

Thanks @stevehollaar for his repro-case and constant support in this quest

## Todo

- [x] Fix breaking tests
- [x] Add new tests


mlcohen:
@steelbrain @jonas-arkulpa This is fantastic! We've been running into this issue and have been debating how to resolve the problem. I agree that automatically "filling in" missing fields is the wrong way to go. We've gone down the path of adding fields causing errors and applying fragments. Not great 😕. Your solution to instead refetch data is a good option, however, I wonder if the query manager could be further refactored so that an optional hook/plugin could be provided to the QM. The QM could then call the hook to delegate how best to handle a detected error. I say this since in our particular scenario, the detected field error pops up not because of direct GQL mutation but, rather, from our system running one or more queries whenever our app receives pushed system events (via websocket). Depending on the circumstance, we could either use a default refetch action (like your solution provides) or in turn perform our own custom error handling that best serves our purposes. Thoughts?

For more detail of the problem we have run into, please see https://github.com/apollographql/apollo-client/issues/2920#issuecomment-363546863

steelbrain:
@mlcohen I also have [a branch in `apollo-client`](https://github.com/steelbrain/apollo-client/tree/return-error-for-partials) that [returns/exposes partial errors to the codebase](https://github.com/steelbrain/apollo-client/commit/eed91a3961e4f838aeac10474b6f2cde658ae9f2) so they can be handled but I am not sure if we should expose them to the React component.

The fact that Apollo React `Query` component returns an object with properties such as `loading` and `data` makes it work to our advantage that we can just add it to that object instead of changing function invocation signature.

Thoughts?

mlcohen:
@steelbrain Yeah, I can certainly see your point directly exposing the error to the React component. We were thinking about updating the apollo client's query manager directly instead of indirectly via a React component. Applying the hook directly against the underlying apollo client instance would provide a way for the system to "globally" handle detected field errors. This means that React components that use apollo client would not be exposed to these types of errors. The global default would be for the QM to do an automatic refetch.

Your idea of adding a `noRefetchOnPartial` prop also sounds intriguing. Perhaps both options are viable. A "global" approach by directly adding a hook the apollo client's QM and a more isolated approach that overrides the global approach when a `noRefetchOnPartial` prop is supplied. Hmm 🤔

steelbrain:
Because including `partialError` in React requires that we also expose it in `QueryManager` I think we should do both.

Other than refetching or handling it on a component-level basis, I am curious how do you intend to handle it directly in Apollo Client? Unless, you are not using React for rendering UI.

mlcohen:
In our application we perform GQL queries and mutation through `react-apollo`'s `graphql` HOC and directly via the apollo client instance calling its `query` and `mutate` methods. Most of the time we use the `graphql` HOC for our React components but where we receive pushed system events, we have actions that use the apollo client instance. Based on these two approaches, it seems that if the `partial` field is exposed in the result, we could handle the error by doing a refetch in our own way. It would be ideal if the `graphql` HOC and apollo client's `query` and `mutate` could perform a data refetch just like what you're doing in the new `Query` component. Of course that likely means duplicating the logic in three places to do the same thing. The common point among the three is within the query manager's `getCurrentQueryResult` method. Perhaps a hook within the QM to handle partial errors isn't necessary since a data refetch seems to suffice, but logic that is more centrally focused to handle partial errors could help.

peisenmann:
If you're calling `data.refetch()`, then you're going to be hitting the network and forcibly updating the local data state to the latest state on the server, right? Even if the user has done a mutation on a small portion of a larger data structure, the user doesn't necessarily want the rest of the cache updated.

What I'd consider the expected behavior is for the response of the mutation to simply overlay/merge into the existing cache, and then fire renders for all subscribers from the current cache state, not from the network response.

steelbrain:
@peisenmann While merging is possible, it's also possible that other fields may have changed 
by mutation and without refetching you are showing outdated data to the user. This is use-case specific tho.

Either way that cannot be done in `react-apollo`, it'd have to be done in `apollo-cache-inmemory` or at least `apollo-client`. This is the least side effective and straight forward way I've evaluated to be (see first post for other methods considered/tried)

peisenmann:
@steelbrain A perfectly valid point, for sure. I'd expect in that case, if you wanted to check for those other fields, you would have them in the query part of your mutation or you would use refetchQueries to trigger the other, bigger query to be re-invoked. By excluding those fields from my mutation, I'm signaling that I don't want them updated at this time.

In an application I'm working on, when a user queues up local changes that are not ready to be persisted to the server yet, we use a writeFragment to write the data back into the Apollo cache. This keeps all views up to date with the user's changes, as long as no rogue refetches are called.

As you've noted, I will take my "expected behavior" points up with the other `apollo-*` libs. My request for this thread is not to introduce automatic refetches (or at least to allow me to specify `noRefetchOnError` or something of that nature).

idris:
@peisenmann see the original issue (https://github.com/apollographql/apollo-client/issues/3267).  Normally the mutation payload _does_ get merged into the cache.  The code in this PR only gets activated when the mutation payload has under-fetched data, and a component is trying to render with that under-fetched data.  In that case you cannot simply merge with the cache because the cache doesn't have what you need.

peisenmann:
@idris Ah, okay, so the original full query is not even populated in the first place. @steelbrain Apologies for wasting your time. Thanks for the PR!

dan-weaver:
@idris I'm not following what your saying but am desperate to understand this behavior better as I *keep* running into it.  @peisenmann I think this occurs in the case that the original full query *is* there.  Someone correct me if I'm wrong, but that is the behavior I am seeing.

So I'm probably overlooking something and am going to look foolish once you explain @idris, but what do you mean here exactly:

>The code in this PR only gets activated when the mutation payload has under-fetched data, and a component is trying to render with that under-fetched data. In that case you cannot simply merge with the cache because the cache doesn't have what you need.

Why is that not possible? You should be able to theoretically merge any fields that *were* in that mutation payload into the cache and not disturb the ones that it *didn't* fetch no?

felixk42:
Would really love to see this PR merged, it's been consistently a serious deal breaking bug for at least 10 months now...

hwillson:
@felixk42 I'm close to getting this merged in, but this change could have implications on a few edge cases. Regardless, you should see more updates here today.

jonas-arkulpa:
@hwillson thanks for you work & [your blog post](https://dev-blog.apollodata.com/evolving-apollo-repository-maintenance-a2f26f319211)!

As this issue is the biggest blocker for (not only) our projects I think it clearly needs more thoughts putting into it. We are really happy to finally have a usable workaround, but it clearly is just that, a workaround.

hwillson:
Thanks @jonas-arkulpa! 

Just a quick heads up for everyone here. I cleaned things up a bit, ran some more tests, and have had a few more discussions about this approach. While this PR does fix the original issue, we're still a bit concerned about the potential performance impact this could have on larger applications. We're currently digging a bit deeper to see if this can be resolved in another way, but we haven't ruled this potential solution out yet. Sorry for the delay all - this issue has been around for a while, and we can't wait to get it fixed, but we just want to make sure the fix doesn't cause more problems. Thanks for your patience!

idris:
@hwillson we (me and my team, original authors of this PR) would be happy to contribute to those discussions and provide input.  Happy to do that here or in Slack.

fracmak:
Is the concern with this approach that react-apollo shouldn't have to be concerned with whether we're missing data or not? apollo-client should realize that we're observing a set of data that has been written into it's cache with a partial result and fetched the missing data for us?

raelmiu:
How does one test if this is what happening?
I have an array of items that I'm subscribing to changes on, but when I trigger an update every item but the updated one disappears. Trying to nail down if this is the same issue?

idris:
@raelmiu try the error template app here for a repro: https://github.com/apollographql/apollo-client/issues/3267

pleunv:
Just wondering if you're any closer to getting this merged? Ran into some other issues (`errorPolicy: 'all'`) that appear to be fixed by upgrading from 2.0.x to 2.1.x, however my upgrade is still blocked by this issue :)

cagdastulek:
(Also posted on apollo-client issue https://github.com/apollographql/apollo-client/issues/3267)
No action on this issue so far. 😞  
Here is another test to show the issue. Basically, what happens is that if query A is already in the client cache and later same query A and a new query B run at the same time, where query B is a superset of query A, the expected behavior is that query B to not return anything from cache and wait for network response as client cache store does not have all the data to fulfill query B's request. However, the current bug results in FooView which was wrapped by query B to get data with missing fields.

```javascript
import { InMemoryCache } from "apollo-cache-inmemory";
import { ApolloClient } from "apollo-client";
import { ApolloLink } from "apollo-link";
import { SchemaLink } from "apollo-link-schema";
import Promise from "bluebird";
import { assert } from "chai";
import { mount } from "enzyme";
import gql from "graphql-tag";
import { makeExecutableSchema } from "graphql-tools";
import React from "react";
import { ApolloProvider, Query } from "react-apollo";

describe("Cache-And-Network", () => {
  function createTestSchema() {
    const typeDefs = `
  type Query {
    foo: Foo!
  }
  
  type Foo {
    id: ID!
    field1: String
    field2: String
  }
`;

    const resolvers = {
      Query: {
        async foo() {
          return {
            id: "id5",
            field1: "value1",
            field2: "value2"
          };
        }
      }
    };

    return makeExecutableSchema({ typeDefs, resolvers });
  }

  const schema = createTestSchema();
  const link = ApolloLink.from([new SchemaLink({ schema })]);
  const apolloCache = new InMemoryCache();

  const client = new ApolloClient({
    cache: apolloCache,
    link
  });

  function FooView({ id, field1, field2, bar = {} } = {}) {
    return <div>{[id, field1, field2, bar].join(" ")}</div>;
  }

  class Parent extends React.Component {
    state = { show: "A" };

    render() {
      const { show } = this.state;
      if (show === "A") {
        return (
          <React.Fragment>
            <a onClick={() => this.setState({ show: "B" })} />
            <Query
              query={gql`
                query A {
                  foo {
                    id
                    field1
                  }
                }
              `}
            >
              {({ loading, data }) =>
                loading ? null : <FooView {...data.foo} />
              }
            </Query>
          </React.Fragment>
        );
      } else if (show === "B") {
        return (
          <React.Fragment>
            <Query
              query={gql`
                query A {
                  foo {
                    id
                    field1
                  }
                }
              `}
            >
              {() => {
                return null;
              }}
            </Query>
            <Query
              fetchPolicy="cache-and-network"
              query={gql`
                query B {
                  foo {
                    id
                    field1
                    field2
                  }
                }
              `}
            >
              {({ data }) => {
                return !data.foo ? null : <FooView {...data.foo} />;
              }}
            </Query>
          </React.Fragment>
        );
      }
    }
  }

  it("should not return data.foo unless all fields are resolved from backend", async () => {
    const app = mount(
        <Parent />
      </ApolloProvider>
    );
    await waitUntil(() => {
      app.update();
      return app.find(FooView).exists();
    });

    assert.deepEqual(app.find(FooView).props(), {
      id: "id5",
      field1: "value1",
      __typename: "Foo"
    });

    app.find("a").simulate("click");

    app.update();

    await waitUntil(() => {
      app.update();
      return app.find(FooView).exists();
    });

    assert.deepEqual(app.find(FooView).props(), {
      id: "id5",
      field1: "value1",
      field2: "value2",
      __typename: "Foo"
    });
  });
});

async function waitUntil(checkFn, timeout = 2000, errorMessage) {
  const start = new Date();
  while (!checkFn()) {
    await Promise.delay(10);
    if (new Date() - start > timeout) {
      throw new Error(errorMessage || "Timeout");
    }
  }
}
```

pleunv:
I don't know why this keeps flying under the radar or if it is deemed not important enough, it's just rather frustrating that something that so drastically changes behavior (and even breaks apps) on a minor version bump gets so little attention.
Is there anything that us users can do to help move this forward?

cagdastulek:
+1

src/Query.tsx
```diff
@@ -172,7 +174,7 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
   fetchData(): Promise<ApolloQueryResult<any>> | boolean {
     if (this.props.skip) return false;
     // pull off react options
-    const { children, ssr, displayName, skip, client, ...opts } = this.props;
+    const { children, ssr, displayName, skip, client, disablePartialRefetch, ...opts } = this.props;
```
dajvido: I wonder, what is the reason to deconstruct this variable in here when you don't use it later anywhere.

danilobuerger: @dajvido To remove it from `opts`: See something like https://codeburst.io/use-es2015-object-rest-operator-to-omit-properties-38a3ecffe90

justinanastos:
🎉

mcmar:
Please let me know when a new version of `react-apollo@2.x.x` is released. I really need to get my hands on this ASAP.

hwillson:
@mcmar It will be released very soon. The plan was to release it today, but we've hit a bit of a snag with a few dependency issues (unrelated to this PR). They should be resolved shortly, and the release will be pushed out right after (current guesstimate - early am tomorrow). Thanks!

hwillson:
[React Apollo 2.2.0](https://github.com/apollographql/react-apollo/releases/tag/2.2.0) is now available. Thanks!

jonas-arkulpa:
@cagdastulek Have you enabled partial refetch?
https://github.com/apollographql/apollo-client/pull/3945/files

cagdastulek:
Yes, it didn't help either.

steelbrain:
For the people that were still having issues with this PR merged, I've opened up a followup PR that should take care of the remaining issues: https://github.com/apollographql/react-apollo/pull/2493

tnrich:
@steelbrain and others, I was thinking it might be nice if this edge case could be detected and the user could be alerted of what is going on without them having to figure it out on their own and find this PR. 

I think a short message in the console could be really helpful. Something like: 

An error occurred while trying to refetch the query XX from the cache. If you want this query to refetch in this instance please use the partialRefetch: true prop, or if you don't please explicitly pass a disablePartialRefetch: true. More info: https://github.com/apollographql/react-apollo/pull/2003 (and a link to the docs)

What do people think about that? IMO it's much better than getting the confusing no error, no data situation.

heyitsaamir:
A question related to this, when the query is being refetched, the loading flag is flipped to true for the main query, which shows a spinner in our case. Is there a recommended way of handling this? Thanks for all the work that went into this fix!

PR #2102, opened 2018-06-13
Update README.md query example

petetnt:
This PR updates the query example in the README.md from `graphql` HOC to `<Query>` component example similar to the one in the [https://www.apollographql.com/docs/react/essentials/queries.html](actual docs).


petetnt:
Updated against #1935 cc @hwillson

PR #2106, opened 2018-06-13
Issue 1812

anand-sundaram-zocdoc:
### Checklist:

This PR fixes a specific bug which occurs in the following situation:

1. A wrapped component that makes a GQL query contains multiple other wrapped components that make GQL queries
2. Multiple of the inner wrapped components throw errors

In this situation, the error returned by react-apollo's getDataFromTree will be along the lines of 'X errors were thrown when executing your fetchData functions.', and this error will have a queryErrors prop which, instead of pointing to a list of the actual errors thrown as obviously intended, points to the wrapper error 'X errors were thrown when executing your fetchData functions.' All of the actual errors that the developer wants to troubleshoot completely disappear.

This was already documented in issue 1812.

The unit test I added demonstrates the issue with the old code (if you run it with the old getDataFromTree code, the last two assertions fail). The new test passes now with the new code.

### DIFF
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -1,193 +1,217 @@
 import * as React from 'react';
 
 export interface Context {
-    [key: string]: any;
+  [key: string]: any;
 }
 
 interface PromiseTreeArgument {
-    rootElement: React.ReactNode;
-    rootContext?: Context;
+  rootElement: React.ReactNode;
+  rootContext?: Context;
 }
 interface FetchComponent extends React.Component<any> {
-    fetchData(): Promise<void>;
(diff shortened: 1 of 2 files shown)


anand-sundaram-zocdoc:
__FYI for reviewers:__ I have an ESLint plugin on my IDE which really did not like the inconsistent formatting on getDataFromTree.ts (half the functions were indented by 2 spaces, half by 4 spaces, etc.). If you select `ignore whitespace changes` in the diff options, the diff becomes a lot easier to see. Also feel free to re-format the file again with whatever opinionated linter you guys use - I selected `allow edits from maintainers.`

PR #2113, opened 2018-06-15
Fix error message in ApolloProvider.tsx

tommymorgan:
When you fail to pass a client instance to ApolloProvider the error message previously claimed you forgot to pass the client to ApolloClient.  This adjusts the message to say you forgot to pass the client to ApolloProvider.

### Checklist:


tommymorgan:
I didn't add a test since this isn't a logic change.  I feel like this is unlikely to regress and a test would be more overhead than anything else.  Happy to add one if requested though.

tommymorgan:
Now I see there is already a test which is failing.  I'll get that fixed.

PR #2142, opened 2018-06-27
Added custom cache option to MockedProvider

JakeDawkins:
Someone using MockedProvider in a project with fragments ran into an issue where they couldn't test a component, because of the default `HeuristicFragmentMatcher`. Since there was no way to control what cache was used with the original `MockedProvider`, I added a prop to it that allows a custom `Cache` to be passed in.

This also will allow people to preload caches, etc when testing components

I added an example in the tests

### Checklist:

### DIFF
--- a/src/test-utils.tsx
+++ b/src/test-utils.tsx
@@ -18,7 +18,7 @@ export class MockedProvider extends React.Component<any, any> {
     const link = new MockLink(this.props.mocks, this.props.addTypename);
     this.client = new ApolloClient({
       link,
-      cache: new Cache({ addTypename: this.props.addTypename }),
+      cache: this.props.cache || new Cache({ addTypename: this.props.addTypename }),
       defaultOptions: this.props.defaultOptions,
     });
   }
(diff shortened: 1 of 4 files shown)


src/test-utils.tsx
```diff
@@ -18,7 +18,7 @@ export class MockedProvider extends React.Component<any, any> {
     const link = new MockLink(this.props.mocks, this.props.addTypename);
     this.client = new ApolloClient({
       link,
-      cache: new Cache({ addTypename: this.props.addTypename }),
+      cache: this.props.cache || new Cache({ addTypename: this.props.addTypename }),
```
jbaxleyiii: @JakeDawkins should we warn if you have a mismatch from `addTypename` as a prop vs in the cache?

williamboman:
This seems to be a common issue (see #2254). Would be super awesome if we could merge this or #2254 soon!

hwillson:
Ooops - @JakeDawkins it looks like we've already merged this functionality in via https://github.com/apollographql/react-apollo/pull/2142. Thanks for working on this! 🙇

PR #2190, opened 2018-07-17
Call onCompleted with data from cache

### LINKED ISSUE #2177: Query.onCompleted isn't fired after fetching data from the cache
**Intended outcome:**
I tried to make a side-effect when query result is fetched using `<Query onCompleted={...}>`

**Actual outcome:**
`onCompleted` fired after the real network request, but when I changed variables and then switched them back, query result was returned from the cache and `onCompleted` handler wasn't executed.
However, it works when I change `fetchPolicy` to `cache-and-network` or `network-only`.

Make a `<Query>` that should log result into console when completed. Run it with some variables,
then change it to cause re-render, then return variables back to make it re-render result from cache.
`onCompleted` handler won't be fired.
fetchPolicy should be `cache-first`

**Versions**
  npmPackages:
    apollo-boost: ^0.1.10 => 0.1.10
    apollo-cache-inmemory: ^1.2.5 => 1.2.5
    apollo-client: ^2.3.5 => 2.3.5
    apollo-link: ^1.2.2 => 1.2.2
    apollo-link-http: ^1.5.4 => 1.5.4
    react-apollo: ^2.1.9 => 2.1.9

If it's not a bug, but a feature, can you explain how could I perform a side-effect after every query completion regardless of it uses cache or not?

olistic:
This is a first attempt to fix #2177. If there's data at the time the component mounts, `onCompleted` will be called.

I'm not sure if this is the right direction, but I didn't have much time to look at the codebase and try to understand it. If anyone could take a look and let me know what they think I'd really appreciate it. If the changes are good, I'll put some effort into fixing the tests and writing new ones.

Thanks!

### DIFF
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -197,6 +197,7 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
   componentDidMount() {
     this.hasMounted = true;
     if (this.props.skip) return;
+    this.updateCurrentData();
     this.startQuerySubscription();
     if (this.refetcherQueue) {
       const { args, resolve, reject } = this.refetcherQueue;


hwillson:
Thanks for taking an initial pass at this @olistic! I'm going to make a few adjustments to this PR, then we'll get this merged shortly.

olistic:
@hwillson Cool! If there's anything else I could do to help, it'd be a pleasure. Just let me know, and anyway, I'll keep an eye on the PR so I learn a bit more about the codebase.

kandros:
I spent like 10 hours finding the cause of a regression caused by this 😔

Apollo is not using semver?

PR #2233, opened 2018-08-02
Allow nested Query components to be populated within one tick when using MockedProvider

tobobo:
This change skips the `setTimeout` when emitting a result with the `MockedProvider` so that tests with nested queries are less unpredictable.

The test is written in the style that's recommended in the latest react-apollo guides, using [`Query` components](https://www.apollographql.com/docs/react/essentials/get-started.html#request) and [`waait` to allow assertions against the final state](https://www.apollographql.com/docs/guides/testing-react-components.html#Testing-final-state) of the component.

For more context: We have a component that uses a `Query` to load data from the server, and another `Query` within that that uses apollo-link-state to access local state. When using the `MockedProvider`, the amount of milliseconds passed to `async wait` was large and unpredictable (requiring more than 100ms to get the component to the final state). Removing this extraneous `setTimeout` allows assertions to be made on the final state using a more comforting `async wait(0);`.

### Checklist:


tobobo:
Closing this while I check out the affected tests

Issue #2253, opened 2018-08-06
errorPolicy: all in defaultOptions causes error to be undefined


adrienharnay:
**Intended outcome:**

Trying to define `errorPolicy: all` in ApolloClient `defaultOptions`.

**Actual outcome:**

`errorPolicy: all` causes `error` in `<Mutation>{(func, { error }) => <SomeComponent />}</Mutation>` to always be `undefined`.

[Screen 1](https://gyazo.com/e32f1fe358bcd206343e77d8cc1df5fc)
[Screen 2](https://gyazo.com/2615f3eca460e360190df72e5d5762b9)

```js
const client = new ApolloClient({
  defaultOptions: {
    watchQuery: {
      fetchPolicy: 'cache-and-network',
      // errorPolicy: 'all',
    },
    query: {
      fetchPolicy: 'cache-and-network',
      // errorPolicy: 'all',
    },
    // mutate: {
    //   errorPolicy: 'all',
    // },
  },
});
```

Commenting it makes errors appear again.

`<Mutation errorPolicy="all">{(func, { error }) => <SomeComponent />}</Mutation>`

Specifying it at Mutation level, and not glabally, also works.

**How to reproduce the issue:**

```js
const client = new ApolloClient({
  defaultOptions: {
    watchQuery: {
      fetchPolicy: 'cache-and-network',
      errorPolicy: 'all',
    },
    query: {
      fetchPolicy: 'cache-and-network',
      errorPolicy: 'all',
    },
    mutate: {
      errorPolicy: 'all',
    },
  },
});
```

when declaring the client, and then performing a mutation with an error.

**Version**

* apollo-client@2.3.7
* react-apollo@2.1.9

Thanks in advance!

amacleay:
I made a branch to demonstrate this issue in [react-apollo-error-template](https://github.com/amacleay/react-apollo-error-template/tree/no-error) in case it's helpful.

I do believe #2297 would resolve this issue.

adrienharnay:
Yes it seems like it did! Thanks for your help on this :+1:

PR #2297, opened 2018-08-22
Fixed Mutation error-state with errorPolicy:all

LowieHuyghe:
With errorPolicy:all Mutation should call the correct callbacks and set the correct state.

Fix for #2253: errorPolicy: all in defaultOptions causes error to be undefined


amacleay:
I would love to see this change get in - figuring out this issue has been a huge pain and I believe your change will fix it @LowieHuyghe !

Let me know if there's anything I can do to help!

hwillson:
Thanks very much for working on this @LowieHuyghe! We're going to go with the slightly different approach outlined in https://github.com/apollographql/react-apollo/pull/2374, but your contributions here helped form that PR. ❤️

PR #2298, opened 2018-08-22
fix: fetchPolicy support for Subscriptions

MatthieuLemoine:
Hi! 👋 

Apollo Client [supports](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/watchQueryOptions.ts#L150) `fetchPolicy` for subscriptions but the `fetchPolicy` option is not passed to the client in react-apollo.

It can be useful when using `no-cache` to avoid automatic cache update & handle the update manually using the new [onSubscriptionData](https://github.com/apollographql/react-apollo/pull/1966).

This PR just forwards the `fetchPolicy` option to apollo-client.

PS: It could be nice to add some kind of note next to a feature in the documentation with the minimum version number needed to use it. Actually, I tried to set `fetchPolicy` to `no-cache` because I thought that I could use `onSubscriptionData` to update the cache but it seems that it was only merged yesterday.


hwillson:
> PS: It could be nice to add some kind of note next to a feature in the documentation with the minimum version number needed to use it. Actually, I tried to set fetchPolicy to no-cache because I thought that I could use onSubscriptionData to update the cache but it seems that it was only merged yesterday.

Yes, I agree this is annoying. This is caused by the fact that our docs are auto published after a docs related PR has been merged. This can happen at any point, and isn't tied to a specific release, which is definitely confusing. Honestly I think the best way to address this (and several other repo workflow issues) is to move away from our current `master` <-> PR branch workflow, to a `master` <-> `devel` <-> PR branch workflow. We could then make sure production docs publishing only happens when `devel` merges into `master`, which would only happen when we publish a new release. More on this soon.

WaldoJeffers:
@hwillson This is a good idea, maybe the issue is less obvious for `react-apollo` maintainers, but it has happened to me a _lot_ of times (seeing something on the docs, and digging through `apollo-client/react-apollo`'s code only to find the feature has not been released yet).

PR #2304, opened 2018-08-24
Fix React.createContext in SSR

### LINKED ISSUE #2135: getDataFromTree not correctly resetting context providers
**Intended outcome:**
When using getDataFromTree, the context providers should be correctly reset to their initial value before the final React.renderToString call is executed

**Actual outcome:**
Context providers hold their last set value

https://github.com/Torsten85/apollo-getDataFromTree

this repo should log:
```
Render #1 <div>Initial: 1<div>Nested Initial: 2</div></div>
Render #2 <div>Initial: 1<div>Nested Initial: 2</div></div>
Render #3 <div>Initial: 1<div>Nested Initial: 2</div></div>
```

but actually logs:
```
Render #1 <div>Initial: 1<div>Nested Initial: 2</div></div>
Render #2 <div>Initial: 1<div>Nested Initial: 2</div></div>
Render #3 <div>Initial: 2<div>Nested Initial: 3</div></div>
```

**Version**

* apollo-client@2.3.5
* react-apollo@2.1.6

### LINKED ISSUE #2139: getDataFromTree is messing up context values
**Intended outcome:**
I'm using the new React Context API to keep a "tree path" and that value is used to choose which components will be rendered. Some of them may have graphql queries that should run on server side. That was working as intended using the old React Context API.

**Actual outcome:**
`getDataFromTree` is messing up context values and that is impacting directly on my server side rendering.

I think it is related to #2135.

https://github.com/brunoabreu/getDataFromTree-context-bug

This repo has this `Node` component that is a context consumer and provider at same time.

```js
const App = (
  <Node id='root'>
    <Node id='one'>
      <Node id='a' />
      <Node id='b' />
    </Node>
    <Node id='two'>
      <Node id='a' />
      <Node id='b' />
    </Node>
  </Node>
)
```

It works as expected with `ReactDom.renderToStaticMarkup`, but not with `getDataFromTree`:

```
without getDataFromTree:
ID: root | PARENT:
ID: one | PARENT: root
ID: a | PARENT: root/one
ID: b | PARENT: root/one
ID: two | PARENT: root
ID: a | PARENT: root/two
ID: b | PARENT: root/two

during getDataFromTree:
ID: root | PARENT:
ID: one | PARENT: root
ID: a | PARENT: root/one
ID: b | PARENT: root/one/a
ID: two | PARENT: root/one/a/b
ID: a | PARENT: root/one/a/b/two
ID: b | PARENT: root/one/a/b/two/a
```

**Version**

* apollo-client@2.3.5
* react-apollo@2.1.6

### LINKED ISSUE #2291: getDataFromTree visits React 16.3 context, but does not actually set the value in the context
**Intended outcome:**

I am storing query variables inside the React 16.3 context, and passing them to my Query component. This works as expected on the client side, but when rendering on the server side, getDataFromTree does not update the context value. 

**Actual outcome:**

The context value is never updated, the default set at the start is used everywhere

```

import { getDataFromTree } from 'react-apollo';

const TestContext = React.createContext({ test: 'defaultValueSetByHardcode' });
const TestChildren = () => {
  return (
    <TestContext.Consumer>
      {({ test }) => {
        console.log('Consume value in TestChildren', { test });
        return (<pre>{ test }</pre>);
      }}
    </TestContext.Consumer>
  );
};

const TestParent = () => {
  console.log('Rendering TestParent, providing test = valueSetByProvider ')
  return (
    <TestContext.Provider value={{ test: 'valueSetByProvider'}}>
      <TestChildren/>
    </TestContext.Provider>
  );
};

getDataFromTree(<TestParent/>, {}).then(() => {
  console.log('Getting data does not update the context');
});
```

Running that will log to the terminal the following lines:

```
Rendering TestParent, providing test = valueSetByProvider 
Consume value in TestChildren { test: 'defaultValueSetByHardcode' }
Getting data does not update the context
```

**Version**

* apollo-client@2.3.5
* react-apollo@2.1.11

mitchellhamilton:
This fixes React.createContext so that the provider value is reset to the previous value it had after its children are walked.

### Checklist:

Closes #2291
Closes #2135
Closes #2139


tgriesser:
Tried this out locally and it seems like a good start but I ran into some other issues with losing the context further down in the tree, I think I have a fix that takes care of it - will submit a little later / tomorrow.

src/getDataFromTree.ts
```diff
@@ -129,51 +133,55 @@ export function walkTree(
...
-        ((element.type as any)._context as any)._currentValue = element.props.value;
+        // this needs to clone the map because this value should only apply to children of the provider
+        newContext = new Map(newContext.entries());
+        newContext.set(element.type, element.props.value);
         child = element.props.children;
       } else {
         // A consumer
-        child = element.props.children((element.type as any)._currentValue);
+        child = element.props.children(
+          newContext.get((element.type as any).Provider) || (element.type as any)._currentValue,
```
tgriesser: I think you need to be more specific with the check here:

```ts
let currentValue = (element.type as any)._currentValue;
if (childContext.subContexts.has((element.type as any).Provider)) {
  currentValue = childContext.subContexts.get((element.type as any).Provider);
}
child = element.props.children(currentValue);
```

In theory, someone could set the value of context to a falsy value, and that would be the value that should be respected. From the React docs:

> The defaultValue argument is only used by a Consumer when it does not have a matching Provider above it in the tree. This can be helpful for testing components in isolation without wrapping them. Note: passing undefined as a Provider value does not cause Consumers to use defaultValue.

tadeuszwojcik:
Thanks @mitchellhamilton , great work. Is there a change to merge it soonish? Or it's still blocked by something?

tadeuszwojcik:
Thanks @hwillson ! Any chance to publish that fix to npm soon? Could alpha/next tag? Thanks!

hwillson:
@tadeuszwojcik I'm pushing out a new release this morning, that will include this.

tadeuszwojcik:
@hwillson that's great, thank you!

PR #2313, opened 2018-08-27
Use Partial<TData> rather than TData | {}

tgriesser:
This is related to the changes in #1983 - I think `Partial<TData>` better captures the intent of the change and plays much nicer with TypeScript, intellisense autocompletion, etc.


rosskevin:
I agree with @donataswix - use of [Partial](http://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types) makes all properties optional and is not correct. 

@hwillson we should get a patch out asap as this is a breaking change to all users who are strongly typed, I'll put up the PR now.

tgriesser:
Yeah, I think main issue here was the change to `TData | {}` in #1983 was breaking in the first place, this was just a band-aid on that from a typing perspective since `{}` is effectively an `any` in TS.

Keeping `TData | undefined` and a boolean flag as to whether data has been returned (I believe I saw this mentioned elsewhere) feels like it would be the best solution here.

PR #2322, opened 2018-08-28
Improved Typings

danilobuerger:
* Deprecated `MutationFunc` in favor of `MutationFn`.

Internally the `graphql` hoc uses the `Mutation` comp and passes the `mutate`
func along, so the typing should reflect this

* Added missing callbacks to `MutationOpts`

These callbacks get passed from the hoc to the `Mutation` comp


williamboman:
I think the return type of `MutationFn` is a bit weird now with `void` being a possible promise result value, why? This breaks our current types and we need to do a null check in every place where we consume the promise result.

hwillson:
@williamboman Any chance you could submit a PR to improve this? Thanks!

williamboman:
Sure! I'm not sure whether it's a correct assumption to make that `void` is never a valid Promise result data type for mutations though, I'll have to 🔎 the code (unless anyone have any pointers 😊)

danilobuerger:
@williamboman the return type of `MutationFn` didn't change. When using the `Mutation` comp instead of the graphql hoc (as in your example) the void check was already necessary. However, it might not be needed at all as the return type of `ApolloClient.mutate` (which is used internally) is `Promise<FetchResult<T>>` (https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L269)

PR #2356, opened 2018-09-06
[WIP] Fix Query onError by formatting currentResult

tgriesser:
This is a fix for a case where `onError` is not called on the `Query` component when there is a query with errors, because the `error` property does not exist on a raw observable `ApolloCurrentResult`.

Here is the current logic for `onError` 

https://github.com/apollographql/react-apollo/blob/6dff69506d7960793bbfd2baba7b6b1d04ef2cb3/src/Query.tsx#L341-L348

Meanwhile in render (after the `forceUpdate` is called right below this):

https://github.com/apollographql/react-apollo/blob/6dff69506d7960793bbfd2baba7b6b1d04ef2cb3/src/Query.tsx#L360-L368

The `error` prop is set by formatting the `errors` from the observable response. This PR just extracts that logic so it can be used in both locations.

I went to add tests, but the current setup for `Query` mocks incorrectly assume that this response formatting has already taken place. Will need to dig in a bit more and figure out how this should be tested.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       "integrity": "sha512-5PgPDV6F5s69XNznTcP0za3qH7qgBkr9DVQTXfZtpF+3iEyuIZB1Mjxu52F5CFxgzQUQJoBYHVxtH4Itdb5MgA==",
       "dev": true,
       "requires": {
-        "chalk": "^2.0.0",
-        "esutils": "^2.0.2",
-        "js-tokens": "^3.0.0"
+        "chalk": "2.4.1",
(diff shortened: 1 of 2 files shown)


hwillson:
Hi @tgriesser - could you please rebase and let us know if this is still an issue? The changes in https://github.com/apollographql/react-apollo/pull/2190 will impact your approach. Thanks!

hwillson:
Hi @tgriesser - are you still interested in working on this PR?

tgriesser:
Don’t have time to see if this is still an issue right now. Will reopen if I come across it in the future.

Issue #2299, opened 2018-08-23
[Feature Idea] Mutation onStart prop


thomassuckow:
**Intended outcome:**

It should be possible to use the Mutate component without arrow functions in render without necessitating making a subcomponent to get the mutation function.

**Actual outcome:**

If you disallow arrow functions in render, you end up needing a subcomponent to handle event functions if you would like to do anything other than JUST call the mutation.

**How to reproduce the issue:**

```jsx
class SaveButtonButton extends React.PureComponent {
  onClick = () => {
    this.props.onSave();
    this.props.doSave();
  };

  render() {
    return (
      <button
        onClick={this.onClick}
        disabled={this.props.disabled}
      >
        Save
      </button>
    );
  }
}

class SaveButton extends React.PureComponent {
  onSave = () => {
    this.props.onLoading(true);
  };

  onCompleted = (data) => {
    this.props.onLoading(false);
  };

  renderButton = (doSave, result) => {
    return <SaveButtonButton doSave={doSave} onSave={this.onSave} disabled={this.props.disabled} />;
  };

  render() {
    return (
      <Mutation
        mutation={SaveMutation}
        onCompleted={this.onCompleted}
        children={this.renderButton}
        variables={this.props.variables}
      />
    );
  }
}
```

**Proposed Syntax:**

I propose adding an `onStart` prop to the Mutation that would be called just before the mutation is invoked. I chose onStart because of the existing internal method `onMutationStart`

```jsx
class SaveButton extends React.PureComponent {
  onStart = () => {
    this.props.onLoading(true);
  };

  onCompleted = (data) => {
    this.props.onLoading(false);
  };

  renderButton = (doSave, result) => {
    return (
      <button
        onClick={this.doSave}
        disabled={this.props.disabled}
      >
        Save
      </button>
    );
  };

  render() {
    return (
      <Mutation
        mutation={SaveMutation}
        onStart={this.onStart}
        onCompleted={this.onCompleted}
        children={this.renderButton}
        variables={this.props.variables}
      />
    );
  }
}
```

**Version**

* apollo-client@2.4.0
* react-apollo@2.1.11

JoviDeCroock:
Would it make sense to also implement this for `Query` (i assume so).

thomassuckow:
I can't see adding it to Query hurting and for consistency it is probably better to have it. I've not used Query in ways that would necessitate it, but I can imagine cases where it could be useful if utilizing refetch/fetchmore.

[...]

PR #2407, opened 2018-09-21
[WIP] Feature/mutation on start prop

JoviDeCroock:
### Checklist:

Feature i implemented:
https://github.com/apollographql/react-apollo/issues/2299

Question: Should i also implement this for `Query`?

### DIFF
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -69,6 +69,7 @@ export interface MutationProps<TData = any, TVariables = OperationVariables> {
   ) => React.ReactNode;
   onCompleted?: (data: TData) => void;
   onError?: (error: ApolloError) => void;
+  onStart?: () => void;
   context?: Record<string, any>;
 }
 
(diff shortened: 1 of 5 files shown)


hwillson:
Thanks for working on this @JoviDeCroock! If you don't mind adding `onStart` support for `<Query />` as well, that would be awesome.

JoviDeCroock:
@hwillson i have tried this but in the test i always bounce into the problem where my thing is called 2 times, because it will be called in `updateCurrentData` and in `getQueryResult`. This is logical but moving it to `updateCurrentData` only will make calling it a bit later

thomassuckow:
Should onStart be called in startQuerySubscription before the actual subscription? Because as you said, in getQueryResult it will get called every time it processes a result instead of just when it started.

JoviDeCroock:
> Should onStart be called in startQuerySubscription before the actual subscription? Because as you said, in getQueryResult it will get called every time it processes a result instead of just when it started.

Yes but you call, `onComplete` and `onError` aswell on every processing so why would this one be any different?

thomassuckow:
True

PR #2427, opened 2018-09-27
Fix potentially undefined queryObservable options

gpoitch:
After upgrading to 2.2.0 noticed my tests failing because options was undefined here. Introduced by: 
https://github.com/apollographql/react-apollo/pull/2003


hwillson:
@gpoitch Can you give me a quick example of what your failing tests looked like? I'm interested in seeing how this occurred. Thanks!

gpoitch:
Not easily.  Essentially just rendering a component wrapped with the `graphql` HOC.  The actual issue is it's not wrapped in a whole `ApolloProvider` for the test.  I fixed it on my end by adding `options` to my mocked context, so I'll close.

PR #2332, opened 2018-08-31
Import lodash/flowRight using ES import to allow for treeshaking

Pajn:
Fixes https://github.com/apollographql/react-apollo/issues/2306

### Checklist:

* [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
* [ ] Make sure all of the significant new logic is covered by tests
* [ ] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

PR #2430, opened 2018-09-28
fix lodash typings

williamboman:
ref: #2332

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -24,4 +24,5 @@ export * from './withApollo';
 export * from './types';
 
 // XXX remove in the next breaking semver change (3.0)
-export { default as compose } from 'lodash/flowRight';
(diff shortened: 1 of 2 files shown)


disintegrator:
I'm curious why pull lodash in to get flowRight. In my mind inlining the function into react-apollo is a safer and lighter option. Am I missing something?

williamboman:
> I'm curious why pull lodash in to get flowRight. In my mind inlining the function into react-apollo is a safer and lighter option. Am I missing something?

For sure - will take a look at it

williamboman:
Actually, never mind. I'm gonna leave it as is. The comment suggests it'll be removed next major version, and ensuring feature and typings parity with Lodash's `flowRight` function is gonna be difficult.

PR #2431, opened 2018-09-28
[WIP] typings: remove void from MutationFn return type

williamboman:
ref: #2322


danilobuerger:
@williamboman this won't work as currently `runMutation` may return `Promise<void>` if there is an error and `onError` is set.

danilobuerger:
My guess (without trying it out) is that without swallowing the error, it would bubble up into react and cause an error there.

williamboman:
The typings are correct in `master` actually.. I find it somewhat strange that if an `onError` handler is passed in the configuration object then the mutation function promise will never reject - is that by design? Otherwise I'd propose making the mutation function always throw, regardless of whether an `onError` handler exists or not. Thoughts?

Issue #2466, opened 2018-10-03
[Feature Request] Create dynamicVariables for mutation MockedProvider


gabrieluizramos:
**Intended outcome:**
Use components at storybook, mocking a response for a mutation based on dynamic variables (e.g: using [HoC](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql) that updates the variables and refetch the mutation) 

**Actual outcome:**
Was throwing an error like `no mocked responses for the query`
<img width="923" alt="screen shot 2018-10-03 at 11 10 24 am" src="https://user-images.githubusercontent.com/13336905/46416024-06ecae00-c6fd-11e8-9469-9d84e6624463.png">

**How to reproduce the issue:**
Creating a component with `MockedProvider` and using dynamic values as variables, such as: 
https://codesandbox.io/s/w27mz5r44w

**Version**

- apollo-boost@0.1.16
- react-apollo@2.1.11

---------------

Maybe, creating some kind of `dynamicVariables` inside the mocks array for mockedProvider would be a nice way to help these cases, as [PR 2467](https://github.com/apollographql/react-apollo/pull/2467)

gabrieluizramos:
@hwillson Hi! Could help me with this Feature request (and the PR too), please? :)

danilobuerger:
Feature requests and non-bug related discussions are no longer managed in this repo. Feature requests should be opened in https://github.com/apollographql/apollo-feature-requests.

PR #2467, opened 2018-10-03
[Feature Request] Insert dynamicVariables at mocked provider

gabrieluizramos:
This PR updates `MockedProvider` test-link validation, inserting a new field `dynamicVariables` on the mockeds request array, such as:

```js
const mocks = [{
  request: {
    query: myQuery
  },
  result: {
    data: {
      myQuery: {
        lead: {
          pk: 1
        },
        errors: [{
          field: 'email',
          message: 'beibe beibe do beibe do biruleibe leibe'
        }],
        success: false
      }
    }
  },
  dynamicVariables: true
}];
```

Providing an environment where dynamic variables are necessary in tests and storybooks using `MockedProvider`.

As opened issue [2466](https://github.com/apollographql/react-apollo/issues/2466)

### DIFF
--- a/examples/typescript/package.json
+++ b/examples/typescript/package.json
@@ -17,7 +17,7 @@
     "@types/node": "10.11.3",
     "@types/prop-types": "15.5.6",
     "@types/react": "16.4.14",
-    "@types/react-dom": "16.0.7",
+    "@types/react-dom": "16.0.8",
     "@types/react-test-renderer": "16.0.2",
     "apollo-codegen": "0.20.2",
     "react-scripts-ts": "2.17.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -219,9 +219,9 @@
       }
     },
     "@types/react-dom": {
-      "version": "16.0.7",
-      "resolved": "https://registry.npmjs.org/@types/react-dom/-/react-dom-16.0.7.tgz",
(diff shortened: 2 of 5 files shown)


gabrieluizramos:
@hwillson Hi! Could help me with these PR and Feature request, please? :)

hwillson:
Thanks for submitting this PR @gabrieluizramos. Unfortunately, this isn't something we're able to accept at this time. The matching of exact variables that are being sent in via the request, with variables that have been defined in the mocked response, is intentional and isn't something we want to support a workaround for at this time. To achieve the same functionality you're looking for in 
https://codesandbox.io/s/w27mz5r44w, can you not just drop the variables being sent into your `this.props.mutate()` call? This will have the same affect as your proposed `dynamicVariables` change. Thanks again!

hwillson:
@gabrieluizramos We're interested in re-visiting this, so I'll re-open this PR. Instead of using the current `dynamicVariables: true` approach, we're considering allowing a `variableMatcher` function to be set, that will be a bit more specific. We'll add some changes to this PR shortly - thanks for kickstarting things!

gabrieluizramos:
@hwillson thank you for re-visiting this feature! I'm glad to help you guys 😄

gabrieluizramos:
@hwillson Hi there! Do you think that this is still interesting for you guys?

PR #2469, opened 2018-10-04
Move apollo-cache-inmemory to a regular dependency

astorije:
https://github.com/apollographql/react-apollo/blob/master/src/test-utils.tsx#L4 depends on `apollo-cache-inmemory`. When using the following in our tests:

```tsx
import { MockedProvider, MockedResponse } from 'react-apollo/test-utils';
```

The test suite fails with `Cannot find module 'apollo-cache-inmemory' from 'test-utils.js'`.

### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,6 @@
     "@types/recompose": "0.27.0",
     "@types/zen-observable": "0.8.0",
     "apollo-cache": "1.1.17",
-    "apollo-cache-inmemory": "1.2.10",
     "apollo-client": "2.4.2",
     "apollo-link": "1.2.1",
     "babel-core": "6.26.3",
@@ -146,6 +145,7 @@
     "zen-observable-ts": "0.8.10"
   },
   "dependencies": {
+    "apollo-cache-inmemory": "1.2.10",
     "fbjs": "^1.0.0",
     "hoist-non-react-statics": "^3.0.0",
     "invariant": "^2.2.2",


fbartho:
I don't believe this is a good idea. `test-utils` appears to be a utility for writing unit tests, react-apollo doesn't otherwise depend on apollo-cache-inmemory, does it?

astorije:
The problem is that, as soon as you import/require `test-utils` in your tests, assuming your `package.json` depends on `react-apollo`, then as shown above `apollo-cache-inmemory` cannot be found.

fbartho:
@astorije if you import test-utils, then you should also add `apollo-cache-inmemory` as a dependency to your project, no?

astorije:
@fbartho, I don't think we should, because it's not something the consuming app's tests depend on, it's what a dependency it uses depends on. In fact, it's not a `peerDependency` nor is it mentioned at https://www.apollographql.com/docs/guides/testing-react-components.html so I actually think it should be a regular dependency.

fbartho:
Let me rephrase my perspective: as I understand, none of the core code of this package depends on ‘inmemory; I see test utiles as a “sub package”. And react-apollo is commonly used with ‘inmemory, therefore this sub-package has tools that depend on it. Refactor proposals that make more sense to me over adding inmemory as a dep: change test-utils into a separate package, or move just the inmemory tests out, or move them into docs / code snippets.

astorije:
@fbartho, I understand what you mean, but while I agree there are solutions for longer-term, at the moment tests on a consuming project will simply break if said project does not directly depend on `apollo-cache-inmemory`, which this package does not specify.

So the immediate fix is to either make it a `peerDependency` or add it to the `dependencies` of this package. I'd argue the latter is better because it doesn't force consumers to depend on an extra package on their end if the structure ends up changing (at which point, consumers will have forgotten about it and `apollo-cache-inmemory` will live in their `package.json`, unused, forever).

Does this make sense? 🙂

rosskevin:
@fbartho is correct, test-utils are provided as part of the package to assist with testing and require the downstream project to setup devDependencies to support it.  These are not runtime resources and should never be treated as such.  The downstream project should be free to use whatever they choose in runtime, therefore adding to `dependencies` is a non-starter.

Similarly, this is not a `peerDependency`.  `peerDependency` has runtime implications and test-utils are explicitly intended for testing and to be pruned from runtime bundles.

astorije:
Alright then, thanks for your input @rosskevin!
Maybe worth adding a note at https://www.apollographql.com/docs/react/recipes/testing.html that importing `react-apollo/test-utils` requires having `apollo-cache-inmemory` as a `devDependency` then?

PR #2513, opened 2018-10-17
Remove all destructive precommit hooks.

benjamn:
If I spend the time to select specific hunks to include in a _local_ commit using `git commit -p`, it is completely unacceptable for a tool to throw that work away, e.g.
```
husky > npm run -s precommit (node v8.9.2)

�  Finding changed files since git revision f084923.
�  Found 5 changed files.
✍️  Fixing up examples/ssr/imports/app.js.
✍️  Fixing up examples/ssr/server/main.js.
✗ Found partially staged file examples/ssr/server/main.js.
✍️  Fixing up src/Query.tsx.
✗ Found partially staged file src/Query.tsx.
✍️  Fixing up src/getDataFromTree.ts.
✗ Found partially staged file src/getDataFromTree.ts.
✗ Partially staged files were fixed up. Please update stage before committing.

husky > pre-commit hook failed (add --no-verify to bypass)
```

From this point on, I will not hesitate to remove/disable any tool that installs a destructive precommit hook in any of our repositories.

Prepush hooks are fine, but only if their role is to reject the push if it fails a check. Editing commits without the developer's explicit consent is never OK.


dburles:
Hey perhaps https://github.com/nrwl/precise-commits might be a solution

PR #2514, opened 2018-10-17
Fix file name case to avoid resolution errors

mattfwood:
### Checklist:

Cloned the repo and tried to run the Mutation example, got the following error:

image

I'm guessing the person who put this together may have been developing on Linux, but because of the silly OSX file system [being case-insensitive](https://stackoverflow.com/questions/8904327/case-sensitivity-in-git), it can't resolve the `AddUser` module.


quentin-:
nice, but maybe the files should be renamed and not copied.

mattfwood:
@quentin- My mistake, should've renamed it that way. Just pushed an updated branch.

Issue #1920, opened 2018-04-10
[typescript] updateQuery in FetchMoreOptions and SubscribeToMoreOptions is not typed


alamothe:
In `GraphqlQueryControls`:
```javascript
export interface GraphqlQueryControls<TGraphQLVariables = OperationVariables> {
    error?: ApolloError;
    networkStatus: number;
    loading: boolean;
    variables: TGraphQLVariables;
    fetchMore: (fetchMoreOptions: FetchMoreQueryOptions & FetchMoreOptions) => Promise<ApolloQueryResult<any>>;
    refetch: (variables?: TGraphQLVariables) => Promise<ApolloQueryResult<any>>;
    startPolling: (pollInterval: number) => void;
    stopPolling: () => void;
    subscribeToMore: (options: SubscribeToMoreOptions) => () => void;
    updateQuery: (mapFn: (previousQueryResult: any, options: UpdateQueryOptions) => any) => void;
}
export declare type DataValue<TData, TGraphQLVariables = OperationVariables> = GraphqlQueryControls<TGraphQLVariables> & Partial<TData>;
export interface DataProps<TData, TGraphQLVariables = OperationVariables> {
    data: DataValue<TData, TGraphQLVariables>;
}
```

`GraphqlQueryControls` does not have `TData` as generic parameter. Same goes for `FetchMoreOptions` and `SubscribeToMoreOptions`. 

At least, `TData` should be passed from `GraphqlQueryControls` to `FetchMoreOptions` and `SubscribeToMoreOptions` to type `updateQuery` callback's previous result

rosskevin:
Please PR.

hwillson:
This was addressed in https://github.com/apollographql/react-apollo/pull/2525. Thanks!

PR #2525, opened 2018-10-23
Added graphql hoc missing data type generic

EricMcRay:
Fix for #1920 

### Checklist:

/label typescript


EricMcRay:
@rosskevin I aware of the order. The reason, I put `TData` as last parameter with `any`, I dont want to do any breaking change in types like you said and get the fixed types in next minor version. If you planned to merge this to major version, you can change the order to preserve consistency.

danilobuerger:
I don't think thats a great idea. It will cause more pain down the road if suddenly TData and TVariables are switched for that one case. This will cause many developer hours wasted on bug hunting as people expect the proper order. In my opinion, it should rather be a breaking change or export new types and deprecate the old ones.

danilobuerger:
But it was exported, so people could be using it. Maybe just rename them to be on the safe side? Then people who use them would catch the change immediately.

src/types.ts
```diff
@@ -47,15 +47,16 @@ export interface QueryOpts<TGraphQLVariables = OperationVariables> {
   context?: Record<string, any>;
 }
 
-export interface GraphqlQueryControls<TGraphQLVariables = OperationVariables> {
+export interface GraphqlQueryControls<TGraphQLVariables = OperationVariables, TData = any> {
```
rosskevin: 1. Alter the order to `<TData = any, TGraphQLVariables = OperationVariables>`
2. Rename `GraphqlQueryControls` to `QueryControls` as @danilobuerger mentioned.  While this is not intended to be exposed, it is public.  The best way to signal this parameterized ordering change to TS users is to change the name of the `interface`.  

It's a simple change and I am comfortable with the above strategy to be used without waiting for a major release cycle.

src/types.ts
```diff
@@ -69,7 +70,8 @@ export type MutationFunc<TData = any, TVariables = OperationVariables> = Mutatio
 >;
 
 export type DataValue<TData, TGraphQLVariables = OperationVariables> = GraphqlQueryControls<
-  TGraphQLVariables
+  TGraphQLVariables,
+  TData
```
rosskevin: Same ordering change as above.

danilobuerger:
@EricMcRay do you intend on finishing this to get it merged?

EricMcRay:
I updated typings as you talked. I extended GraphqlQueryControls from QueryControls to prevent any inconsistency in the future.

danilobuerger:
Great stuff @EricMcRay . Thanks! Just some minor changes and its ready to go.

src/types.ts
```diff
@@ -49,28 +49,33 @@ export interface QueryOpts<TGraphQLVariables = OperationVariables> {
   partialRefetch?: boolean;
 }
 
-export interface GraphqlQueryControls<TGraphQLVariables = OperationVariables> {
+export interface QueryControls<TData = any, TGraphQLVariables = OperationVariables> {
```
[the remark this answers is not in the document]
EricMcRay: I changed all usages and added TData generic for better typing but I am not sure it will break something. I don't have good knowledge about all code base. Can you check and confirm changes?

hwillson:
It looks like we're all set here - @danilobuerger @rosskevin would you agree?

Issue #615, opened 2017-04-10
getDataFromTree throws errors when any query fails


choonkending:
### Background

I asked a question on StackOverflow today [here](http://stackoverflow.com/questions/43316336/how-should-we-handle-an-error-while-getting-initial-state-from-store-in-getdataf). 

I thunk about it harder afterwards and decided that this is a discussion worthy of a Git issue.

### Problem

_*Any*_ error gets thrown in [getDataFromTree](https://github.com/apollographql/react-apollo/blob/fb7065912178c67e35c8c3a765ab69473af67d63/src/server.ts#L141-L148). 

### Expected/Proposed Behaviour

Return the data with an `error` Array, without throwing an error.

***Why?***

Our GraphQL server speaks to multiple APIs and returns a graphQL response which is handled by Apollo UI.

Certain GraphQL errors are **not considered critical**, and we want to still render a page with it. E.g.

```js
data: {
   thing: {
       id: "1234"
       name: "Thing"
       nonCriticalData: null
   },
   error: {
       graphQLErrors: [
           path: ['thing', 'nonCriticalData']
       ]
   }
```

With the above response we get from apollo, we want to still render a page with `thing`.

I have followed https://github.com/apollographql/react-apollo/pull/488#issuecomment-284756892 and got to this point.

In my mind the errors should be returned and not thrown unless necessary, because not all errors are equal. 

Ideally I would like to do the following

```js
getDataFromTree(app)
    .then(() => {
             const content = ReactDOMServer.renderToString(app);
             const { data } = client.getInitialState();   <------ Data (with errors)
             // check for expected errors and render
             return renderPage(data);
     })
    .catch(error => {
         // handle unexpected errors
    });
```

### Version
- apollo-client@1.0.0-rc2
- react-apollo@1.0.1

P.S. Thanks for an awesome library. Happy to help out if needed.

helfer:
[...]

PR #2533, opened 2018-10-26
Reimplement getDataFromTree using ReactDOM.renderToStaticMarkup.

benjamn:
The `getDataFromTree` function is used during server-side rendering (and sometimes also on the client) to await all `fetchQuery` promises generated by `Query` components in a given tree of React components, so that subsequent `ReactDOM.renderToString` and/or `renderToStaticMarkup` calls can render the tree synchronously, without worrying about fetching data.

Until now, `getDataFromTree` was implemented by walking the component tree, instantiating components along the way, fetching any data for `Query` components, calling `render()` methods, and recursively processing children. This simulated rendering process was similar to the official rendering process performed by React, but it was not exactly the same, so it created a nontrivial maintenance burden for `react-apollo` to accommodate any new React features.

The recent introduction of [Hooks](https://reactjs.org/docs/hooks-intro.html) is just one such feature, but hooks are particularly troublesome for `getDataFromTree`, because they depend intimately on internal React state that cannot be manipulated using any stable, public API. In other words, updating our tree-walking implementation of `getDataFromTree` to support hooks would have been very difficult, if not impossible. Even if it was possible, it would never be truly future-proof, since there could always be other new React features that threatened doom for `getDataFromTree`.

In general, the only way to make `getDataFromTree` truly future-proof is to implement it using the official React rendering functions, and/or wait for React to implement direct support for asynchronous server-side rendering (which would be AMAZING).

For now, this PR implements `getDataFromTree` by calling `renderToStaticMarkup` repeatedly, collecting any `Query` data promises and awaiting them between renderings. The number of renderings should never be greater than the length of the longest chain of nested `Query` components (plus one), and the result of the final rendering is a `Promise<string>` whose HTML value can be used immediately, instead of calling `renderToString` again.

This approach is worse in some ways than our previous approach, since parts of the tree get re-rendered multiple times, and HTML is generated unnecessarily. Also, exceptions thrown during component rendering that are not related to data fetching can now abort the entire rendering, whereas previously we trapped errors within their specific subtrees. This was possible because we controlled the entire "rendering" process. Now that we're putting our faith in `ReactDOM`, we just have to roll with the punches, as far as exceptions are concerned. But that seems acceptable in practice.

**On the upside, we have a fully hooks-compatible implementation of `getDataFromTree` on the same day hooks were announced!**


test/client/getDataFromTree.test.tsx
```diff
@@ -849,10 +856,11 @@ describe('SSR', () => {
...
 
-      return getDataFromTree(app).catch(e => {
-        expect(e.toString()).toEqual(expect.stringContaining('2 errors were thrown'));
-        expect(e.queryErrors.length).toBeGreaterThan(1);
-        expect(e.toString()).not.toEqual(e.queryErrors[0].toString());
+      return getDataFromTree(app).then(() => {
+        throw new Error('Should have thrown an error');
+      }, e => {
+        expect(e.toString()).toEqual('Error: foo');
+        expect(e).toBe(fooError);
```
benjamn: This test was previously enforcing the functionality where we tried to trap multiple exceptions within their subtrees. Needless to say, this test is less useful now that only one exception will be thrown.

test/client/getDataFromTree.test.tsx
```diff
@@ -505,8 +505,15 @@ describe('SSR', () => {
...
-      return getDataFromTree(app).then(() => {
+      await getDataFromTree(app).then(html => {
+        const markup = ReactDOM.renderToStaticMarkup(app);
+        expect(markup).toEqual(html);
+        expect(markup).toMatch(/James/);
+      });
+
+      await getDataFromTree(app, ReactDOM.renderToString).then(html => {
         const markup = ReactDOM.renderToString(app);
+        expect(markup).toEqual(html);
```
benjamn: If you want to use `ReactDOM.renderToString` instead of `renderToStaticMarkup`, you can. The benefit here is that you can just call `getDataFromTree` once, without having to call `renderToString` afterwards.

benjamn:
For the [`examples/ssr`](https://github.com/apollographql/react-apollo/tree/master/examples/ssr) application, here's how I instrumented the SSR code:
```js
const start = +new Date;
await getDataFromTree(WrappedApp);
const body = renderToString(WrappedApp);
console.log("server rendering took", new Date - start, "ms");
sink.renderIntoElementById('app', body);
```

Here are the timings for ten requests using the previous tree-walking implementation of `getDataFromTree`:
```bash
~/dev/react-apollo/examples/ssr% meteor
[[[[[ ~/dev/react-apollo/examples/ssr ]]]]]

=> Started proxy.
=> Started MongoDB.
=> Started your app.

=> App running at: http://localhost:3000/
I20181026-14:17:07.631(-4)? server rendering took 146 ms
I20181026-14:17:13.668(-4)? server rendering took 23 ms
I20181026-14:17:14.432(-4)? server rendering took 20 ms
I20181026-14:17:15.321(-4)? server rendering took 21 ms
I20181026-14:17:16.261(-4)? server rendering took 18 ms
I20181026-14:17:17.130(-4)? server rendering took 29 ms
I20181026-14:17:18.036(-4)? server rendering took 23 ms
I20181026-14:17:18.982(-4)? server rendering took 17 ms
I20181026-14:17:19.902(-4)? server rendering took 17 ms
I20181026-14:17:20.807(-4)? server rendering took 17 ms
```

Here are the timings for ten requests using the new implementation of `getDataFromTree`:
```bash
I20181026-14:19:07.723(-4)? server rendering took 202 ms
I20181026-14:19:11.579(-4)? server rendering took 27 ms
I20181026-14:19:12.565(-4)? server rendering took 17 ms
I20181026-14:19:13.579(-4)? server rendering took 23 ms
I20181026-14:19:14.593(-4)? server rendering took 21 ms
I20181026-14:19:15.623(-4)? server rendering took 20 ms
I20181026-14:19:16.639(-4)? server rendering took 15 ms
I20181026-14:19:17.605(-4)? server rendering took 11 ms
I20181026-14:19:18.578(-4)? server rendering took 18 ms
I20181026-14:19:19.567(-4)? server rendering took 20 ms
```

If I only use `renderToStaticMarkup`, and just render the final result of `getDataFromTree` (instead of calling `renderToString` again), it gets a little better:
```bash
I20181026-14:19:52.365(-4)? server rendering took 80 ms
I20181026-14:19:53.485(-4)? server rendering took 15 ms
I20181026-14:19:54.436(-4)? server rendering took 16 ms
I20181026-14:19:55.293(-4)? server rendering took 17 ms
I20181026-14:19:56.162(-4)? server rendering took 19 ms
I20181026-14:19:57.004(-4)? server rendering took 26 ms
I20181026-14:19:57.790(-4)? server rendering took 15 ms
I20181026-14:19:58.636(-4)? server rendering took 15 ms
I20181026-14:19:59.481(-4)? server rendering took 20 ms
I20181026-14:20:00.210(-4)? server rendering took 15 ms
```

In other words, there's no apparent performance regression for small component trees. Here's the entire server-rendered HTML string for this app (to quantify "small"):
```html
<div><div><div><h3>R2-D2</h3><h6>Luke Skywalker: newhope, empire, jedi</h6><h6>Han Solo: newhope, empire, jedi</h6><h6>Leia Organa: newhope, empire, jedi</h6></div></div></div>
```

My hunch is that these rendering times are dominated by the `fetchData` promises, not the rendering itself, but that's not surprising since the component tree is so small.

benjamn:
Data confirming that `fetchData` (which communicates with the local GraphQL server) takes roughly 80% the time here:
```text
91 'ms to resolve' { data: 
   { hero: 
      { name: 'R2-D2',
        id: '2001',
        friends: [Array],
        __typename: 'Droid' } },
  loading: false,
  networkStatus: 7,
  stale: false }
server rendering took 112 ms

27 'ms to resolve' { ... } 
server rendering took 33 ms

25 'ms to resolve' { ... } 
server rendering took 29 ms

19 'ms to resolve' { ... } 
server rendering took 20 ms

21 'ms to resolve' { ... } 
server rendering took 22 ms

28 'ms to resolve' { ... } 
server rendering took 32 ms

24 'ms to resolve' { ... } 
server rendering took 26 ms

28 'ms to resolve' { ... } 
server rendering took 30 ms

21 'ms to resolve' { ... } 
server rendering took 24 ms

23 'ms to resolve' { ... } 
server rendering took 26 ms

19 'ms to resolve' { ... } 
server rendering took 22 ms
```

Here's how I instrumented the code:
```ts
if (!this.queryGraveyard.has(query, JSON.stringify(variables))) {
  const start = Date.now();
  this.queryPromises.set(
    queryInstance,
    new Promise(resolve => {
      resolve(queryInstance.fetchData());
    }).then(result => {
      console.log(Date.now() - start, "ms to resolve", result);
      return result;
    })
  );
  ...
```
Note that every request gets a new `ApolloClient` and `InMemoryCache` instance, so the extra time taken by the first request is not because of caching. I guess there's just some additional initialization overhead for the first time calling `renderToStaticMarkup`, but it gets amortized once the server is up and running.

tizmagik:
This is great, I’m especially glad to hear about moving off a parallel implementation of walking the tree and instead relying on the public React API.

However, I’m definitely concerned about the performance impact this would have, since renderToStaticMarkup() is synchronous, I expect a non-trivial impact on the performance for large apps (NYTimes). I can try and pull this branch down and run it against our app to see how it performance and post back, but in case it is severely impacted, do we have a plan B?

tizmagik:
If you have any specific test procedures/parameters you’d like me to use to benchmark, please share and I’d be happy to run them. Otherwise will try and run some `ab` runs and compare the numbers.

tizmagik:
> For now, this PR implements `getDataFromTree` by calling `renderToStaticMarkup` repeatedly, collecting any `Query` data promises and awaiting them between renderings. The number of renderings should never be greater than the length of the longest chain of nested `Query` components (plus one), and the result of the final rendering is a `Promise<string>` whose HTML value can be used immediately, instead of calling `renderToString` again.

@benjamn -- is it known when this *last* render is? If so, would it be possible to have this last render optionally do a `renderToString()` (instead of a `renderToStaticMarkup()`)? This is because if an app is calling `hydrate()` on the client, the markup needs to be from `renderToString()`, so this would save an additional render in userland for some apps.

mike-marcacci:
This is a very solid strategy, and is almost identical to how I implemented our SSR for Relay (before moving over to Apollo). It's a far less brittle solution than walking the tree, and is worth whatever performance overhead.

As I'm sure we all know, the react team is still plugging away on suspense and async rendering, which will eventually add first-class support for this kind of thing. In the meantime, I think the priority should be providing a way to handle SSR with Apollo which doesn't break on each new React release. Using `renderToStaticMarkup` to populate the store is a really great way to do this.

tmeasday:
@benjamn keep in mind that it isn't just rendering latency that is the problem; even if unnecessary CPU load (rendering strings) is only contributing a small percentage to the latency, it can seriously affect throughput by multiplying per-request CPU usage significantly.

Having said that, I think it's the right choice!

Plus I'm generally an advocate of not layering your GraphQL queries too deep anyway -- if you have only one or two queries per page this isn't going to be a big performance hit for you, right?

benjamn:
> Plus I'm generally an advocate of not layering your GraphQL queries too deep anyway -- if you have only one or two queries per page this isn't going to be a big performance hit for you, right?

@tmeasday That's right! If you don't have any `Query` components that are descendants of other `Query` components, then this approach will only call the rendering function twice (once to collect the single layer of `Query` components, and once to render with the cached data). And if you use `renderToString` instead of `renderToStaticMarkup`, and reuse the final HTML string instead of calling `renderToString` again, then you'll be doing *less* work than before:
```js
// This HTML should be renderable as-is:
const html = await getDataFromTree(root, renderToString);
// No need to call renderToString(root) here!
```

benjamn:
In case anyone has been itching to try this out, I just published `react-apollo@2.3.0-beta.0` with these changes included.

benjamn:
> is it known when this *last* render is? If so, would it be possible to have this last render optionally do a `renderToString()` (instead of a `renderToStaticMarkup()`)?

@tizmagik I see your point, but unfortunately I don't think it's possible to detect when the last rendering is going to happen, since it's now up to `ReactDOM` to traverse the component tree to find any remaining `Query` components. Even the first rendering could be the last, if no `Query` components are found.

I think the benefit of getting your final HTML from `await getDataFromTree(root, renderToString)` should outweigh the additional cost of using `renderToString` for rendering, especially if you make an effort to flatten your `Query` component hierarchy. In other words, if calling `renderToString` twice is really more expensive than calling `renderToStaticMarkup` twice followed by calling `renderToString` once, then `renderToString` must be more than twice as slow as `renderToStaticMarkup`, which would be surprising.

tizmagik:
@benjamn instead of rendering to string, would it make sense to try using [react-reconciler](https://www.npmjs.com/package/react-reconciler) to basically write a `renderToMemory()` renderer to avoid the CPU-blocking step of producing the HTML markup that's thrown away on each successive render down the tree?

Admittedly, such an implementation would be more effort than this PR, but should hopefully be easier to maintain/support compared to the custom walker that's in `master`, as the heavy lifting/implementation details are mostly left up to `react-reconciler`. Thoughts?

src/getDataFromTree.ts
```diff
@@ -1,275 +1,133 @@
...
 export default function getDataFromTree(
   rootElement: React.ReactNode,
-  rootContext: any = {},
-): Promise<any> {
-  const errors: any[] = [];
-  const storeError = (error: any) => errors.push(error);
+  // The rendering function is configurable! We use renderToStaticMarkup as
+  // the default, because it's a little less expensive than renderToString,
+  // and legacy usage of getDataFromTree ignores the return value anyway.
+  renderFunction = renderToStaticMarkup,
```
benjamn: As a side note, it would be really awesome if dynamic `import()` was safe for libraries to use already, since it would enable us to call `import("react-dom/server")` here only if no `renderFunction` is provided, so `react-dom/server` and its dependencies would not be forced into the initial JS bundle. Since `getDataFromTree` is already async, a one-time dynamic `import()` would be pretty much invisible.

benjamn:
Though it would still be great to discuss possible performance improvements in a follow-up issue, I've merged this PR and released `react-apollo@2.3.0` with the `next` tag (not `latest` yet), because
1. this implementation solves a lot of existing problems with `getDataFromTree`,
2. I'm not aware of any other equally future-proof options, and
3. performance optimization is largely in the hands of application developers:
    - minimize `Query` component nesting
    - move `Query` components closer to the root of the tree
    - reuse `await getDataFromTree(root, renderToString)` instead of calling `renderToString(root)` again

tizmagik:
Thanks @benhjames -- agreed this is a good approach for now, and the fact that the renderer is configurable means one could implement a sort of `renderToMemory()` using `react-reconciler` instead of the `renderToString()` approach.

I might look into doing just that, would you be open to a PR?

src/getDataFromTree.ts
evenchange4: ##### Notes:
There is a breaking change of `getDataFromTree` in `react-apollo^2.3.0`. 
The second argument `rootContext` is replaced with `renderFunction`.

benjamn: Good point!

A few thoughts:
1. That's why it's a minor version bump.
2. I will have to think about whether `RenderPromisesProvider` can still provide a root context now that we don't have complete control over how contexts are managed.
3. Did you have any use cases for `rootContext`?

benjamn:
@tizmagik I think I would like to see `react-reconciler` maintained a little more actively with respect to development of `react` and `react-dom`. Right now it seems to lag behind the other packages. That said, it would be great to see how much of a difference it makes!

src/getDataFromTree.ts
evenchange4: Here is an example how I use `react-apollo v2.2.4` in next.js:
 
```js
await getDataFromTree(
  <ApolloProvider client={apollo}>
    <Component />
  </ApolloProvider>,
  { router }
)
```

ref: https://github.com/zeit/next.js/blob/canary/examples/with-apollo-and-redux/lib/withApollo.js#L42-L53

My workaround for `react-apollo v2.3.0`:
```diff
await getDataFromTree(
  <ApolloProvider client={apollo}>
+   <RouterProvider router={router}>
      <Component />
+   </RouterProvider>
  </ApolloProvider>,
- { router }
)
```

RouterProvider: https://github.com/zeit/next.js/issues/2908#issuecomment-333107793

examples/ssr/server/main.js
```diff
@@ -27,10 +27,11 @@ export const render = async sink => {
     </ApolloProvider>
   );
 
+  const start = +new Date;
```
steelbrain: This seems to already be merged but It's generally a good idea to just do `Date.now()` instead of creating a Date object and then getting it's value like here and in the few lines below

src/getDataFromTree.ts
benjamn: @evenchange4 Does this PR resolve your concerns? https://github.com/apollographql/react-apollo/pull/2586

evenchange4: Great. Thank for your awesome work!

klarstrup:
For anyone else having piggybacked on old getDataFromTree's fetchData behavior for non-apollo things(I was using it to await an async Redux thunk), it's because this PR and thus 2.3.0+ removed it.

I know this was probably never intended to be used like that, but uhhhhh a lil note in the changelog would've saved me half a day of debugging. 😅

OllieJennings:
@benhjames I know this has been merged, but can we please get some support with this issue (re-opening it could be a start): https://github.com/apollographql/react-apollo/issues/615 its been around for almost 2yrs with nobody looking into it.

I know the Apollo guys have a lot of things going on with new features etc, but i just don't think SSR should be forgotten, and solving this issue will give Apollo the whole package, making it seamless to add SSR to an Apollo app, since all errors would be handled the exact same way.

hwillson:
@OllieJennings I've re-opened https://github.com/apollographql/react-apollo/issues/615. Thanks for the heads up!

(and quick FYI - you @ mentioned the wrong Ben - easy to do since `benhjames` often comes up first in the autocomplete list, but you wanted @benjamn)

OllieJennings:
@hwillson cheers, yeah they look so similar, my bad there.

Issue #2539, opened 2018-10-26
React Hooks support


vovaguguiev:
I'd like to start a discussion about the possible next-gen API based on [react hooks](https://reactjs.org/docs/hooks-intro.html) for `react-apollo`.
I think this API brings a perfect opportunity for `react-apollo` to improve already excellent DX even more.

```js
// isn't this beatiful?
function Avatar() {
  const { loading, error, data } = useQuery(`
    {
       me {
          initial
       }
    }
  `);

  if (loading) return <div>Loading...<div>;

  if (error) return <div>Error happened</div>

  return (
    <div>{data.me.initial}<div>
  );
} 
```

What do you guys think?

wzrdzl:
Looks like the first step to make it happen is to start using the react 16.3 context API, I opened #2540 to address this

kristianmandrup:
Yes, the `Mutation` and `Query` tags should also be available as `useMutation` and `useQuery` :)
Let's do it!

kristianmandrup:
I started a very basic attempt at hooks here: https://github.com/kristianmandrup/react-apollo/tree/hooks

Feel free to continue this effort. So far extracted Query and Mutation as classes so they are not React Components any longer but can be used as standalone classes.

wzrdzl:
@lifeiscontent no idea, would love to hear from the maintainers. it makes sense to close one of the issues to keep the discussion in one place, but I don't really know which place is the right one

kristianmandrup:
I now have ~80% of the grunt work done [here](https://github.com/kristianmandrup/react-apollo/tree/hooks). Should be able to get it fully working and all tests passing, with ~2-3 hrs more work... Would love to see someone fork it and continue. Over and out here: midnight in London.

hyperparsa:
Integration with Suspense will also be nice.

albertorestifo:
@hyperparsa Suspense is not yet ready to be used with data fetching.

trojanowski:
I created sample hooks to use with Apollo Client: <https://github.com/trojanowski/react-apollo-hooks>.

FredyC:
Even though Suspense is not ready for data fetching, isn't it like double work to do it with hooks first and then rewrite to Suspense? I mean if you look at [this video from React Conf](https://youtu.be/ByBPyMBTzM0?t=1415) (time included) it's kinda clear that any data fetching won't be just a simple hook. Andrew even mentions Apollo there and that it should probably wait for a proper implementation.

[...]

PR #2540, opened 2018-10-26
Refactor ApolloProvider/ApolloConsumer implementation to use React 16.3 context API

### LINKED ISSUE #1937: Support React.createContext
I am the creator of [Loadable Components](https://github.com/smooth-code/loadable-components/). I would like to thank you for this awesome library. I inspired from your code, especially the [getDataFromTree](https://github.com/apollographql/react-apollo/blob/master/src/getDataFromTree.ts).

I found that React.createContext is not supported. I patched it in https://github.com/smooth-code/loadable-components/pull/65. You can take the patch, it works.

But I think it would be nice to create a separated package to walk a tree in React, a lot of other libraries rely on it.

vovaguguiev:
The old way of passing the context still exists side-by-side with the new way to allow gradual migration of `<Query>`, `<Mutation>` and other modules that access the context in the "old" way.
So this change is completely backwards compatible

This should unlock experiments with React Hooks version of `react-apollo` API (#2539) because if you want to access the context(client) in a hook, the context has to be exposed using the "new" context API

closes #1937


package.json
```diff
@@ -143,11 +143,13 @@
     "zen-observable-ts": "0.8.10"
   },
   "dependencies": {
+    "@types/memoize-one": "^3.1.1",
```
janek26: Should be a dev dep I guess

janek26:
CI for Preact will fail until [developit/preact#963](https://github.com/developit/preact/pull/963) is resolved

package.json
wzrdzl: good catch! fixed

wzrdzl:
@jbaxleyiii what shall we do about Preact? Wait until [developit/preact/pull/963](https://github.com/developit/preact/pull/963) is merged or add a some handling for the case when createContext is not a function?

package.json
janek26: no problem, thanks fot the work you‘re doin‘ for open source 🙂

FredyC:
Would love to see this merged as well. Right now it's basically necessary to have a custom Provider in the tree if I want to consume the ApolloClient eg. by hooks.

I agree that waiting for Preact is not feasible in a long run with all that sweetness of Suspense and Hooks, it's unlikely they will be able to catch up. Perhaps `apollo-preact` could be made out of the current master and let it go its own path?

prateekrastogi:
maybe creating a different branch and publishing the alpha-version of this package supporting react-hooks and suspense will satisfy all the stakeholders, and consequently save the time in the main release when preact supports new context api? We can't wait every time for two upstream repos especially when other projects including react actively publish alpha packages.

arianon:
Why not just use the [create-react-context](https://github.com/jamiebuilds/create-react-context) polyfill? It works well and the preact tests pass when using it.

FredyC:
@arianon That could be also a way, but only if it can be tree shaken well enough. It's not a good idea to bloat build just to support some React clone :)

arianon:
@FredyC It's small enough (1.78kB min+gzip) to keep react-apollo under its current `bundlesize` target (11.5kB), so I wouldn't worry about that.

prateekrastogi:
Maybe bloating on alpha version will be an acceptable tradeoff? 😬

FredyC:
> Maybe bloating on alpha version will be an acceptable tradeoff? 😬

To me, an acceptable tradeoff is not to support Preact in alpha version if they do not care about it (assuming by 4 months old issue).

prateekrastogi:
@FredyC i agree with you. But, i guess it's more of business issue. As big enterprise client might be using preact in production and supporting them is essential priority for Apollo. Apollo revenue strategy seems to be focused on high traffic enterprise client. But, sacrificing DX which brings enterprise customers is somewhat awkward. Although, i am a fan of and use Apollo server and react Apollo for the quality DX that Apollo stack provides. But, may be sending this feature request from our company email ids' can be helpful 🤐

benjamn:
As for Preact, perhaps there's a way for this code to detect if Preact is being used (e.g. `React.createContext` is not defined), and instead behave like it did before, without returning `ApolloContext.Provider` or `ApolloContext.Consumer`?

Autarc:
Is there an update about the current state of the PR @wzrdzl regarding @benjamn concerns to avoid the additional dependency? Otherwise I would like to take a look to ensure context migration works in [StrictMode](https://reactjs.org/docs/strict-mode.html#detecting-legacy-context-api) 😄

src/ApolloConsumer.tsx
```diff
@@ -1,20 +1,28 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import ApolloClient from 'apollo-client';
+import ApolloContext from './ApolloContext';
 
 const invariant = require('invariant');
```
sandiiarov: Looks like you can remove it.

hwillson:
I agree with @benjamn regarding the removal of memoization, and until Preact X is launched (which does support `createContext`), I think leveraging the `create-react-context` polyfill is a good workaround (since it only adds about 50 bytes to the bundle, after gzip).

These changes have been made, so we should be close to getting this merged. Let me know if anyone objects - thanks!

src/ApolloProvider.tsx
```diff
@@ -49,6 +50,10 @@ export default class ApolloProvider<TCache> extends Component<ApolloProviderProp
   }
 
   render() {
-    return this.props.children;
+    return (
+      <ApolloContext.Provider value={this.getChildContext()}>
```
joshjg: Should also remove `static childContextTypes` from this component.

hwillson: @joshjg I left this intentionally. This PR isn't intended to be a full blown replacement of the old context handling approach, as it's still being used by React Apollo internally. There are more changes required to be able to remove `childContextTypes` completely (which will be coming soon, in `react-apollo` 3.0).

src/ApolloContext.ts
```diff
@@ -0,0 +1,10 @@
+import createReactContext from 'create-react-context';
+import ApolloClient from 'apollo-client';
+import { DocumentNode } from 'graphql';
+
+export interface ApolloContextValue {
+  client?: ApolloClient<Object>;
+  operations?: Map<string, { query: DocumentNode; variables: any }>;
+}
+
+export default createReactContext<ApolloContextValue | undefined>(undefined);
```
benjamn: Thanks for addressing my previous feedback @hwillson!

Given our concern with bundle sizes (#2743), it seems undesirable to add a new dependency to polyfill `React.createContext`, which has been available unpolyfilled [since React 16.3](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#official-context-api).

If we detect that `React.createContext` is missing here, I think we should print/throw an error recommending either updating to React 16.3+ or manually polyfilling it with `create-react-context`, but we should not do anything that results in adding `create-react-context` to the bundle for developers who are up-to-date with the latest React version.

hwillson: Thanks @benjamn - I was hoping to avoid a major version bump, so the polyfill seemed like a good alternative (given the small size and that we could remove it in `react-apollo` 3). Happy to adjust though - thanks!

benjamn: Is there any way we could fall back to the old behavior if `React.createContext` is not available?

benjamn: I think I found a way to do it. I'll push a commit soon.

src/ApolloContext.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ApolloClient from 'apollo-client';
+import { DocumentNode } from 'graphql';
+
+export interface ApolloContextValue {
+  client?: ApolloClient<Object>;
+  operations?: Map<string, { query: DocumentNode; variables: any }>;
+}
+
+export const ApolloContext = React.createContext
```
benjamn: I manually ran the tests with `false && React.createContext` here instead, and they still pass. Any thoughts on a good way to run tests automatically both ways?

hwillson: As long as the tests are passing, we should be golden with your approach @benjamn (since the Preact tests passing prove that the legacy API fallback is working). This is great - thanks!

mattfwood:
Wouldn’t we need to also export ApolloContext in `index.ts`? If you try to useContext with a consumer or provider, it doesn’t work properly

FredyC:
@mattfwood The `useApolloClient` (or similar) should be exposed instead imo.

leebenson:
Just a heads-up: after bumping to 2.5.3 which uses this, I started to get console errors about `fbjs/lib/warning` being missing.

This seems to have come from `create-react-context`, which is odd since `fbjs` is listed as a dep (although pegged to 0.8.0); `npm i fbjs` (1.0.0) fixed it.

Just an FYI for anyone else stuck on this.

MrLoh:
@hwillson since #2872 will only be part of 3.0, couldn't the ApolloContext just be exposed directly, then everyone who want's this hook can use it by doing `useContext(ApolloContext)` but there is no requirement for anything else?

notgiorgi:
> @hwillson since #2872 will only be part of 3.0, couldn't the ApolloContext just be exposed directly, then everyone who want's this hook can use it by doing `useContext(ApolloContext)` but there is no requirement for anything else?

As it's not gonna be exposed, here's the workaround I came up with: https://github.com/apollographql/react-apollo/pull/2872#issuecomment-483612213

PR #2543, opened 2018-10-27
Extend QueryProps from QueryOpts

nikita-graf:
`QueryOpts` interface is currently missing a property `partialRefetch` which is present in `QueryProps`. It prevents using `partialRefetch` in a `query` HOC. Rather than adding this property to `QueryOpts` I'm using extending. Hope it'll help to prevent such a diff between two interfaces in the future.

### Checklist:


src/types.ts
```diff
@@ -42,9 +42,10 @@ export interface QueryOpts<TGraphQLVariables = OperationVariables> {
   fetchPolicy?: FetchPolicy;
   errorPolicy?: ErrorPolicy;
   pollInterval?: number;
-  client?: ApolloClient<any>;
+  client?: ApolloClient<Object>;
```
rosskevin: I don't believe `ApolloClient<Object>` to be correct.  What is the rationale for changing it?

nikita-graf: As far as `QueryProps` interface defines `client` as [`ApolloClient<Object>`](https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L95) I preserved it within my changes. I agree that `any` would be a better option for both interfaces.

rosskevin: In this file you have changed it so I would change it back.  I'm pretty certain `Object` is incorrect.

nikita-graf: Reverted back `any` type.

rosskevin:
Thanks @nikita-graf

PR #2570, opened 2018-11-06
Pass through fetchPolicy on Mutation

danilobuerger:
Supported by apollo-client, so should be supported by react-apollo too.


danilobuerger:
@rosskevin @hwillson @benjamn ping

PR #2573, opened 2018-11-07
Expose graphql hocs to allow choosing them explicitly and improve typing

### LINKED ISSUE #2568: Typescript graphql() hoc issue with option.prop and ChildDataProps typings
**Intended outcome:**

The documentation TS examples should work.

**Actual outcome:**

The documentation TS examples do not compile due to typings related to the "prop" option and ChildDataProp

https://www.apollographql.com/docs/react/recipes/static-typing.html#props

Create a project with the example here: https://www.apollographql.com/docs/react/recipes/static-typing.html#props

**Version**

- apollo-client@latest
- react-apollo@latest

danilobuerger:
Fixes #2568

* Allows user to explicitly choose the hoc they want to use instead of relying on the parsed document type.
* Enabled typescript user to profit from tighter typings by not using Partials for the child props.


RIP21:
@helfer @hwillson anyone can take a look at this one?

RIP21:
@benjamn @steelbrain @helfer @hwillson  anyone once again? :) 
We need it, It's simple as hell, but will be amazing to have for TS.
Also, it will require a bit of a change to docs eventually. But right now it's a mess to work with TS and HOCs. 
Please, please, please.

RIP21:
@benjamn @steelbrain @helfer @hwillson @helfer 
I really sorry for being an annoying ass. But this is SO needed.

src/mutation-hoc.tsx
```diff
@@ -12,11 +12,11 @@ import {
   GraphQLBase,
 } from './hoc-utils';
 
-export function mutation<
+export function withMutation<
```
rosskevin: FYI - This is a breaking change and relies on a semver bump.

rosskevin: My bad, we use rollup and this wasn't included before.

danilobuerger: Yeah, this has never been exported by the actual react-apollo module. Also was never documented as public api.

rosskevin:
Thanks @danilobuerger

PR #2586, opened 2018-11-12
Restore getDataFromTree(rootElement, rootContext) API.

benjamn:
In `react-apollo@2.3.0`, the `rootContext` parameter was replaced by the custom `renderFunction` parameter, which was a breaking change: https://github.com/apollographql/react-apollo/pull/2533#discussion_r230251846

This commit restores the previous API, while providing an alternative API that is more flexible and more accurately named: `getMarkupFromTree`

If you want to specify a custom rendering function, you should now use `getMarkupFromTree` instead of `getDataFromTree`:
```js
import { getMarkupFromTree } from "react-apollo";
import { renderToString } from "react-dom/server";

getMarkupFromTree({
  tree: <App/>,
  // optional; defaults to {}
  context: { ... },
  // optional; defaults to renderToStaticMarkup
  renderFunction: renderToString,
}).then(markup => {
  // Use the markup returned by the final rendering...
});
```
Hopefully, since `react-apollo@2.3.0` was never assigned the `latest` tag on npm, we can push this change without another minor version bump (likely 2.3.1).

cc @evenchange4 @hwillson


csantos1113:
Hi @benjamn - thanks for your hard work on the SSR support. Are you publishing this version soon to `npm` registry?

benjamn:
@csantos1113 It's published now as version 2.3.1, though it's still using the `next` tag instead of `latest`.

csantos1113:
@benjamn - thank you!

PR #2593, opened 2018-11-16
Fix problems with importing react-dom/server in React Native.

### LINKED ISSUE #2592: react-apollo-browser.umd.js requiring react-dom/server on react native
**Intended outcome:**

When using react-apollo with a react-native app (specifically an expo app), the app should be able to successfully build and run

**Actual outcome:**

Expo app is unable to build, throwing this error:
`Unable to resolve "react-dom/server" from "node_modules\react-apollo\react-apollo.browser.umd.js"`

So it seems like the newest version(s) of react-apollo is incompatible with react native due to react-apollo.browser.umd.js

image

1. Create a new expo app
2. npm install react-apollo
3. Import ApolloProvider from react-apollo
4. Run expo start and open app on expo client on device

Issue reproduced on expo snacks

App throwing react-dom/server error:
https://snack.expo.io/@lijtong/apollo-new-umd

App without react-dom/server error:
https://snack.expo.io/@lijtong/apollo-old-umd

It seems to be these lines specifically (version 2.3.1):
`    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('react-dom/server'), require('apollo-client'), require('prop-types'), require('react')) :`
`    typeof define === 'function' && define.amd ? define(['exports', 'react-dom/server', 'apollo-client', 'prop-types', 'react'], factory) :`

Whereas the version that works (2.1.11) has:
`    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('react'), require('prop-types'), require('apollo-client')) : `
`    typeof define === 'function' && define.amd ? define(['exports', 'react', 'prop-types', 'apollo-client'], factory) :`

**Version**

Working versions
- apollo-client@<2.3.8>
- react-apollo@<2.1.11>

Incorrect UMD versions
- apollo-client@<2.3.8>
- react-apollo@<2.3.1>

benjamn:
After #2533, `getDataFromTree` is now implemented in terms of `renderToStaticMarkup`, which comes from `react-dom/server`, but the presence of the `react-dom` dependency was not enforced, as was revealed by #2592.

If `react-apollo` depends directly on `react-dom`, and tries to be even a little bit strict about the version, we run the risk of duplicating the entire `react-dom` package, which is often one of the largest npm dependencies in React-based apps.

A peer dependency won't automatically solve #2592, but it will provide an important clue to the solution, if my comment https://github.com/apollographql/react-apollo/issues/2592#issuecomment-439531481 is accurate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-apollo",
-  "version": "2.3.2-beta.1",
+  "version": "2.3.2-beta.2",
   "author": "opensource@apollographql.com",
   "private": true,
   "browser": "lib/react-apollo.browser.umd.js",
--- a/package.json
+++ b/package.json
@@ -12,6 +12,9 @@
   ],
   "license": "MIT",
   "main": "lib/react-apollo.umd.js",
+  "react-native": {
+    "./lib/react-apollo.umd.js": "./lib/react-apollo.umd.native.js"
+  },
   "module": "src/index.ts",
   "typings": "lib/index.d.ts",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-apollo",
-  "version": "2.3.1",
+  "version": "2.3.2-beta.0",
   "author": "opensource@apollographql.com",
   "private": true,
   "browser": "lib/react-apollo.browser.umd.js",
--- a/package.json
+++ b/package.json
@@ -127,7 +127,6 @@
     "preact-compat": "3.18.4",
     "prettier": "1.15.2",
     "react": "16.5.2",
-    "react-dom": "16.6.1",
     "react-test-renderer": "16.6.1",
     "recompose": "0.30.0",
     "recursive-rename": "2.0.0",
@@ -149,6 +148,7 @@
     "invariant": "^2.2.2",
     "lodash.flowright": "^3.5.0",
     "lodash.isequal": "^4.5.0",
-    "prop-types": "^15.6.0"
+    "prop-types": "^15.6.0",
+    "react-dom": "16.6.1"
   }
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -149,6 +148,7 @@
     "invariant": "^2.2.2",
     "lodash.flowright": "^3.5.0",
     "lodash.isequal": "^4.5.0",
-    "prop-types": "^15.6.0"
+    "prop-types": "^15.6.0",
+    "react-dom": "16.6.1"
```
benjamn: Note that tests only pass if this is an exact version. If I relax the constraint with `^` or explicitly update to the latest version (16.6.3), tests begin to fail.

package.json
```diff
@@ -88,7 +88,8 @@
   },
   "peerDependencies": {
     "apollo-client": "^2.3.8",
-    "react": "0.14.x || 15.* || ^15.0.0 || ^16.0.0",
```
benjamn: @hwillson @jbaxleyiii Is it reasonable to stop advertising support for `react@0.14.x` at this point? That version was last published two years ago (0.14.9). Also, this is just a peer dependency, not actually a guarantee that `react-apollo` works with `react@0.14.x`, which we probably don't anymore…

hwillson: @benjamn Quite reasonable!

test/client/getDataFromTree.test.tsx
```diff
@@ -186,7 +186,7 @@ describe('SSR', () => {
         expect(elementCount).toEqual(2);
       });
 
-      it('function stateless components with React 16.3 context', () => {
+      xit('function stateless components with React 16.3 context', () => {
```
hwillson: Should we just remove these?

benjamn:
Closing in favor of https://github.com/apollographql/react-apollo/pull/2627.

PR #2627, opened 2018-11-27
Remove top-level react-dom/server import to fix #2592.

### LINKED ISSUE #2592: react-apollo-browser.umd.js requiring react-dom/server on react native

benjamn:
Supersedes #2593.
Should fix #2592.

By pushing the `react-dom/server` import down into the relevant functions that need it, we can avoid unconditionally importing that dependency tree, which helps in environments like React Native where `react-dom/server` either doesn't work or seems undesirable (see discussion in #2592).

Since the React Native bundler will still try to traverse the `require("react-dom/server")` dependencies, it's important to prune that dependency with a
```json
"react-native": {
  "react-dom/server": false
}
```
section in `react-apollo/package.json`. Note that this does not prevent React Native apps from using `getMarkupFromTree` with an appropriate `renderFunction`; it simply prevents React Native's bundler from bundling the `react-dom/server` dependency just because `react-apollo` is imported.

Tested with both react-native@0.55.4 and @0.57.7 (Expo SDKs 30 and 31).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,21 @@
 
 ## vNext
 
+## 2.3.2
+
+### Improvements
+
+### Bug Fixes
+
+- This package no longer imports `react-dom/server` unconditionally at the
+  top level, making `react-apollo` safer to use in environments like React
+  Native that are neither browser-like nor Node-like, and thus struggle to
+  import `react-dom/server` and its dependencies. Additionally, the React
+  Native bundler has been instructed to ignore all `react-dom/server`
+  dependencies within `react-apollo`, so `react-dom` will not be bundled
+  in React Native apps simply because they import `react-apollo`.
+  [PR #2627](https://github.com/apollographql/react-apollo/pull/2627)
+
 ## 2.3.1 (November 15, 2018)
 
 ### Improvements
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-apollo",
-  "version": "2.3.1",
+  "version": "2.3.2-beta.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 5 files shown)


davetownsend:
Thanks for this.  Fixes my issue.

Using RN v0.57.3 (no expo)

treyp:
This also fixes the issue for me using RN 0.57.1 (no Expo). Thanks!

Issue #1811, opened 2018-03-19
Skip not working as expected


YannisMarios:
**How to reproduce the issue:**

Hi I am using the same component/form to create and edit a company in my app.

When I edit an existing company the code below correctly prints the id of the company:
```
graphql(getCompany,
	{ options: (props) => {
		console.log(props.location.state.id)
		 return {
			variables: { id: props.location.state.id },
			skip : !props.location.state
	}}}
 )
```

When I create a new company I want to skip the getCompany query because obviously the id is undefined. But skip is not working as expected and the query is not skipped and an error is thrown:

TypeError: Cannot read property 'id' of undefined

Any ideas?

I have also tried:

```
export default compose(
	graphql(getCompany,
		{ skip: (props) => !props.location.state },
		{ options: (props) => ( { variables: { id: props.location.state.id } } ) }
	)
)(withApollo(CreateCompany));

```
query: 

```
	query getCompany($id: ID!) {
		getCompany(id: $id) {
			id
			index
			enabled
			name
			email
			address
			zip
			phone
			notes
			users {
				id,
				index,
				name,
				surname,
				email
			}
			area {
				id
				index
				name
				country {
					id
					name
				}
			}
			tax_office {
				id
				index
				name
			}
			activity {
				id
				index
				name
			}
			contacts {
				id
				index
				job_title {
					id
				}
				title {
					id
				}
				name
				office_phone
				office_fax
				mobile
				emergency_phone
				email
			}
			trading_name
			tax_code
			created_at
		}
	}

```

**Version**

* apollo-client@2.2.7
* react-apollo@2.1.0-beta.3

YannisMarios:
This finally worked for me:

```
export default compose(
	graphql(getCompany,
		{
			skip: (props) => !props.location.state,
			options: (props) => ( { variables: { id: props.location.state.id } } )
		}
	)
)(withApollo(CreateCompany));

```

galvanu:
Looks like this changes related to the newer version of apollo, we had the same issue after apollo client upgrade. After moving 'skip' out of 'options' object all works.

7yp0:
Looks like the issue here ist that since 2.2.0 `data` is set to `undefined` when skip property is set: https://github.com/apollographql/react-apollo/pull/1916/files?utf8=%E2%9C%93&diff=unified#diff-45a733ae4e94396f8fa0947302418680R366

I created a pull request for this: https://github.com/apollographql/react-apollo/pull/2628

[...]

PR #2628, opened 2018-11-28
return data when Query Component skip prop is set

7yp0:
Related to this: https://github.com/apollographql/react-apollo/issues/1811


danilobuerger:
This is working as expected. `data` is either `TData` or `undefined`. Since we are skipping, `data` is set to undefined. 

See also comment in source:
```
    // When skipping a query (ie. we're not querying for data but still want
    // to render children), make sure the `data` is cleared out and
    // `loading` is set to `false` (since we aren't loading anything).
```

danilobuerger:
@7yp0 You could use `fetchPolicy` for that. For example: set it to cache first (which is the default anyway) and if you want to update call a refetch until you don't anymore.

7yp0:
I tried that already, but sadly it won't work for my special case, since the caching is not working for this query (long story... the issue here is probably that we failed at chunking the query properly. Instead we have one huge query, which fucks up caching of course)

I had hoped to solve (work around) this with the skip property but as you mentioned it is not supposed to be as I expected.

Thanks for your clarification.

danilobuerger:
Best bet is to ask on Slack about it, I bet someone can help you.

PR #2659, opened 2018-12-13
Chore/reduce bundle size

JoviDeCroock:
### Checklist:

I attempted to reduce the bundlesize by removing dependencies, I personally prefer making several small PR's for this so we can handle problem by problem.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -28,4 +28,4 @@ export * from './types';
 
 export * from './walkTree';
 
-export { default as compose } from './utils/flowRight';
+export { compose } from './utils/flowRight';
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -28,6 +28,4 @@ export * from './types';
 
 export * from './walkTree';
 
-// XXX remove in the next breaking semver change (3.0)
-const compose = require('lodash.flowright');
-export { compose };
+export { default as compose } from './utils/flowRight';
--- a/src/utils/flowRight.ts
+++ b/src/utils/flowRight.ts
@@ -2,12 +2,13 @@ export default(...funcs: any[]) => {
   const functions = funcs.reverse();
   functions.forEach((fnc, i) => {
     if (typeof fnc !== 'function') {
-      throw new Error(`flowRight at index ${i} is not a function`);
+      throw new Error(`In "compose" the argument at index "${i}" is not a function.`);
     }
   });
-  return function (...args: [any]) {
-    let result = functions[0].apply(null, args);
-    functions.forEach((fnc) => {
+  return function (...args: any[]) {
+    const [firstFunction, ...restFunctions] = functions
+    let result = firstFunction.apply(null, args);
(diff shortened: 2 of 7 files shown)


src/utils/flowRight.ts
```diff
@@ -0,0 +1,16 @@
+export default(...funcs: any[]) => {
```
benjamn:
```suggestion
export function compose(...funcs: any[]) {
```

src/browser.ts
```diff
@@ -28,6 +28,4 @@ export * from './types';
 
 export * from './walkTree';
 
-// XXX remove in the next breaking semver change (3.0)
-const compose = require('lodash.flowright');
-export { compose };
+export { default as compose } from './utils/flowRight';
```
benjamn:
```suggestion
export { compose } from './utils/flowRight';
```

src/utils/flowRight.ts
```diff
@@ -0,0 +1,16 @@
...
+  const functions = funcs.reverse();
+  functions.forEach((fnc, i) => {
+    if (typeof fnc !== 'function') {
+      throw new Error(`In "compose" the argument at index "${i}" is not a function.`);
+    }
+  });
+  return function (...args: any[]) {
+    const [firstFunction, ...restFunctions] = functions
+    let result = firstFunction.apply(null, args);
+    (restFunctions || []).forEach((fnc) => {
```
benjamn:
```suggestion
    restFunctions.forEach((fnc) => {
```
I believe `restFunctions` is guaranteed to be an array here by the `...restFunctions` syntax above.

src/utils/flowRight.ts
```diff
@@ -0,0 +1,16 @@
+export default(...funcs: any[]) => {
+  const functions = funcs.reverse();
+  functions.forEach((fnc, i) => {
+    if (typeof fnc !== 'function') {
+      throw new Error(`In "compose" the argument at index "${i}" is not a function.`);
```
benjamn: I don't think we need this extra dynamic type checking, since the functions will fail when we call them below, and TypeScript consumers will be warned about passing non-`Function` arguments (if you accept the change I suggested above).

JoviDeCroock: Okay, i'll fix it :)

src/browser.ts
benjamn: On second thought, how about just inlining the `flowRight.ts` module into this module?

JoviDeCroock:
Thank you so much for the quick response, let's make this wonderfull OSS project even better.

benjamn:
Just as a heads up, I'm planning to open a PR here that mirrors https://github.com/apollographql/apollo-client/pull/4234, so I will merge this PR into that branch once I've pushed it. We can publish beta versions to npm, but I'd like to keep track of how much bundle size we've saved, in one long-running PR.

JoviDeCroock:
Okay, i'll probably be posting more of these, my plan was to start at react-apollo and then move on to others.

PR #2661, opened 2018-12-13
[WIP] Reduce bundle size.

benjamn:
Part of issue #2743, whose description previously appeared here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+.rpt2_cache
 
 # Logs
 logs
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,23 @@
 
 ### Improvements
 
+- The `walkTree` function has been deprecated, since there's no way to
+  make its behavior consistent with the latest versions of React. To save
+  bundle size, `walkTree` is no longer exported from `react-apollo`,
+  though you can still access it as follows:
+  ```js
+  import { walkTree } from "react-apollo/walkTree"
+  ```
+
+## 2.4.0
+
+### Bug Fixes
+
+- Invoke `onCompleted`/`onError` even if `Mutation` unmounts. <br/>
+  [PR #2710](https://github.com/apollographql/react-apollo/pull/2710)
+
+### Improvements
+
 - Update the typescript example app to use the raw Query component directly,
   with generics, to avoid generating the extra object that's created (in the
   compiled code) when extending the Query component as a class.  <br/>
--- a/jest.browser.umd.config.js
+++ b/jest.browser.umd.config.js
@@ -3,6 +3,7 @@ const { jest } = require('./package.json');
 jest.moduleNameMapper = {
   '\\.\\./src$': '<rootDir>/lib/react-apollo.browser.umd.js',
   '\\.\\./src/test-utils': '<rootDir>/lib/test-utils.js',
+  '\\.\\./src/walkTree': '<rootDir>/lib/walkTree.js',
   // Force other imports to /src/whatever to fail
   '\\.\\./src': '<rootDir>/test/fail-no-entry-point.js',
 };
--- a/jest.server.umd.config.js
+++ b/jest.server.umd.config.js
@@ -3,6 +3,7 @@ const { jest } = require('./package.json');
 jest.moduleNameMapper = {
   '\\.\\./src$': '<rootDir>/lib/react-apollo.umd.js',
   '\\.\\./src/test-utils': '<rootDir>/lib/test-utils.js',
+  '\\.\\./src/walkTree': '<rootDir>/lib/walkTree.js',
   // Force other imports to /src/whatever to fail
   '\\.\\./src': '<rootDir>/test/fail-no-entry-point.js',
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -324,6 +324,49 @@
         "json-schema-traverse": "^0.3.0"
       }
     },
+    "ansi-align": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/ansi-align/-/ansi-align-3.0.0.tgz",
+      "integrity": "sha512-ZpClVKqXN3RGBmKibdfWzqCY4lnjEuoNzU5T0oEFpfd/z5qJHVarukridD4juLO2FXMiwUQxr9WqQtaYa8XRYw==",
+      "dev": true,
+      "requires": {
+        "string-width": "^3.0.0"
+      },
+      "dependencies": {
+        "ansi-regex": {
+          "version": "4.0.0",
+          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-4.0.0.tgz",
+          "integrity": "sha512-iB5Dda8t/UqpPI/IjsejXu5jOGDrzn41wJyljwPH65VCIbk6+1BzFIMJGFwTNrYXT1CrD+B4l19U7awiQ8rk7w==",
+          "dev": true
+        },
+        "is-fullwidth-code-point": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-2.0.0.tgz",
+          "integrity": "sha1-o7MKXE8ZkYMWeqq5O+764937ZU8=",
+          "dev": true
+        },
+        "string-width": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/string-width/-/string-width-3.0.0.tgz",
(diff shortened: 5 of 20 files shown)


rosskevin:
> Reimplementing surprisingly heavy dependencies locally, a la 4bac03a.

@benjamn as noted in the code removed, `lodash.flowright` is slated to and should be removed altogether.  It is not part of this package and is re-exported.  I would _definitely_ rather not see us copy something into this project that should not have been here to begin with.  Is there an argument for keeping/supporting it from here instead of dropping it?

> Transforming imports to be more specific using a compiler plugin, so that import { SpecificThing } from "react-apollo" becomes import { SpecificThing } from "react-apollo/lib/specific-thing". If we end up pursuing this approach for apollo-client, we will no doubt reuse the same tools here.

This should be unnecessary with tree shaking.  Aside from the conveniences of browser vs server indexes, file specific paths should be a thing of the past and we should be able to confidently use an index to generate the rollup bundle and have dead code eliminated in the application bundling phase.

> Introducing a production bundle in addition to the development bundle, with more aggressive elimination of unnecessary code (e.g. error strings).

👍 

> Moving away from monolithic bundles and towards individual modules.

I'm not sure this makes sense.

----

The key distinction here: let's not conflate a rollup bundle size produced with the size of the library in an application bundled with a modern tool like webpack or parcel that eliminates dead code.

While perhaps there is an argument for different entry points for server vs browser, I do not understand going further down this path when it should have a zero impact on app bundling over the current approach. With that said, things like prod vs dev bundles and #2677 can definitely help.

rosskevin:
WOW - my error.  We are only producing umd for consumption.

We need to configure our rollup to be producing an es module to aid in dead code elimination.  Some examples:

- https://github.com/final-form/react-final-form/blob/master/rollup.config.js
- https://github.com/dmtrKovalenko/material-ui-pickers/blob/develop/lib/rollup.config.js

We definitely need to be shipping an es/esm variant otherwise we are stuck with a non-shakeable umd.

JoviDeCroock:
> WOW - my error. We are only producing umd for consumption - which appears to be the root cause of this.
> 
> We need to configure our rollup to be producing an es module to aid in dead code elimination. Some examples:
> 
> * https://github.com/final-form/react-final-form/blob/master/rollup.config.js
> * https://github.com/dmtrKovalenko/material-ui-pickers/blob/develop/lib/rollup.config.js
> 
> We definitely need to be shipping an es/esm variant otherwise we are stuck with a non-shakeable (or poorly shakeable depending on the tool) umd.

I can add this in my PR / make a seperate one for that if needed?

rosskevin:
@JoviDeCroock definitely separate.

JoviDeCroock:
Afaik the tsc compiled bundle suffices for ESM, ESM is just ecmascript module syntax. Some packages give an .mjs there now but that isn't doable in frontend.

Maybe it should be es6 target in tsconfig that's still a question i'm playing with.

rosskevin:
> Right, I would have thought it was sufficient that the module field of package.json points to lib/index.js, which is ESM? Isn't that what Rollup cares about?

My error once again, I saw the lack of target in the `rollup.config` as well as `"module": "src/index.ts"`.  I just verified https://unpkg.com/react-apollo@2.3.3/package.json and in my local app build that we are tree shaking - no sign of the umd.  But with that said, the `tsc` is producing `es5` instead of `es2015`.  So we aren't actually providing an `es` `module`, but a transpiled down es5 `src`.

----

So this issue is about optimizing bundle size - is there a reason to focus on the umd bundle size?  Is anyone focused on bundle size in their app using a umd build?  If so why?  perhaps I'm too far down my own path to understand what the problem we are attempting to solve.

One thing I am sure of - we need a `js` file next to a `.d.ts` in the artifact to satisfy downstream usage of ts definitions.  There currently (as of two months ago) was no rollup for .d.ts files.

I would think we could provide `src` as we do currently, but change the `module` to point to a rollup-produced `es2015` target.  This would cover dev workflow and optimal bundling (I think, unless I'm missing something.  Seems like we just touched on this in material-ui-pickers referenced above).

JoviDeCroock:
I have been giving this some thought and for the `pkg.json` wouldn't it be more normal to make the es5 tsc compiled index.js the main and a new rollup generated ES2015 bundle the module. This in turn allows for better tree shaking. I'll make a POC PR this afternoon/evening when I find some time

rosskevin:
Agreed @JoviDeCroock, just like I PR'd in apollo-client https://github.com/apollographql/apollo-client/pull/4261:

1. `main` exploded cjs files and .d.ts
2. `module` esm bundle
3. `browser` umd bundle

The umd bundle is legacy and helps with putting your finger on the pulse of bundlesize (though not a foolproof indicator as we have discussed). 

Users concerned with tree shaking should ensure they are using the targeted `module` bundle.

JoviDeCroock:
I was wondering, now we are creating a seperate bundle for the umd libraries, can't this be solved with the `rollup 1.0` feature to have multiple entry points?

rosskevin:
> I was wondering, now we are creating a seperate bundle for the umd libraries, can't this be solved with the rollup 1.0 feature to have multiple entry points?

Looks like there could be a few different ways to tackle this, one would be using the same index inputs and have multiple outputs (just makes the config more DRY):

```
{
  input: 'main-b.js',
  output: [
    {
      file: 'dist/bundle-b1.js',
      format: 'cjs'
    },
    {
      file: 'dist/bundle-b2.js',
      format: 'esm'
    }
  ]
}
```

Or the multiple inputs with pattern named outputs:

```
input
Type: string | string [] | { [entryName: string]: string }
CLI: -i/--input <filename>

The bundle's entry point(s) (e.g. your main.js or app.js or index.js). If you provide an array of entry points or an object mapping names to entry points, they will be bundled to separate output chunks. Unless the output.file option is used, generated chunk names will follow the output.entryFileNames option. When using the object form, the [name] portion of the file name will be the name of the object property while for the array form, it will be the file name of the entry point.

This will generate at least two entry chunks with the names index-a.js and index-b.js:

// rollup.config.js
export default {
  ...,
  input: {
    a: 'src/main-a.js',
    b: 'src/main-b.js'
  },
  output: {
    ...,
    entryFileNames: 'entry-[name].js'
  }
};
```

Definitely plenty of good/new options there.

Pajn:
Is it possible to try out this PR, eg. is it published anywhere?

It would be interesting to see if the esm bundle would allow some tree shaking.

JoviDeCroock:
> Is it possible to try out this PR, eg. is it published anywhere?
> 
> It would be interesting to see if the esm bundle would allow some tree shaking.

I'd also love to see this being tried out. Been reading some things about webpack being not that good at treeshaking bundles but leaving that to the tersers dead code elimination. If we can get it tested before we do a real publish we could still alter the course for that matter.

JoviDeCroock:
> According to BundlePhobia, version `2.5.0-bundle-size.0` is now [25.4KB minified and 7.4KB after gzip](https://bundlephobia.com/result?p=react-apollo@2.5.0-bundle-size.0), compared to [32.7KB minified and 8.7KB after gzip](https://bundlephobia.com/result?p=react-apollo@2.4.0) for the latest official version, 2.4.0. Since the gzip size is what really matters here, that's a 15% improvement! 🎉

Why does it indicate 2.4.0 as tree-shakeable, the UMD build provided there should not be tree shakeable as far as I know

benjamn:
> Why does it indicate 2.4.0 as tree-shakeable

I wonder if that just means there's an ESM `module` field in `package.json`?

JoviDeCroock:
Does anyone have an SSR application using react-apollo. To test the new bundle-size build? Because I don't know how node reacts to the new pkg.json entries.

JoviDeCroock:
Something that comes in my mind when thinking about reducing bundlesize is maybe implementing our render props/hoc/... with a hook. That way we get the advantage of exposing a hook AND our components can be reduced to only functions.

This reduces the transpiling overhead of classes needing to be added.

Just throwing it out there because well I can see as to why you could be opposed to this idea (supporting more react versions)

JoviDeCroock:
> why is the module field in package.json pointing to a ts-file? IMHO rollup should produce ES modules and this field should point there.

This is the case in master but not in 2.5.0-bundle-size.0
https://github.com/apollographql/react-apollo/pull/2661/files#diff-f764c456bc424cc70f25f27247ccf338R23

OneCyrus:
ah ok i see. but the tsconfig.json still transpiles to target:es5 which doesn't produce ES modules

JoviDeCroock:
> ah ok i see. but the tsconfig.json still transpiles to target:es5 which doesn't produce ES modules

It does, it literally sais "es", it is just a transpiled down version of es6. No library transpiles to es6 in their module field. This would imply very long builds since your node modules should be transpiled to support all browsers

PR #2662, opened 2018-12-13
[WIP]: Chore/new context

JoviDeCroock:
I was working on integrating the new context API into React-Apollo because the propTypes dependency could be worked away like that, when I kind off hit a road block that needs a bit of an architectural insight.

The problem persists in the `getDataFromTree` here we give a tree that possibly has a Provider in it, this makes our `renderPromises` obsolete since that child Provider will never implement it.

I've written a `// TODO` where the problem persists, making this PR since a second mind would be a big help.

The second problem being preact having a seperate package for context.

I also don't really understand why this fails on the `commonjs`, `browser` build these tests should succeed.

Please note that this is different from the other PR since this actually makes `react-apollo` work with StrictMode in terms of context API since it is not using `getChildContext` anymore. This could be the first step (of the two steps needed) to be fully compatible.

The issue is rather vague to me as to why it works on node but not on browser, if someone is more up to date don't be afraid to contact me I'll gladly solve it.

Thanks for reading

### DIFF
--- a/src/getDataFromTree.tsx
+++ b/src/getDataFromTree.tsx
@@ -105,6 +105,7 @@ export type GetMarkupFromTreeOptions = {
 
 export function getMarkupFromTree({
   tree,
+  context = {},
   // The rendering function is configurable! We use renderToStaticMarkup as
   // the default, because it's a little less expensive than renderToString,
   // and legacy usage of getDataFromTree ignores the return value anyway.
@@ -124,7 +125,7 @@ export function getMarkupFromTree({
           {(value: any) => (
             // TODO: this is where the issue lives, we provide a tree
             // that starts with an ApolloContext.Provider effectively overriding this one.
-            <ApolloContext.Provider value={{ ...value, renderPromises }}>
+            <ApolloContext.Provider value={{ ...context, ...value, renderPromises }}>
               {tree}
             </ApolloContext.Provider>
           )}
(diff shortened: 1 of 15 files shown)


src/getDataFromTree.tsx
```diff
@@ -104,37 +105,34 @@ export type GetMarkupFromTreeOptions = {
...
       // Always re-render from the rootElement, even though it might seem
       // better to render the children of the component responsible for the
       // promise, because it is not possible to reconstruct the full context
       // of the original rendering (including all unknown context provider
       // elements) for a subtree of the orginal component tree.
-      return tree;
+      return (
+        <ApolloContext.Consumer>
+          {(value: any) => (
+            // TODO: this is where the issue lives, we provide a tree
```
JoviDeCroock: This is what i'm talking about. When tree starts with <ApolloProvider client={client}> it will kind off fail. 

This could be solved by adjusting the `ApolloProvider` analogue to this one, first consume a possibly existing context and then override the possibly existing context

JoviDeCroock: I implemented this, making it work on node env's but browser is still an issue.

JoviDeCroock:
For preact, we could use this: https://github.com/valotas/preact-context

I was wondering if this polyfill was desired for react: https://github.com/jamiebuilds/create-react-context

The one thing my mind can't handle is that it's working on node envs but not in Browser, Commonjs and server.

PR #2669, opened 2018-12-16
chore: provide replacement for lodash.isequal

JoviDeCroock:
### Checklist:


JoviDeCroock:
The two failing tests aren't impacted by the change so I don't really undertand what's going on there.

package.json
```diff
@@ -91,8 +91,7 @@
   },
   "peerDependencies": {
     "apollo-client": "^2.3.8",
-    "react": "^15.0.0 || ^16.0.0",
-    "react-dom": "^15.0.0 || ^16.0.0",
+    "react": "0.14.x || 15.* || ^15.0.0 || ^16.0.0",
```
benjamn: Careful, I think you might be branching off an older version of `master`, since this seems to be undoing recent changes.

benjamn:
As tempting as this is, I worry that the new implementation does not behave the same way as the previous `lodash.isequal` implementation—yes it's big, but that's because equality testing in JavaScript has a lot of corner cases, and we can't just ignore those cases. For example, the new code doesn't do anything to defend against circular structures, whereas lodash does.

JoviDeCroock:
> As tempting as this is, I worry that the new implementation does not behave the same way as the previous `lodash.isequal` implementation—yes it's big, but that's because equality testing in JavaScript has a lot of corner cases, and we can't just ignore those cases. For example, the new code doesn't do anything to defend against circular structures, whereas lodash does.

Allright, well then i'll have to find some way of only bundling this in the test-utils class since 1.55KB is pretty significant imo :-) will see what I can do

package.json
JoviDeCroock: Oh, will check that. Sometimes it's pretty weird keeping up with an upstream sorry bout that

PR #2677, opened 2018-12-18
chore: provide esm bundle

JoviDeCroock:
So I optimised the build process so PURE_CALLS etc are annotated,.... This in turn shaves off 3KB.
I've seen this babel-jest issue before and I'm searching on it, if anyone passes here and knows more about it feel free to correct it on my branch.

I'll gladly make work of this in `apollo-client` aswell if this one is according to your wishes.

Some other things i'll be looking into after this is probably if closure compiler is a viable idea


package.json
```diff
@@ -113,8 +110,10 @@
     "apollo-cache-inmemory": "1.2.10",
     "apollo-client": "2.4.7",
     "apollo-link": "1.2.1",
-    "babel-core": "6.26.3",
+    "babel-core": "7.0.0-bridge.0",
     "babel-jest": "23.6.0",
+    "babel-plugin-annotate-pure-calls": "0.4.0",
+    "babel-plugin-dev-expression": "0.2.1",
```
rosskevin: `babel-plugin-dev-expression` doesn't appear to be used in this project, why add it?

rollup.config.js
```diff
@@ -25,7 +26,14 @@ function umd(inputFile, outputFile) {
       node({
         module: true,
         only: ['tslib']
-      })
+      }),
+      babel({
+        exclude: 'node_modules/**',
+        plugins: [
+          'annotate-pure-calls',
+          'dev-expression',
```
rosskevin: Why plugins here and presets in `babel.config.js`?  I think it best to just put the config in one place.  This is split.

rollup.config.js
```diff
@@ -56,7 +64,17 @@ export default [
...
     },
     plugins: [
-      node(),
+      node({
+        module: true,
+        only: ['tslib']
+      }),
+      babel({
+        exclude: 'node_modules/**',
+        plugins: [
```
rosskevin: Same comment as above, split babel config doesn't make sense to me.

rollup.config.js
JoviDeCroock: Thes are bundling plugins and not literal babel plugins

package.json
JoviDeCroock: It's used in rollup as a bundling posttranspilation step afaik

JoviDeCroock: It is used in the babelConfig to cut out the invariant messages in prod mode

benjamn:
My concern about this PR is that it seems to be optimizing the way third-party npm packages are bundled into the `dist/bundlesize.js` bundle, but not actually reducing the compressed size of `lib/react-apollo.umd.js` or `lib/react-apollo.browser.umd.js`. In other words, the techniques you're using here might be a good recommendation for developers who are bundling an application that uses `react-apollo` (along with other packages), but I'm not sure we can really claim we've reduced the footprint of `react-apollo` itself. Am I understanding this correctly?

package.json
benjamn: Those `invariant` messages are all found inside the various React dependencies, though, right?

babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(api) {
+  api.cache(true);
+  return {
+    plugins: [
+      'annotate-pure-calls',
+      'dev-expression',
```
rosskevin: This appears to be unused/unnecessary in react-apollo, or am I missing something?

JoviDeCroock:
> My concern about this PR is that it seems to be optimizing the way third-party npm packages are bundled into the `dist/bundlesize.js` bundle, but not actually reducing the compressed size of `lib/react-apollo.umd.js` or `lib/react-apollo.browser.umd.js`. In other words, the techniques you're using here might be a good recommendation for developers who are bundling an application that uses `react-apollo` (along with other packages), but I'm not sure we can really claim we've reduced the footprint of `react-apollo` itself. Am I understanding this correctly?

Well in production now the invariant messages are cut away and while bundling pure functions won't be repeated. Allthough what you are stating is partially true it's also safe to assume this reduces the react-apollo bundle size.

Feel free to counter this point. Because I can't really see why this is not improving this package.

babel.config.js
benjamn: I think you're right. There might be an opportunity to make use of `__DEV__` within `react-apollo`, but right now this only helps because some of `react-apollo`'s (React) dependencies use the `__DEV__` expression, and the `dist/bundlesize.js` file generated by the `bundlesize` check includes those dependencies.

package.json
JoviDeCroock: And in the react components we are giving where we check for !!client etc

babel.config.js
JoviDeCroock: Also this is not even the big savior, the big savior is the pure calls. I can just remove dev expression and still save 4KB

benjamn:
@JoviDeCroock Can you explain more about how the `/*#__PURE__*/` comments work? What part of the bundled code are we hoping to eliminate? What bundling tools will benefit from those annotations? Are there any recommendations that we should be making to developers?

From what I can tell, this PR slightly *increases* the sizes of `lib/react-apollo.umd.js` (6257B to 6333B, with minification+gzip) and `lib/react-apollo.browser.umd.js` (6231B to 6307B). What's your understanding of why that would be happening? At a first glance, it seems important because `lib/react-apollo.umd.js` is the `main` field in `package.json`, whereas `dist/bundlesize.js` is just a CircleCI build artifact. But maybe these changes mean that some npm dependencies could be dropped, so the (small) increase is offset by other savings?

JoviDeCroock:
@benjamn what it does is it function at the top level when they're pure. This in turns helps tools like uglify do a way better job at dead code elimination. This implies a bit more code in our umd bundles (since it has some annotations but after passing through terser/uglify/... 

Potentially way less code. These top level calls will also be recognised in assignment contexts and thus could be reused.

rosskevin:
> Note that one of the few things you can do then to reduce bundle size in RA is to cut in dependencies because they take up 50% of RA's bundle.

But again, umd bundle size matters very little to an app's bundled usage with tree shaking of an es module. The umd is a data point, in some simple libraries a perfect measure, but be aware for example that I don't use the reexported `compose` and therefore pay no price for the size of it.  I just want to make sure that are on the same page.

The first order of business to to make sure that our bundle and our dependencies bundles are tree shakable, then go from there. 

Umd size _can_ be a vanity metric, let's make sure we don't mistake it as the only indicator of a healthy bundle.

JoviDeCroock:
Yes but things like __DEV__ Are meant for library authors so the pure functions get annotated for things like your terser plugin.

That being said you are right that an ES Bundle will probably decrease the bundle size but for example the huge dep `lodash.isEqual` will always be included since it is part of a function that get used a lot. I'm just wondering if we can evaluate if a function of that magnitude is needed there.

rosskevin:
Definitely agree on that.

JoviDeCroock:
Okay, I've put some thought into this and I would fully DROP the build artifact and use the build artifact in a more optimal manner. Let `bundleSize` itself figure out the gzipped size and validate the ES, CJS and UMD output itself.

This way discussions like these can be avoided? If this gets approved by @benjamn @hwillson  @rosskevin I'll gladly implement it, this way we can cleanly show the bundleSizes instead of this umd artifact.

I think this will also make a huge benefit for `apollo-client`.

rosskevin:
I think that is a good idea. Let's PR that here and review it, then I'll take those changes to the apollo-client monorepo PR.

rosskevin:
https://github.com/TrySound/rollup-plugin-size-snapshot looks nicer for sizing.

JoviDeCroock:
> https://github.com/TrySound/rollup-plugin-size-snapshot looks nicer for sizing.

I think that can't be integrated in our build process (as in the PR checks etc). Maybe it can i'll look into it

JoviDeCroock:
I have drastically altered this proposal in favor of just providing an esm bundle. Up to you to decide whether or not you like the current approach and if you would like the exploded cjs output to have another cjs step (?) I can do this by removing the tsc step and moving to the typescript-rollup-plugin

JoviDeCroock:
Last comment I saw was bundles cjs, using tsc beforehand results in the same as rollup-tsc just 4 seperate times vs once

rosskevin:
We are removing the cjs bundle from rollup, there is one tsc run for exploded cjs.  This is clear, no need to wait.

We can also remove totally the umd - it would only be useful for a bundlesize and is not exposed externally.

JoviDeCroock:
I have made one more attempt, it just feels like the requirement changes a lot and if this changes. I'll probably wait until the apollo-client one gets merged so we have a unary steering direction.

Made a seperate tsconfig.cjs.json since rollup can't compile commonjs

rosskevin:
I think this looks good.  

Timewise there are a few things we can do to improve it:
1. the price should only paid on bundling for release, so `postcompile` should really be `prepublishOnly` - that kills the rollup step entirely for the dev workflow - see https://docs.npmjs.com/misc/scripts
2. Even in the rollup step, I think we may be able to kill the umd builds later

Nonetheless, I do recommend `postcompile` -> `prepublishOnly` which I think is an improvement intended by npm that we haven't been using in the past.

JoviDeCroock:
> I think this looks good.
> 
> Timewise there are a few things we can do to improve it:
> 
> 1. the price should only paid on bundling for release, so `postcompile` should really be `prepublishOnly` - that kills the rollup step entirely for the dev workflow - see https://docs.npmjs.com/misc/scripts
> 2. Even in the rollup step, I think we may be able to kill the umd builds later
> 
> Nonetheless, I do recommend `postcompile` -> `prepublishOnly` which I think is an improvement intended by npm that we haven't been using in the past.

Do you wnat me to change that now or?

Issue #2643, opened 2018-12-05
`awaitFetchQueries` does not obey `mutate`'s `defaultOptions`


justinanastos:
The default value set for `awaitFetchQueries` in [`react-apollo`'s `src/Mutation.tsx`](https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L187-L194) will always override spreading in of `defaultOptions` [`apollo-client`'s `src/ApolloClient.ts`](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L305-L310).

Caused by https://github.com/apollographql/react-apollo/commit/924c80e29683d29d7e7b97f95993ad2b33e64273?w=1#diff-bc5fe8407816a086fbe0a9890a115403R202

Piwaii:
nothing new here ?

Piwaii:
you mean `awaitRefetchQueries` instead of `awaitFetchQueries` ?

hwillson:
React Apollo has been refactored to use React Hooks behind the scenes for everything, which means a lot has changed since this issue was opened (and a lot of outstanding issues have been resolved). We'll close this issue off, but please let us know if you're still encountering this problem using React Apollo >= 3.1.0. Thanks!

tutok:
looks like the issue is still valid

https://github.com/apollographql/react-apollo/blob/91cbdcba17413db3e19ae909c35723e71cd57c38/packages/hooks/src/data/MutationData.ts#L87

Borales:
@hwillson even though a lot has changed - the bug is still alive in `3.1.3`.

The reason is that you call `ApolloClient.mutate` function with a predefined value for `awaitRefetchQueries = false` here (lines 87 and 104):

https://github.com/apollographql/react-apollo/blob/b2a0a2528c44610f41d38f59c39970fa52899532/packages/hooks/src/data/MutationData.ts#L78-L111

but when ApolloClient merges the provided `options` param with the `defaultOptions` here:

https://github.com/apollographql/apollo-client/blob/f0c571597c6019e79ba81a81723ad398d4ea16ad/packages/apollo-client/src/ApolloClient.ts#L344

It's too late, it'll always remain `awaitRefetchQueries = false`. And you do this for `query` method the same thing.

I really hope you can fix this, as we use Apollo packages pretty intensively in our project ✌️

PR #2709, opened 2019-01-06
awaitRefetchQueries defaultOption not being respected by the Mutation component

JoviDeCroock:
### Checklist:

In the issue: https://github.com/apollographql/react-apollo/issues/2643

It was brought to my attention that the defaultOptions provided to the ApolloClient are not respected by react-apollo. That is why I opted to fix this and add a test for it.

### DIFF
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -184,19 +184,19 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
       });
   };
 
-  private mutate = (options: MutationOptions<TData, TVariables>) => {
+  private mutate = (options: MutationOptions<TData, TVariables> = {}) => {
     const {
       mutation,
       variables,
       optimisticResponse,
       update,
       context = {},
-      awaitRefetchQueries = false,
+      awaitRefetchQueries = (this.client.defaultOptions.mutate && this.client.defaultOptions.mutate.awaitRefetchQueries) || false,
       fetchPolicy,
     } = this.props;
-    const mutateOptions = { ...options };
 
-    let refetchQueries = mutateOptions.refetchQueries || this.props.refetchQueries;
+    const mutateOptions = { ...options };
+    let refetchQueries =  mutateOptions.refetchQueries || this.props.refetchQueries;
     // XXX this will be removed in the 3.0 of Apollo Client. Currently, we
     // support refectching of named queries which just pulls the latest
     // variables to match. This forces us to either a) keep all queries around
--- a/test/client/Mutation.test.tsx
+++ b/test/client/Mutation.test.tsx
@@ -561,6 +561,37 @@ it('renders an error state and does not throw when encountering graphql errors w
   );
 });
 
+it('does not override the defaultOptions.awaitRefetchQueries by defaults.', done => {
+  let count = 0;
+  const mutateSpy = jest.fn().mockImplementation(() => Promise.resolve({ data: {} }));;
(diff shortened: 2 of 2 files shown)


src/Mutation.tsx
```diff
@@ -184,19 +184,19 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
...
       context = {},
-      awaitRefetchQueries = false,
+      awaitRefetchQueries = (this.client.defaultOptions.mutate && this.client.defaultOptions.mutate.awaitRefetchQueries) || false,
       fetchPolicy,
     } = this.props;
-    const mutateOptions = { ...options };
 
-    let refetchQueries = mutateOptions.refetchQueries || this.props.refetchQueries;
+    const mutateOptions = { ...options };
+    let refetchQueries =  mutateOptions.refetchQueries || this.props.refetchQueries;
```
danilobuerger: There are 2 spaces after the `=`, please remove one of them.

src/Mutation.tsx
```diff
@@ -184,19 +184,19 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
       });
   };
 
-  private mutate = (options: MutationOptions<TData, TVariables>) => {
+  private mutate = (options: MutationOptions<TData, TVariables> = {}) => {
```
danilobuerger: Why are you adding a default value to `options`? Is this necessary?

src/Mutation.tsx
```diff
@@ -184,19 +184,19 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
...
     const {
       mutation,
       variables,
       optimisticResponse,
       update,
       context = {},
-      awaitRefetchQueries = false,
+      awaitRefetchQueries = (this.client.defaultOptions.mutate && this.client.defaultOptions.mutate.awaitRefetchQueries) || false,
```
danilobuerger: I don't think this is a great idea, since the default options are being applied inside apollo-client. Instead, you should just remove the default value for `awaitRefetchQueries` and only pass it into `client.mutate` if it is `!== undefined`. Then the default option will be applied inside the client.

JoviDeCroock: Hmm just assumed this was defaulted  for a reason so I kept it that way, will change then.

danilobuerger: It defaults to `false` inside the query manager anyway, so thats alright.

JoviDeCroock: The defaultOptions aren't being applied it seems... Test fails if I remove it

JoviDeCroock: Also gets defaulted here, so how can it respect the defaultOptions then? https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/QueryManager.ts#L127

danilobuerger: As I wrote, you also need to make sure not to pass it into `mutate` if its `undefined`, as that counts as the option being set.
```
const a = {foo: "bar"}
const b = {foo: undefined}
const c = {...a, ...b}
// c will be {foo: undefined}
```

JoviDeCroock: Yes but it is getting defaulted anyway as linked there, so when passing undefined it results in false anyway. When my defaultOptions is true it won't get respected since client makes no implication of respecting defaultOptions...

If it defaults an argument it has to make an assumption or it can't merge it with defaultOptions

PR #2710, opened 2019-01-06
Invoke onCompleted/onError even if Mutiation unmounts

### LINKED ISSUE #2293: Query onCompleted not invoked
I want a function invoked when the Query completes. The callback function passed to my Query onCompleted should be invoked
 

The onCompleted callback was not invoked (I used a simple console.log function)...
The Query works, the data is available, just not seeing onCompleted

Create a simple Query component - just a graphql query - no Subscription
Set a simple onCompleted={()=>console.log('onCompleted')}

<Query
        query={MEMBER_WITH_EMAIL}
        variables={{ email: `${email}` }} 
        onCompleted={()=>console.log('The Query onCompleted callback was invoked')}>

        {({ loading, error, ...result }) => (
          <div>
            <MemberWithEmailResults
              {...result}
              loading = {loading}
              error = {error}
              email = {email}
            />
          </div>
        )}
      </Query>
 

**Version**

* apollo-client@2.4.0
* react-apollo@2.1.4

wsmd:
Closes #2293 [(Query onCompleted not invoked)](https://github.com/apollographql/react-apollo/issues/2293)

## The issue

When passing both `onCompleted` and `refetchQueries` to a `Mutation` that un-mounts as a result of `refetchQueries`, `onCompleted` will never be called.

Consider the following example:

A logout button that invokes a mutation, which re-fetches a query to propagate this change to the rest of the app (e.g. removing references to the currently logged-in user).

On that mutation, we have an `onCompleted` to redirect to the login page after both _a successful logout mutation_ AND _a successful refetch_ .

With the current implementation, `onCompleted` will never be called because as soon as `refetchQueries` is finished, the logout mutation will unmount and the redirect will never happen.

```jsx
{loggedIn && ( // ← mutation rendering is conditional 
  <Mutation
    awaitRefetchQueries
    mutation={logoutMutation}
    refetchQueries={[/* a query to update the current user info (e.g. loggedIn) */]}
    onCompleted={redirectToLogin /* this callback will never be called */}
  >
    {/* a logout button to run the mutation when clicked */}
  </Mutation>
)}
```

This might not be the best example, but hopefully it can demonstrate the real issue the PR is trying to address.

### Replicating the Issue

I was able to replicate this issue easily when I tried to pass an `onCompleted` callback to the Mutation in this test and assert that it was called:
 - https://github.com/apollographql/react-apollo/blob/master/test/client/Mutation.test.tsx#L1479-L1518

## The Cause

It appears that this is intentional in order to avoid updating the Mutation internal state after it un-mounts.

https://github.com/apollographql/react-apollo/blob/cce02a346643a2cea5547304d56f2e2f05b74fcd/src/Mutation.tsx#L244-L247

Unless I'm missing something here or there is a reason not do so, this doesn't necessarily mean that an `onCompleted` callback should never be called when a mutation is resolved and has been unmounted.

While it can be hard to think that an unmounted component would have callbacks called, it is important to keep in mind that consumers of an unmounted Mutation have no indication to when the said mutation is resolved successfully or when its refetched queries are finished (same for errors).

## Possible Fix

By changing the logic of [`onMutationCompleted`](https://github.com/apollographql/react-apollo/blob/cce02a346643a2cea5547304d56f2e2f05b74fcd/src/Mutation.tsx#L244-L261) a little bit, it looks like it's possible to invoke the user-provided `onComplete` even when the `Mutation` unmounts.

I also went ahead and applied the same changes to `onError`/`onMutationError` for consistency.

### DIFF
--- a/test/client/Mutation.test.tsx
+++ b/test/client/Mutation.test.tsx
@@ -1477,11 +1477,13 @@ it('errors when changing from mutation to a subscription', done => {
 });
 
 it('does not update state after receiving data after it has been unmounted', done => {
+  const onCompleted = jest.fn();
   let success = false;
   let original = console.error;
   console.error = jest.fn();
   const checker = () => {
     setTimeout(() => {
+      expect(onCompleted).toHaveBeenCalled();
       expect(console.error).not.toHaveBeenCalled();
       console.error = original;
       success = true;
@@ -1500,7 +1502,7 @@ it('does not update state after receiving data after it has been unmounted', don
         return null;
       } else {
         return (
-          <Mutation mutation={mutation}>
+          <Mutation mutation={mutation} onCompleted={onCompleted}>
             {(createTodo, result) => {
               if (!result.called) {
                 setTimeout(() => {
@@ -1529,10 +1531,12 @@ it('does not update state after receiving data after it has been unmounted', don
 });
 
 it('does not update state after receiving error after it has been unmounted', done => {
+  const onError = jest.fn();
   let original = console.error;
   console.error = jest.fn();
   const checker = () => {
     setTimeout(() => {
+      expect(onError).toHaveBeenCalled();
       expect(console.error).not.toHaveBeenCalled();
       console.error = original;
       done();
@@ -1550,7 +1554,7 @@ it('does not update state after receiving error after it has been unmounted', do
         return null;
       } else {
         return (
-          <Mutation mutation={mutation}>
+          <Mutation mutation={mutation} onError={onError}>
             {(createTodo, result) => {
               if (!result.called) {
                 setTimeout(() => {
--- a/test/client/Mutation.test.tsx
+++ b/test/client/Mutation.test.tsx
@@ -1476,111 +1476,210 @@ it('errors when changing from mutation to a subscription', done => {
   console.log = errorLogger; // tslint:disable-line
 });
 
-it('does not update state after receiving data after it has been unmounted', done => {
-  const onCompleted = jest.fn();
-  let success = false;
-  let original = console.error;
-  console.error = jest.fn();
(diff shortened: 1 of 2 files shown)


test/client/Mutation.test.tsx
```diff
@@ -1500,7 +1502,7 @@ it('does not update state after receiving data after it has been unmounted', don
         return null;
       } else {
         return (
-          <Mutation mutation={mutation}>
+          <Mutation mutation={mutation} onCompleted={onCompleted}>
```
wsmd: I used an existing test just to get something out there and to make sure it's working as expected. I can also add a separate test for this use case. Please let me know.

JoviDeCroock: Would you mind extracting this to a seperate test? Improves error reporting

test/client/Mutation.test.tsx
```diff
@@ -1476,107 +1476,210 @@ it('errors when changing from mutation to a subscription', done => {
...
-      success = true;
-      done();
-    }, 100);
-  };
-
-  class Component extends React.Component {
-    state = {
-      called: false,
+describe('after it has been unmounted', () => {
+  it('calls the onCompleted prop after the mutation is complete', done => {
```
wsmd: I added a new describe block for when Mutation is unmounted, and two new tests to make sure Mutation is calling `onCompleted` and `onError` after it has been unmounted. This should cover all changes introduced in this PR.

```
  after it has been unmounted
    ✓ calls the onCompleted prop after the mutation is complete (109ms)
    ✓ calls the onError prop if the mutation encounters an error (109ms)
    ✓ does not update state after receiving data (111ms)
    ✓ does not update state after receiving error (123ms)
```

PR #2716, opened 2019-01-07
Add complete handler to Subscription, to handle the case when the ser…

sujeetsr:
Handle finite subscriptions
(https://github.com/apollographql/subscriptions-transport-ws/issues/471)

Depends on 
https://github.com/apollographql/apollo-client/pull/4290

@hwillson - please review.


sujeetsr:
Sample client and server app to test this: https://github.com/sujeetsr/subscriptions-example

src/Subscriptions.tsx
```diff
@@ -29,6 +29,7 @@ export interface SubscriptionProps<TData = any, TVariables = OperationVariables>
   shouldResubscribe?: any;
   client?: ApolloClient<Object>;
   onSubscriptionData?: (options: OnSubscriptionDataOptions<TData>) => any;
+  onSubscriptionComplete?: () => any;
```
hwillson: Since we're not doing anything with the return value of `onSubscriptionComplete`, let's change the `any` return type to `void`.

src/Subscriptions.tsx
```diff
@@ -165,6 +168,15 @@ class Subscription<TData = any, TVariables = any> extends React.Component<
     });
   };
 
+  private completeSubscription = () => {
+    const {
+      client,
```
hwillson: Can you remove `client` here, since it's not being used?

sujeetsr:
@hwillson thanks for merging!

Issue #1749, opened 2018-03-02
After a query has resulted in an error, variable updates that should result in an Apollo QueryResult with no errors do not update the QueryResult's error field in componentWillReceiveProps


benjaminsaurusrex:
**Actual Result**

* Component mounts using `graphql` HOC, with a Query that produces an error
* Update props so that variables passed to the operation options will now result in a query that succeeds with no errors
* No network request is made, and inside `componentWillReceiveProps`, the `nextProps.data.error` is still defined with the previous error

**Intended Result**

* Component mounts using `graphql` HOC, with a Query that produces an error
* Update props so that variables passed to the operation options will now result in a query that succeeds with no errors
* Either a network request is made, or the data is returned from the cache such that `nextProps.data.error` is no longer defined

## How am I testing this?

Good question. I have a basic query that requests a field that always succeeds, and another field that fails when included:

```
// query.gql
query SomeQuery($shouldFail: Boolean!) {
  successField {
    id
  }
  failField @include(if: $shouldFail) {
    id
  }
}
```

Now for my components. I have 2 components, one that acts as a parent and determines whether its child `shouldFail` or not so that the child gets prop updates, which trigger a variable update when the operation options are different.

```
// Child
class ChildPresentation extends React.Component<Props> { ... }

// Child connected to graphql HOC
const Child = graphql(query, {
  options: (props: Props) => ({
    variables: {
      shouldFail: props.shouldFail,
    },
  }),
})(ChildPresentation);

// Parent
class Parent extends React.Component<ParentProps> {
  public render() {
    return <Child shouldFail={this.state.shouldFail} />;
  }
}
```

With this set up, you can see how `Parent` can easily update its state, which causes Child's `shouldFail` prop to update, which causes the variable inside the operation options to update... and down the rabbit hole.

## The rabbit hole

When a GraphQL error comes through, the Observable decides to clean itself up.

https://github.com/zenparsing/zen-observable/blob/e924b1ddcde8aaac157cedb1908b2ad83b20e7bc/zen-observable.js#L184

Which removes it from `ObservableQuery`'s list of observers (emptying it).

https://github.com/apollographql/apollo-client/blob/2.0/packages/apollo-client/src/core/ObservableQuery.ts#L538

Then when the new query variables come in, it follows the path of:

* `GraphQL.componentWillReceiveProps`
[...]

PR #2718, opened 2019-01-09
Add test for Query bug which is caused by changing props (#1749)

MerzDaniel:
The bug was reported some months ago and I stumbled over it after 1h of trying out react-apollo. This test show-cases the wrong behaviour and fails.

In the related bug ticket someone already debugged into the cause of this issue. 

https://github.com/apollographql/react-apollo/issues/1749


hwillson:
Thanks for this @MerzDaniel - something definitely appears to be off in `case 2`. I've re-enabled the test so this PR will now fail, and will dig into this further. One thing to note however - the `expect(props.data.allPeople).toBeFalsy()` check you had in `case 1` is invalid. React Apollo makes the last known good data available when an error is received, in-case your UI wants to leverage the good data in some way. I've adjusted your test accordingly.

hwillson:
Noting this here in-case I'm sidetracked and someone else is interested in taking a look - this test is currently failing because of the query re-subscription logic that happens when the error is received:

https://github.com/apollographql/react-apollo/blob/b591d030ee1d7784995eb2568d2705c52ca43b61/src/Query.tsx#L348

The `resubscribeToQuery` approach is a known hack (see the discussion in https://github.com/apollographql/react-apollo/pull/1580). This should be addressed in `apollo-client` proper.

sfcgeorge:
Not sure if this is the same bug but I have had issues with changing props not updating the UI properly.

I think the first query made and cached with the first set of props causes issues. Subsequent changes to props and thus queries render fine, but going back to the initial set of props doesn't update the UI.

I tried to reproduce using CodeSandbox but couldn't, so I think it needs a real network layer to occur.

PR #2720, opened 2019-01-10
remove connection directive before creating mock key

### LINKED ISSUE #2719: MockedProvider and @connection
**Intended outcome:**

When providing a query to `MockedProvider`, it works when the exact same query is passed to the `Query` component.

**Actual outcome:**

Error: Network error: No more mocked responses for the query.

I have put together a code sandbox that demonstrates the issue pretty clearly: https://codesandbox.io/s/085pqppxn

If you have an `@connection` directive on the query that is built with `gql`, when passing it to `Query`'s `query` prop, it seems that the directive is removed before sending the query over the wire.

However, when passing the same query as a mock to `MockedProvider`, the directive is left in, resulting in the query never matching and the 'No more mocked responses for the query' happening.

See https://codesandbox.io/s/085pqppxn

**Version**

- apollo-client@2.4.8
- react-apollo@2.3.3

(both latest at time of submission)

deecewan:
Fixes #2719 

Before a query is sent by the Apollo client, the `@connection` directive is stripped out. To save people from having to make two queries that very close match for the sake of a `request.query`, I think the mocked request queries should have the same transformation applied

### Checklist:

### DIFF
--- a/src/test-links.ts
+++ b/src/test-links.ts
@@ -8,7 +8,7 @@ import {
 } from 'apollo-link';
 
 import { print } from 'graphql/language/printer';
-import { addTypenameToDocument } from 'apollo-utilities';
+import { addTypenameToDocument, removeConnectionDirectiveFromDocument } from 'apollo-utilities';
 const isEqual = require('lodash.isequal');
 
 export interface MockedResponse {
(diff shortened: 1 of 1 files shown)


hwillson:
This was recently fixed in https://github.com/apollographql/react-apollo/pull/2523, but thanks very much for this PR @deecewan!

PR #2729, opened 2019-01-15
enhance hoist-non-react-statics types

HugeSkull:
Use @types/hoist-non-react-statics replace custom type definitions.

The current custom type definition return value  [TTargetComponent](https://github.com/apollographql/react-apollo/pull/2729/files#diff-ba025dacde35aa8c151d5104ea946a81L15) type not contain `TSourceComponent` static member.


rosskevin:
Thanks @HugeSkull

Issue #2424, opened 2018-09-27
Breaking typescript change - definition of "data" on Query Result changed


tomitrescak:
Since upgrade to newest version of react-apollo the data definition on QueryProps has changed from

```ts
export interface QueryResult<TData = any, TVariables = OperationVariables> extends ObservableQueryFields<TData, TVariables> {
...
data: TData | undefined;
}
```

to

```ts
export interface QueryResult<TData = any, TVariables = OperationVariables> extends ObservableQueryFields<TData, TVariables> {
...
data: TData | {};
}
```

This is a major breaking change as all projects with updated client do not compile and fail CI. What is the reason for this change? Will this be rectified or shall we start changing our code bases? Thanks for the wonderful work you are doing guys.

**Version**

* react-apollo@2.2.1

danilobuerger:
This was introduced by #1983. The reason was that when using `data`, the `undefined` check can be omitted. Instead of doing `data && data.user` you can just check `data.user`. Full changelog:

- When a query failed on the first result, the query result `data` was being
  returned as `undefined`. This behavior has been changed so that `data` is
  returned as an empty object. This makes checking for data (e.g.
  instead of `data && data.user` you can just check `data.user`) and
  destructring (e.g. `{ data: { user } }`) easier. **Note:** this could
  potentially hurt applications that are relying on a falsey check of `data`
  to see if any query errors have occurred. A better (and supported) way to
  check for errors is to use the result `errors` property.

@tomitrescak can you provide an example of what you were doing that now doesn't compile and fails?

JakeSidSmith:
I had a look into this yesterday, and had a poke inside `Query.test.tsx`. The tests only have no issues because any test cases where the keys of the data are accessed use the `Query` class itself (rather than extending it), and default data to `any`.

I added the following (using some existing test data), and you can see that without casting the data to `any`, you cannot access its keys.

<img width="463" alt="screen shot 2018-09-28 at 09 26 37" src="https://user-images.githubusercontent.com/5850625/46197135-ca861000-c300-11e8-8af0-92a395d12d2b.png">

```
it('result type definition lets you access data keys', () => {
  const Component = () => (
    <AllPeopleQuery query={allPeopleQuery}>
      {(result) => {
        if (result.data && result.data.allPeople) {
          return null;
        }

        if (result.data && result.data!.allPeople) {
          return null;
        }

        const { allPeople } = result.data;

        return null;
      }}
    </AllPeopleQuery>
  );
});
[...]

PR #2734, opened 2019-01-17
Fix type definition for QueryResult data property

LarsJK:
Query result is an empty object or TData, not undefined or TData.

I ran into a problem with the current type definition.

```tsx
interface MyResult {
  item: {
    name: string;
  };
}

<Query<MyResult, ..>>
{({ data }) => {
  // The old type of data would be MyResult | undefined so I unwrap it like this:
  const name = data && data.item.name || 'default name'
  // TypeError: Cannot read property 'name' of undefined
}}
</Query>
```

[Partial](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types) makes each of its generic type's properties optional

I could ofcourse change my TData type to have optional keys:

```ts
interface MyResult {
  item?: {
    name: string;
  };
}
```

But then it doesn't directly reflect what response my graphql server gives, and in my case the types are generated..

### DIFF
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -75,13 +75,14 @@ export interface QueryResult<TData = any, TVariables = OperationVariables>
   // I'm aware of. So intead we enforce checking for data
   // like so result.data!.user. This tells TS to use TData
   // XXX is there a better way to do this?
-  data: TData | undefined;
+  data: Partial<TData>;
   error?: ApolloError;
   loading: boolean;
   networkStatus: NetworkStatus;
 }
 
-export interface QueryProps<TData = any, TVariables = OperationVariables> extends QueryOpts<TVariables> {
+export interface QueryProps<TData = any, TVariables = OperationVariables>
+  extends QueryOpts<TVariables> {
   children: (result: QueryResult<TData, TVariables>) => React.ReactNode;
   query: DocumentNode;
   displayName?: string;
--- a/src/query-hoc.tsx
+++ b/src/query-hoc.tsx
@@ -87,7 +87,7 @@ export function withQuery<
               // we massage the Query components shape here to replicate that
               const result = Object.assign(r, data || {});
               const name = operationOptions.name || 'data';
-              let childProps = { [name]: result };
+              let childProps: { [x: string]: any | undefined } = { [name]: result };
               if (operationOptions.props) {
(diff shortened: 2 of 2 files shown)


danilobuerger:
We are not going to merge this. Please see https://github.com/apollographql/react-apollo/issues/2424 and related for the reasoning.

LarsJK:
Could you please reconsider this?

Having a type definition that is wrong and can not be trusted is dangerous. As you can see in my example it causes runtime crashes, which are exactly the kind of things typescript helps to prevent.

The referenced issue is also not the same as this implementation so it doesn't really give any reason to why this should be closed, and I **do** understand why that was reverted..


Given an interface:
```ts
interface Foo {
  bar: {
    baz: number;
  };
}
```

```ts
var foo: Foo | {} = { bar: { baz: 100 }}
foo.bar.baz // error TS2339: Property 'bar' does not exist on type '{} | Foo'.
  Property 'bar' does not exist on type '{}'.
```

```ts
var foo: Foo | undefined = {}
foo.bar.baz // With strict null checks TS warns me about foo: Object is possibly 'undefined'.
// So I try too first check fo null:
foo && bar.baz // Now TS is happy but this is actually worse 😱:
// I get runtime crashes: TypeError: Cannot read property 'name' of undefined
```

```ts
var foo: Partial<Foo> = {}
foo.bar.baz // With strict null checks TS warns me about bar: Object is possibly 'undefined'.
// So I try too first check fo null:
foo.bar && foo.bar.baz // Now TS is happy and..
// Yay no runtime crashes 💯 
```

So a partial directly reflects the value that is actually used, it doesn't loose type and it can prevent runtime crashes!

danilobuerger:
> Having a type definition that is wrong and can not be trusted is dangerous. 

The type definition is not wrong though. It is exactly what it is supposed to be. We either get the whole `TData` or `undefined`. 

> var foo: Foo | undefined = {}

This is not correct anyway. You can't assign an empty object here.

LarsJK:
@danilobuerger do you ever do that tho? I always get back an empty object, and when I get an empty object it is wrong.. it's not TData and it's not undefined.. If you do assign an empty object then it should be `Partial<TData> | undefined`..

LarsJK:
Okay thanks. Will open a bug then. The way I understood it the empty object return was added as a feature to allow safe destruction of data..

Issue #2738, opened 2019-01-18
MockedProvider - Cannot read property 'registeredQueries' of undefined


rosskevin:
**Intended outcome:**
MockedProvider works in storybook(?) without error.

**Actual outcome:**
```
Cannot read property 'registeredQueries' of undefined
    TypeError: Cannot read property 'registeredQueries' of undefined
    at MockedProvider.push../node_modules/react-apollo/test-utils.js.MockedProvider.componentWillUnmount (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:672705:35)
    at callComponentWillUnmountWithTimer (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:692391:12)
    at HTMLUnknownElement.callCallback (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:676475:14)
    at Object.invokeGuardedCallbackDev (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:676525:16)
    at invokeGuardedCallback (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:676582:31)
    at safelyCallComponentWillUnmount (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:692398:5)
    at commitUnmount (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:692772:11)
    at commitNestedUnmounts (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:692803:5)
    at unmountHostComponents (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:693064:7)
    at commitDeletion (http://localhost:6016/vendors~main.92813365353641d68d2e.bundle.js:693116:5)
  ```

**How to reproduce the issue:**
On initial click on a story, everything is good.  Switch to a different story- good.  Switch _back_ to the first story- boom.

This appears to be a race condition that is adjacent to what we saw in #2164 and @danilobuerger fixed in #2165.  I have spotted a potential root cause - working on it and PR now.

**Version**

- react-apollo@2.3.3
- apollo-client@2.4.9

danilobuerger:
This seems unrelated. `scheduler` was removed by @benjamn  a few hours ago thus breaking here: https://github.com/apollographql/apollo-client/commit/b4f0c8ea6c7564d8762b663903847919a6d37105

rosskevin:
[...]

PR #2741, opened 2019-01-19
Use ApolloClient#stop to clean up MockedProvider client.

benjamn:
This `stop` method was introduced in apollo-client@2.4.12: https://github.com/apollographql/apollo-client/pull/4336

Credit to @danilobuerger for suggesting we hide the details of stopping the `MockedProvider` client using a new `ApolloClient` method: https://apollographql.slack.com/archives/C2FERSTGC/p1547849096341900

The `stop` method would have made https://github.com/apollographql/react-apollo/issues/2738 a non-issue, since we could easily change the implementation of `stop` without affecting its consumers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -696,9 +696,9 @@
       }
     },
     "apollo-client": {
-      "version": "2.4.11",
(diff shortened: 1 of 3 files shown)


danilobuerger:
Works on my machine 😅

PR #2746, opened 2019-01-21
Add test that demonstrates why query result types are wrong

LarsJK:
Hi @danilobuerger.

Regarding #2734 

I modified one of the Query tests to prove why I mean the QueryResult data field type is wrong.

I don't think that the query returning {} instead of undefined for TData is a bug tho. It looks like it is intentional and there because it makes it easier to destruct:

```js
var {data} = undefined
// TypeError: Cannot destructure property `data` of 'undefined' or 'null'.
var {data} = {}
```


danilobuerger:
Thanks for the example. I will get back to you here as soon as we have fixed it. We are going to stick with `TData | undefined` though. If you are interested why, read through past issues (there have been a couple) on this subject.

danilobuerger:
Fixed in https://github.com/apollographql/apollo-client/pull/4356
Going to update react-apollo once we have a new apollo-client release

hwillson:
@LarsJK thanks for this PR. I think there are a few issues with your test though:

1. Your `AllPeopleQuery` component is configured to use the `allPeopleQuery` query that's defined outside of your test case, but in your test case you've defined a new `query` that I'm assuming you want to use instead.

2. You're defining `variables` in the mock request you're passing into `MockedProvider`, but you're not also passing those `variables` into the `AllPeopleQuery` query component. This means your incoming query can't be associated with the mock request you've setup, so your mocked response isn't being returned.

2. Your `AllPeopleQuery` component is ignoring the initial component load state, which means when `result.data` comes back as an empty object (first), it will cause your test to fail due to the attempts to reference non-existent properties, multiple levels deep in the returned `{}`. 

Putting the above together, you'll end up with:

```js
  it('should be able to access data keys without a type guard', done => {
    const query = gql`
      query people($first: Int) {
        allPeople(first: $first) {
          people {
            name
          }
        }
      }
    `;

    const data = {
      allPeople: {
        people: [{ name: 'Luke Skywalker' }],
      },
    };

    const variables = {
      first: 1,
    };

    const mocks = [
      {
        request: { query, variables },
        result: { data },
      },
    ];

    const Component = () => (
      <AllPeopleQuery query={query} variables={variables}>
        {result => {
          if (!result.loading) {
            if (result.data && result.data.allPeople.people) {
              done();
              return null;
            }

            if (result.data && result.data!.allPeople.people) {
              done();
              return null;
            }

            const {
              allPeople: { people },
            } = result.data!;
            done();
            return null;
          }
          return null;
        }}
      </AllPeopleQuery>
    );

    wrapper = mount(
      <MockedProvider mocks={mocks} addTypename={false}>
        <Component />
      </MockedProvider>,
    );
  });
```

When this is run, the test passes properly. Given the above, I don't think there is anything to do here? I'm happy to discuss this further, but I'll close this PR for now. Thanks!

LarsJK:
Thanks @hwillson 

You are absolutely right about 1 and 2. I dint know how I managed to commit the query, data, and mock variables. They shouldn't be there at all.

Regarding 3 the whole point of the tests was to prove that the type definitions was wrong and allowed me to do `result.data && result.data.allPeople.people` which would crash.

This has been addressed however in apollographql/apollo-client#4356

Do you want me to fix the tests so that you can avoid regressions on this, or do you feel that you don't need them?

hwillson:
Got it - thanks for the follow up @LarsJK. As you mentioned, I think we're okay as this would be caught by `tsc`. Thanks again!

PR #2753, opened 2019-01-25
fix: When doing a SSR with errorPolicy='all', then error should not be lost

### LINKED ISSUE #615: getDataFromTree throws errors when any query fails

### LINKED ISSUE #1781: Errors lost on cached results after errorPolicy: "all"
I'm using apollo and react to call a query that returns mixed results: error(s) and some data.
In order to handle the errors, but still be able to use the information given, I'm using the `errorPolicy: "all"` option in the query.

```
<Query query={query} variables={variables as any} errorPolicy='all'>
    {({loading, error, data}) => { ... }}
</Query>
```

The first time I mount the component `data` is populated with the partial informations and `error` with the errors returned by the query. If I change the route (unmounting the component) and then return on it, it shows the cached partial data **but no errors**, so I'm not able to handle errors anymore and detect that these are partial informations.

**Intended outcome:**
The component shows me the original errors along with the cached data.

**Actual outcome:**
The props `error` is `undefined`, the partial data are passed as if the query didn't return any error.

I don't know how to reproduce a query that returns both data and error; anyone can help me with that so I can show you some working code?

**Version**
- apollo-client@2.2.6
- react-apollo@2.1.0-rc.3

Initially I tought that issue #1389 could related to this problem, but after all I think these are 2 different things. Maybe it's something related with the RC version I'm using?

### LINKED ISSUE #1972: SSR ERROR not pass down to children `renderToStringWithData`
Error on `renderToStringWithData` error data can't reach to components on SSR. Even If i apply `errorPolicy: "all"`.

**Intended outcome:** `this.props.data.error=error_data`

**Actual outcome:**`this.props.data.error= undefined`

**How to reproduce the issue:** Try server rendering with a query(component) return any error on graphql resolver.

**Version**

* apollo-client@2.2.8
* react-apollo@2.1.3

### LINKED ISSUE #2134: SSR: error render props do not be passed into <Query /> as expected (errorPolicy: 'all')
I built a codesandbox app (https://codesandbox.io/s/r7mxnx486q) and a next ssr app(https://github.com/chentsulin/react-apollo-issue) to confirm that they have totally different behavior when rendering the error. 

For demo, I use following code exmaple

```js
function App() {
  return (
    <div className="App">
        <Query query={Hello}>
          {({ data, error, loading }) => {
            console.log({ data, error, loading });
            if (loading) return "loading..";
            if (error) return error.message;
            return data.hello;
          }}
        </Query>
      </ApolloProvider>
    </div>
  );
}
```

And a launchpad to throw a error in Hello query:

https://launchpad.graphql.com/kqrn1n1qj7

### Client Side Rendering

(example: https://codesandbox.io/s/r7mxnx486q)

It works as expected in client side rendering. 

Props were logged twice with following format (got error in second time rendering):

```js
{ data: {}, error: undefined, loading: true }
{ data: undefined, error: ApolloError, loading: false }
```

### Server Side Rendering

(example: https://github.com/chentsulin/react-apollo-issue)

Props were logged twice with no error:

{ data: { hello: null }, error: undefined, loading: false }
{ data: { hello: null }, error: undefined, loading: false }

### Relevant Issues

https://github.com/apollographql/apollo-link/issues/437
https://github.com/apollographql/apollo-client/issues/2880
https://github.com/apollographql/react-apollo/issues/2124
https://github.com/apollographql/react-apollo/issues/2087
https://github.com/apollographql/react-apollo/issues/1389

I dived into source of `<Query />`, and logged some details. I found that it call `watchQuery` twice and create two different queryIds.

```
initializeQueryObservable watchQuery - queryId { queryId: '1' }
fetchData watchQuery - queryId { queryId: '2' }
```

Then it save errors into queryId: 2

```js
queryStore.markQueryResult { queryId: '2',
  result:
   { data: { hello: null },
     errors: [ [Object] ] }
```

And finally it found `queryStoreValue` from queryId: 1 but store only have value for queryId: 2, so it pass undefined as result to render props:

```js
queryId { queryId: '1' }
getStore { '2':
   { document: { kind: 'Document', definitions: [Array], loc: [Object] },
     previousVariables: null,
     networkError: null,
     graphQLErrors: [ [Object] ],
     networkStatus: 7,
     metadata: null } }
{ queryStoreValue: undefined }
```

what is the purpose of those two queryObservable? It is a little confusing.

https://github.com/apollographql/react-apollo/blob/3985b47f086f38f891099732842ca2f16c7966c4/src/Query.tsx#L183-L189

### LINKED ISSUE #2680: errorPolicy all behaves the same as ignore for server side rendering
I believe there is a bug with error handling in react-apollo when using server side rendering which causes the `error` field in a query result to be `undefined`, even when the errorPolicy is set to `all`. 

**Intended outcome:**

When a `<Query>` component is server side renderered with errorPolicy=all, and an error thrown in a resolver, the render props function should be passed a result with an `error` field being populated as [documented here](https://www.apollographql.com/docs/react/features/error-handling.html#policies).

**Actual outcome:**

The `error` field is undefined on the server.

I have created a reproduction example here: https://codesandbox.io/s/3xx4w96o6

I also have a version on github here: https://github.com/ganemone/apollo-ssr-error-handling

**Version**

- apollo-client@latest
- react-apollo@latest

I believe this is a similar issue to https://github.com/apollographql/react-apollo/issues/2655 and https://github.com/apollographql/react-apollo/issues/615 however there was some issues with people being rude in those threads so I am creating a new issue here with more clear reproduction steps. 

I think this should be somewhat high priority since there is not a viable errorPolicy for gracefully handling errors in server side rendering right now.  `none` will cause `renderToStringWithData` to throw, and `all` behaves the same as `ignore`, so you are forced to either render a full error page or ignore errors completely.

mikebm:
Closes #2680
Closes #2134
Closes #1972
Closes #1781 
Closes #615

This change tracks the original observable used during a SSR. When the Query component is recreated due to a Query's promise resolving, then the original queryId used to track the response error is lost. By tracking the original observable, the newly instantiated Query component can reference the original observable to retrieve the actual result. This makes the feature work like the documentation states, which allows server side rendered components to properly handle errors rather than throwing exceptions.

### DIFF
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -172,6 +172,12 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
       ...opts,
       fetchPolicy,
     });
+
+    // Register the SSR observable, so it can be re-used once the value comes back.
+    if (this.context && this.context.renderPromises) {
+      this.context.renderPromises.registerSSRObservable(this, observable);
+    }
+
     const result = this.queryObservable!.currentResult();
 
     return result.loading ? observable.result() : false;
@@ -284,7 +290,16 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
     const opts = this.extractOptsFromProps(props);
     // save for backwards compat of refetcherQueries without a recycler
     this.setOperations(opts);
-    this.queryObservable = this.client.watchQuery(opts);
+
+    // See if there is an existing observable that was used to fetch the same data and
+    // if so, use it instead since it will contain the proper queryId to fetch
+    // the result set. This is used during SSR.
+    if (this.context && this.context.renderPromises) {
+      this.queryObservable = this.context.renderPromises.getSSRObservable(this);
+    }
+    if (!this.queryObservable) {
+      this.queryObservable = this.client.watchQuery(opts);
+    }
   }
 
   private setOperations(props: QueryProps<TData, TVariables>) {
--- a/src/getDataFromTree.ts
+++ b/src/getDataFromTree.ts
@@ -4,89 +4,47 @@ import Query from './Query';
 import { ObservableQuery } from 'apollo-client';
 import { DocumentNode } from 'graphql';
 
-// Like a Set, but for tuples. In practice, this class is used to store
-// (query, JSON.stringify(variables)) tuples.
-class Trie {
-  private children: Map<any, Trie> | null = null;
-  private added = false;
-
-  has(...keys: any[]) {
-    let node: Trie = this;
-    return keys.every(key => {
-      const child = node.children && node.children.get(key);
-      return !!(child && (node = child));
-    }) && node.added;
-  }
-
-  add(...keys: any[]) {
-    let node: Trie = this;
-    keys.forEach(key => {
-      const map = node.children || (node.children = new Map);
-      const child = map.get(key);
-      if (child) {
-        node = child;
-      } else {
-        map.set(key, node = new Trie());
-      }
-    });
-    node.added = true;
-  }
+type QueryInfo = {
+  seen: boolean;
+  observable: ObservableQuery<any, any> | null;
 }
 
-interface ObservableCacheEntry {
-  variablesString: string; // Stringified query variables.
-  observable: ObservableQuery<any, any>; // Observable used to issue the fetch.
+function makeDefaultQueryInfo(): QueryInfo {
+  return {
+    seen: false,
+    observable: null,
+  };
 }
 
 export class RenderPromises {
   // Map from Query component instances to pending fetchData promises.
   private queryPromises = new Map<Query<any, any>, Promise<any>>();
 
-  // Stores ObservableQueries by a query and variables combination.
-  private ssrObservable = new Map<DocumentNode, ObservableCacheEntry[]>();
-
-  // A way of remembering queries we've seen during previous renderings,
-  // so that we never attempt to fetch them again in future renderings.
-  private queryGraveyard = new Trie();
+  // Two-layered map from (query document, stringified variables) to QueryInfo
+  // objects. These QueryInfo objects are intended to survive through the whole
+  // getMarkupFromTree process, whereas specific Query instances do not survive
+  // beyond a single call to renderToStaticMarkup.
+  private queryInfoTrie = new Map<DocumentNode, Map<string, QueryInfo>>();
 
   // Registers the server side rendered observable.
   public registerSSRObservable<TData, TVariables>(
     queryInstance: Query<TData, TVariables>,
     observable: ObservableQuery<any, TVariables>,
   ) {
-    const { query, variables } = queryInstance.props;
-    const variablesString = JSON.stringify(variables);
-    // Cache the Observable so new instances of the Query Component
-    // can use the original observable. Necessary for errorPolicy="all".
-    let mapEntry = this.ssrObservable.get(query);
-    if (!mapEntry) {
-      mapEntry = [];
-      this.ssrObservable.set(query, mapEntry);
-    }
-    mapEntry.push({ variablesString, observable });
+    this.lookupQueryInfo(queryInstance).observable = observable;
   }
 
   // Get's the cached observable that matches the SSR Query instances query and variables.
   public getSSRObservable<TData, TVariables>(queryInstance: Query<TData, TVariables>) {
-    const { query, variables } = queryInstance.props;
-    const variablesString = JSON.stringify(variables);
-    const mapEntry = this.ssrObservable.get(query);
-    if (mapEntry) {
-      const observableEntry = mapEntry.find(entry => entry.variablesString === variablesString);
-      if (observableEntry) {
-        return observableEntry.observable;
-      }
-    }
-
-    return null;
+    return this.lookupQueryInfo(queryInstance).observable;
   }
 
   public addQueryPromise<TData, TVariables>(
     queryInstance: Query<TData, TVariables>,
     finish: () => React.ReactNode,
   ): React.ReactNode {
-    const { query, variables } = queryInstance.props;
-    if (!this.queryGraveyard.has(query, JSON.stringify(variables))) {
+    const info = this.lookupQueryInfo(queryInstance);
+    if (!info.seen) {
       this.queryPromises.set(
         queryInstance,
         new Promise(resolve => {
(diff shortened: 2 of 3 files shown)


mikebm:
> Lots of noise in the diff from unrelated formatting changes :/
Not that much noise.  I can go revert those few changes if necessary, but this was related to the the "noise" not matching the prettier config in package.json.

mikebm:
> Lots of noise in the diff from unrelated formatting changes :/

This was related to the code not aligning to the prettier config in package.json since I use the prettier plugin. I reverted said noise.

mikebm:
> Thanks so much @mikebm!
> 
> Rebased/squashed a bit and added an additional commit that simplifies the `RenderPromises` implementation (our fault not yours).

Perfect. I was thinking along the same lines but shied away from my first PR here to be too invasive :). Thank you much for quickly accepting this.

mikaelgramont:
Many thanks for this fix, folks!

Dattaya:
Thank you for the explanation, I had no idea. So, does that mean that `renderToStringWithData` is deprecated, it used to return rendered app [according to the docs](https://www.apollographql.com/docs/react/features/server-side-rendering.html#renderToStringWithData)?
I must be making another mistake, because `apolloClient.extract()` does not contain an error, and the client side script overrides the correct html from the server.

mikebm:
> Thank you for the explanation, I had no idea. So, does that mean that `renderToStringWithData` is deprecated, it used to return rendered app [according to the docs](https://www.apollographql.com/docs/react/features/server-side-rendering.html#renderToStringWithData).
> I must be making another mistake, because `apolloClient.extract()` does not contain an error, and the client side script overrides the correct html from the server.

I still use it, it simply just calls getMarkupFromTree.  extract won't cache the error for re-hydration because the InMemoryCache won't contain errors, just the cached result.  You should be handling the errors on the server in this case. What I do is render a 404 component that sets the status code that is handled on the server.
Example:
```
export const Status: FunctionComponent<Props> = ({ code, children }) => (
    <Route
        render={({ staticContext }) => {
            if (staticContext) {
                staticContext.statusCode = code;
            }
            return children;
        }}
    />
);
```

On the server:
```
const context = {};
<StaticRouter context={context} location={req.path}>
	<AppComponent />
</StaticRouter>
```

Then on the server, you just check context.statusCode and set the http response code accordingly.

jdmoody:
@mikebm, why doesn't the InMemoryCache contain errors for re-hydration?

I'm running into the exact problem @Dattaya mentioned above. I have an Error component that renders specific information about the GraphQL error. The initial SSR shows the error info as expected, but upon client-side hydration, the error information is removed since `extract` doesn't cache errors.

I'm not sure I understand your solution. In the error case, does the server render static HTML that isn't re-hydrated by the client?

Thanks for all your help in this thread 🙏

mikebm:
> @mikebm, why doesn't the InMemoryCache contain errors for re-hydration?
> 
> I'm running into the exact problem @Dattaya mentioned above. I have an Error component that renders specific information about the GraphQL error. The initial SSR shows the error info as expected, but upon client-side hydration, the error information is removed since `extract` doesn't cache errors.
> 
> I'm not sure I understand your solution. In the error case, does the server render static HTML that isn't re-hydrated by the client?
> 
> Thanks for all your help in this thread 🙏

This is just how the Apollo team designed the InMemoryCache in ApolloClient. It only contains cache keys and cache data.  In my specific use case, I handle the error during the SSR by using react-router's `<Redirect />` component to either a 404, or 500 page depending on the error. On the client side, the apollo client will load whatever data it can from cache depending on your `errorPolicy` and 'fetchPolicy`.

I use `errorPolicy=all`, which allows for partial data to be returned during an error, otherwise you get an exception during the SSR on the server.
Now this is where it gets interesting. If your schema allows for your resolver to return null, then it could cause null to be cached in the InMemoryCache, which would mean (depending on your fetchPolicy) that null will be returned on the client. If your schema resolver does not allow null, then nothing would be cached, and your client would re-issue the request and you would get the original error (assuming it still occurs). Ideally, you would have handled this error in the first place on the server though.

Your use case may be different, but I hope this helps.

Dattaya:
Redirecting is the most elegant solution. So far I save error in `window.__APOLLO_STATE__` and check if it's present on the client, it's quite ugly, but it works. Also, I had to create my own `getDataFromTree` as [recommended in the react-apollo source](https://github.com/apollographql/react-apollo/blob/aff6ed3bc455ce922c750474e3debbfa9dd269dc/src/getDataFromTree.ts#L105) to use `renderToString` instead of `renderToStaticMarkup`:
```js
import { renderToString } from 'react-dom/server'
const getDataFromTree = (tree, context = {}) => getMarkupFromTree({ tree, context, renderFunction: renderToString })
```

maxchehab:
What is the status on this PR?

mikebm:
> @mikebm Are there alternatives to `InMemoryCache` that cache errors?

None that I know of.

PR #2758, opened 2019-01-28
Release 2.5.0

benjamn:
This 2.5.0 release branch/PR provides a staging area for the next **minor** version of `react-apollo`. PRs that introduce minor breaking changes should be merged into the `release-2.5.0` branch instead of `master`, so that we can keep releasing patch updates in the meantime.

Similar PRs on other Apollo repositories:
* https://github.com/apollographql/apollo-client/pull/4361
* https://github.com/apollographql/apollo-server/pull/2215


danilobuerger:
2.5.0-beta.0 works in my react-native builds without any regressions.

brunojdo:
Hey guys, if this is really a breaking change shouldn't be a major? 🙏

benjamn:
@brunojdo The only known breaking change is that `walkTree` is no longer exported from the top-level `react-apollo` package. You will be able to import it from `react-apollo/walkTree` if you still need to use it for some reason, even though there is no way for us to make it compatible with the latest React versions, because of React hooks.

Though it is theoretically a breaking change, I do not think this warrants jumping to `react-apollo@3.0.0`. If you disagree, please consider switching to a `~x.y.z` version constraint for `react-apollo` (instead of `^`) so that you will not accidentally update to new minor versions of the library. And of course it's a good idea to use a `package-lock.json` file so that you won't ever have to worry about unintentional upgrades, but I assume everyone knows that.

hwillson:
Hi all - the React Apollo 2.5.0 release candidate is now ready for testing. If you're interested in trying it out, update to `react-apollo@next` (and `apollo-client@next`). Thanks!

PR #2771, opened 2019-01-31
Enable stripping invariant(condition, message) strings in production.

benjamn:
By replacing
```ts
invariant(condition, message)
```
with
```ts
process.env.NODE_ENV === "production"
  ? invariant(condition)
  : invariant(condition, message)
```
during the Rollup ESM bundling step, we can empower minifiers to replace `process.env.NODE_ENV` with a string literal to eliminate the unreachable branch of the conditional expression, which effectively removes error message string literals in production, without having to build or choose between separate development and production bundles. This optimization removes almost 1KB of minified+gzip size from the `react-apollo` package.

I am somewhat worried about `process.env.NODE_ENV` not being defined in browsers in non-production builds. However, I think we can supply (or recommend) a polyfill for those cases. As a counterpoint, React relies on the existence of `process.env.NODE_ENV` and does not provide a polyfill: https://github.com/facebook/react/blob/master/packages/react/npm/index.js

Because this PR pushes the `process.env.NODE_ENV` checks down into the code where it matters, rather than checking it at the entry point (`index.js`) to choose between the development and production build, we are not forced to use CommonJS like React does (and worse: React reassigns `module.exports` in `index.js`, which prevents tree shaking), so `react-apollo` remains tree-shakeable whether you're using CommonJS via the `main` field of `package.json` or ESM via the `module` field.


src/utils/invariant.ts
```diff
@@ -0,0 +1,15 @@
...
+  framesToPop = 1;
+  name = genericMessage;
+  constructor(message: string = genericMessage) {
+    super(message);
+  }
+}
+
+export default function invariant(condition: any, message: string) {
+  if (!condition) {
+    throw new InvariantError(message);
```
benjamn: Although we're not using `InvariantError` directly outside of this module yet, this line of code gets transformed to
```ts
throw process.env.NODE_ENV === "production" ? new InvariantError(message) : new InvariantError();
```
which is fine/harmless… and even saves a couple bytes in production!

src/utils/invariant.ts
```diff
@@ -0,0 +1,15 @@
+const genericMessage = "Invariant Violation";
+
+export class InvariantError extends Error {
+  framesToPop = 1;
+  name = genericMessage;
```
benjamn: These class properties preserve the API of the original https://www.npmjs.com/package/invariant error objects, in case that matters.

rollup/plugin-invariant.js
```diff
@@ -0,0 +1,77 @@
...
+
+const b = recast.types.builders;
+
+export default function invariantPlugin(options = {}) {
+  const filter = createFilter(options.include, options.exclude);
+
+  return {
+    transform(code, id) {
+      if (!filter(id)) {
+        return;
```
benjamn: Using a `filter` here is apparently good hygiene / the "right" thing to do when implementing a Rollup transformer plugin: https://rollupjs.org/guide/en#transformers

rollup/plugin-invariant.js
```diff
@@ -0,0 +1,77 @@
...
+    node.callee.type === "Identifier" &&
+    node.callee.name === name &&
+    node.arguments.length > length
+  );
+}
+
+function makeNodeEnvTest() {
+  return b.binaryExpression(
+    "===",
+    b.memberExpression(
```
benjamn: Even though this is a `.js` file, VSCode was able to auto-complete all of these builder functions and enforce their parameter types, thanks to the great work @brieb did in https://github.com/benjamn/ast-types/pull/300.

benjamn:
@rosskevin I consider the implementation of the plugin a private concern of this particular package, and I disagree that Babel is the best tool for that job.

This is a build-time plugin, so it doesn't add any runtime dependencies to the library. As long as the compiled code is correctly transformed, the implementation of the plugin should not matter to any consumers of this library.

Because the plugin is local, if we need to change the transform, we can do so immediately, without waiting for another project to respond. That's a big advantage in my mind as we iterate on the best way to implement these optimizations.

Not everything needs to be published to npm, though I certainly agree this plugin should be shared among Apollo packages. We'll cross that bridge when we get to it.

The `babel-plugin-transfom-dev-warning` transform isn't what we want here, because it completely removes the warning, rather than just stripping the error string. The `invariant(condition)` function is supposed to throw (an opaque error) in production. With that said, nothing stops you from using both `warning` and `invariant` outside of this package.

rollup.config.js
```diff
@@ -1,9 +1,9 @@
 import node from 'rollup-plugin-node-resolve';
 import { uglify } from 'rollup-plugin-uglify';
-import replace from 'rollup-plugin-replace';
 import typescript from 'typescript';
 import typescriptPlugin from 'rollup-plugin-typescript2';
 import filesize from 'rollup-plugin-filesize';
+import invariantPlugin from './rollup/plugin-invariant';
```
JoviDeCroock: This is a wonderfull initiative but doesn't this exclude our "exploded" CJS bundle from having benefit from this change?

benjamn: You're right about that. Maybe this argues in favor of compiling `react-apollo.esm.js` to `react-apollo.cjs.js` and pointing the `main` field of `package.json` at that CommonJS bundle?

The only immediate downside that I can think of is that one-off modules like `react-apollo/test-links` and `react-apollo/walkTree` would no longer be built if we don't run `tsc`.

JoviDeCroock: We can keep the exploded bundle by running tsc which enables the /walkTree etc, aslong as the types are exploded there should be close to no issue for web with making a cjs bundle on `main`.

benjamn: Ok, that's a good point. I think people realize that they're taking additional responsibility for their choices when they do sub-module imports like `require("react-apollo/test-links")`, so the risk of loading both the CJS bundle and some overlapping exploded modules should be limited to folks who chose to use sub-module imports.

JoviDeCroock: I think we should however include a warning about it, this way someone encountering a weird issue with it will know what to look for.

rollup.config.js
```diff
@@ -27,15 +27,16 @@ export default [
...
+      invariantPlugin(),
       filesize(),
     ],
     onwarn,
   },
   {
     input: 'lib/react-apollo.esm.js',
     output: {
-      file: 'lib/react-apollo.umd.js',
-      format: 'umd',
```
JoviDeCroock: Is there a reason to fully remove the UMD? Sites like https://unpkg.com/ still look for these to index them so they can be required by CDN. Don't wanna come off as nosy but just trying to get some insights into things like these

benjamn: Ahh, I would be fine with keeping it if there's any value, and unpkg.com seems like a valid use case.

rollup.config.js
benjamn: Took a stab at this verbal warning in https://github.com/apollographql/react-apollo/pull/2771/commits/d6792eb0ba2dc883643e91fa5c31bdb7c1d466dc.

PR #2776, opened 2019-02-04
Changes to accommodate AC's integrated local state handling

hwillson:
This PR makes a few odds/ends changes to React Apollo, to better align it with the upcoming Apollo Client 2.5.0 release and local state. It includes:

- Updating the `apollo-*` deps to use the `beta` versions from Apollo Client (for `react-apollo@beta` testing)
- Updates to the Apollo Link dependency and associated Typescript changes
- Disabling HOC prop variable validation, since it won't work with `@client @export` (see the description in https://github.com/apollographql/react-apollo/commit/cce3316e7e3ed65f554346fe59d6c320d6b695da for more details)
- Prevent React Apollo's `MockedProvider` from using `@client` fields in mocked out links, since `@client` fields are now local (and shouldn't be included in links).
- Start tracking the full last received query result in the `Query` component, so it can be used to help decide when a new render should be prevented (if the incoming result hasn't changed).

Please note that this PR is currently a work in progress, as there are a handful of test failures that need to be worked through. These failures aren't related to React Apollo changes, but are related to updating to `apollo-*` `beta` deps.

### DIFF
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -35,14 +35,19 @@ export declare type MutationUpdaterFn<
   }
 > = (proxy: DataProxy, mutationResult: FetchResult<T>) => void;
 
-export declare type FetchResult<C = Record<string, any>, E = Record<string, any>> = ExecutionResult<
-  C
-> & {
+export declare type FetchResult<
+  TData = Record<string, any>,
+  C = Record<string, any>,
+  E = Record<string, any>
+> = ExecutionResult<TData> & {
   extensions?: E;
   context?: C;
 };
 
-export declare type MutationOptions<TData = any, TVariables = OperationVariables> = {
+export declare type MutationOptions<
+  TData = { [key: string]: any },
+  TVariables = OperationVariables
+> = {
   variables?: TVariables;
   optimisticResponse?: TData;
   refetchQueries?: Array<string | PureQueryOptions> | RefetchQueriesProviderFn;
@@ -174,11 +179,11 @@ class Mutation<TData = any, TVariables = OperationVariables> extends React.Compo
     const mutationId = this.generateNewMutationId();
 
     return this.mutate(options)
-      .then(response => {
+      .then((response: ExecutionResult<TData>) => {
         this.onMutationCompleted(response, mutationId);
         return response;
       })
-      .catch(e => {
+      .catch((e: ApolloError) => {
         this.onMutationError(e, mutationId);
         if (!this.props.onError) throw e;
       });
--- a/src/Mutation.tsx
+++ b/src/Mutation.tsx
@@ -45,7 +45,7 @@ export declare type FetchResult<
 };
 
 export declare type MutationOptions<
-  TData = { [key: string]: any },
+  TData = Record<string, any>,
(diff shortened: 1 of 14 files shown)


src/Mutation.tsx
```diff
@@ -35,14 +35,19 @@ export declare type MutationUpdaterFn<
...
+  C = Record<string, any>,
+  E = Record<string, any>
+> = ExecutionResult<TData> & {
   extensions?: E;
   context?: C;
 };
 
-export declare type MutationOptions<TData = any, TVariables = OperationVariables> = {
+export declare type MutationOptions<
+  TData = { [key: string]: any },
```
benjamn: Any reason not to use `Record<string, any>` here (just for consistency)?

src/Query.tsx
```diff
@@ -220,7 +221,10 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
     }
 
     this.updateQuery(nextProps);
+
     if (nextProps.skip) return;
+
+    this.lastResult = this.queryObservable!.getLastResult();
```
benjamn: I'm curious if this is the best/only place we could capture the last result. In particular, since we're setting `this.lastResult = null` in `removeQuerySubscription`, and this line comes right before `this.startQuerySubscription()`, perhaps we should move the management of `this.lastResult` into `{start,remove}QuerySubscription`?

benjamn: I also see that `resubscribeToQuery` is calling `this.queryObservable!.getLastResult()` separately from this logic, so I wonder if `resubscribeToQuery` could be simplified now that we're tracking `this.lastResult`?

src/Mutation.tsx
hwillson: No reason at all - I'll change it, thanks!

src/Query.tsx
hwillson: The timing of capturing the last result is definitely tricky. I've tried it in a few different places, only to find out that it was either captured too soon or too late. Capturing it here addresses the timing issues, but moving it into `{start, remove}QuerySubscription` should work as well. I'll move it and we'll see - and I'll look into cleaning up `resubscribeToQuery` as well. Thanks!

hwillson: @benjamn I was able to move the storing of `lastResult` into `{start, remove}QuerySubscription`, but it took a bit of wrangling because `startQuerySubscription` is used when receiving new props and when the component first mounts. When it first mounts we want to ignore the `lastResult` no matter what (since this can cause problems if we mount a component, then unmount, then mount it again, and it had previous results that cause it to no longer render), so I tweaked the logic a bit to accommodate. This should be ready for another review - thanks!

src/Query.tsx
```diff
@@ -135,6 +135,7 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
 
   private hasMounted: boolean = false;
   private operation?: IDocumentDefinition;
+  private lastResult?: ApolloQueryResult<TData> | null;
```
benjamn:
```suggestion
  private lastResult: ApolloQueryResult<TData> | null = null;
```

PR #2788, opened 2019-02-08
Allow `MockedResponse` results to be returned from a function

hwillson:
The `result` property of a `MockedResponse` is currently restricted to being a pre-configured fixed value. Allowing results to be calculated/returned from a function instead of being pre-configured, opens up new test verification paths, that can be useful when using `MockedProvider`.

This PR adjusts `MockedResponse`’s `result` such that if a function is set for its value, that function will be fired and its result will be passed to the Observable matching the incoming mocked `request`.

Fixes https://github.com/apollographql/apollo-feature-requests/issues/84.

### DIFF
--- a/src/test-links.ts
+++ b/src/test-links.ts
@@ -95,7 +95,7 @@ export class MockLink extends ApolloLink {
         } else {
           if (result) {
             observer.next(
-              Object.prototype.toString.call(result) === '[object Function]'
+              typeof result === 'function'
                 ? (result as ResultFunction<FetchResult>)()
                 : result
             );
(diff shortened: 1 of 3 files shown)


src/test-links.ts
```diff
@@ -86,12 +88,18 @@ export class MockLink extends ApolloLink {
...
-    return new Observable<FetchResult>(observer => {
+    return new Observable(observer => {
       let timer = setTimeout(() => {
         if (error) {
           observer.error(error);
         } else {
-          if (result) observer.next(result);
+          if (result) {
+            observer.next(
+              Object.prototype.toString.call(result) === '[object Function]'
```
benjamn:
```suggestion
              typeof result === 'function'
```
should be perfectly safe here.

PR #2825, opened 2019-02-27
Allow resolvers to be passed into `MockedProvider` and set a default

hwillson:
After the changes in https://github.com/apollographql/apollo-client/pull/4499, when `MockedProvider` is currently used it doesn't have any `resolvers` set. This means `@client` directives are passed into the link chain. Since we don't currently allow people to modify the link chain that `MockedProvider` uses, it's safe to assume people won't want this behaviour, as they can't test things like `apollo-link-state` this way. This PR sets a default `resolvers` value of `{}` to enable AC 2.5's new local state handling, and opens the door for passing custom resolvers into the `MockedProvider` via props.

Helps with https://github.com/apollographql/fullstack-tutorial/pull/73.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,15 @@
 # Change log
 
+## 2.5.1
+
+### Bug Fixes
+
+- Make sure `MockedProvider` enables Apollo Client 2.5's local state handling,
+  and allow custom / mocked resolvers to be passed in as props, and used with
+  the created test `ApolloClient` instance.  <br/>
+  [@hwillson](https://github.com/hwillson) in [#2825](https://github.com/apollographql/react-apollo/pull/2825)
+
+
 ## 2.5.0
 
 ### Improvements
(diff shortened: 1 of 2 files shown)


Gongreg:
Hey @hwillson, after updating  from 2.5.0 to 2.5.1 I am now getting empty responses after I've mocked apollo provider and passed my own resolvers. I am not using apollo-link-state anymore.

2.5.0 version returns proper result but it warns in console that:

`Found @client directives in query but no client resolvers were specified. You can now pass apollo-link-state resolvers to the ApolloClient constructor.`

Also if I pass null to resolvers in v2.5.1 it also works but still gives the warning.

hwillson:
Interesting - thanks for the head up @Gongreg, I'll take a look.

hwillson:
@Gongreg - Let's track https://github.com/apollographql/react-apollo/pull/2825#issuecomment-468255909 in https://github.com/apollographql/apollo-client/issues/4520.

Issue #2743, opened 2019-01-20
[Ongoing] Reduce bundle sizes.


benjamn:
> Note: This issue tracks work that began in PR #2661.

In the same spirit as https://github.com/apollographql/apollo-client/issues/4324, the bundle footprint of the `react-apollo` package (currently [32.7KB minified, 8.7KB after gzip](https://bundlephobia.com/result?p=react-apollo@2.4.0)) could be, well, smaller.

This issue will serve as a place for discussing and coordinating bundle size improvements of any kind, including but not limited to the following techniques:
* Reimplementing surprisingly heavy dependencies locally, a la 4bac03af40690db68fa79e4fc7df944a7256d861.
* Making it easier to import specific parts of the `react-apollo` package, by supporting multiple entry points, e.g. `react-apollo/browser` and `react-apollo/server`. Note: this may be less important if we can rely on effective tree-shaking and ESM bundling.
* Transforming imports to be more specific using a compiler plugin, so that `import { SpecificThing } from "react-apollo"` becomes `import { SpecificThing } from "react-apollo/lib/specific-thing"`. If we end up pursuing this approach for `apollo-client`, we will no doubt reuse the same tools here. Note: because this would impose new build-time requirements, we should wait to try this approach until it seems like the only/best option.
* Somehow code-splitting the library, even though dynamic `import()` is not an option for libraries that must be consumed by bundlers with different runtime implementations of dynamic modules.
* Introducing a production bundle in addition to the development bundle, with more aggressive elimination of unnecessary code (e.g. error strings).
* Playing to the strengths of the advanced bundlers that our most bundle-size-conscious users are likely to be using.

Pull requests are more than welcome! Here's a list of PRs that have contributed (or promise to contribute) to reducing bundle sizes:
- [x] Changes released in `react-apollo@2.5.0`: #2661
- [x] Holistic application-level bundle measurement: #2839
- [ ] Reductions inspired by `examples/rollup`: #2842

sventschui:
@benjamn one thing about the bundle size are the graphql queries. I understand that apollo uses some parts of the queries (e.g. default values). Is there a way to throw away part of the queries on the client side and replace the queries with a hash (similar to automated persisted queries)?

I guess this would help a lot in practice, what do you think?

hwillson:
[...]

PR #2839, opened 2019-03-04
Add a basic example app bundled with Rollup.

benjamn:
The goal of this application is to get a precise, real-world measurement of the total bundle footprint of essential Apollo Client and React Apollo packages, using a modern bundler, including the weight of dependencies (excluding `react` and `react-dom`), but also taking into account the overlap between shared dependencies, which is not possible with per-package tools like https://bundlephobia.com/result?p=apollo-client@2.5.1.

If you run
```bash
cd react-apollo/examples/rollup
npm install
npm run analyze
```
in this application, you'll see a breakdown of everything in the bundle that we think should be attributed to using Apollo Client and React Apollo. The current total stands at about 143KB (minified but *not* compressed with gzip, since that's how `source-map-explorer` works). After gzip, the total is about 39KB.

Of that 143KB, `apollo-client` is the biggest single offender, at 39.5KB, followed closely by `graphql`, at 31.6KB. If we included `react-dom`, it would weigh in at 106KB, but there's not much we can do about that, so we don't bother measuring it. By contrast, Apollo can pick and choose what it imports from the `graphql` package, so the `graphql` package is worth measuring, even though it's not maintained by Apollo.

I welcome anyone and everyone to examine the measurement methodology we're using here, because early feedback about any inaccuracies will make this application much more useful for deciding where and how to invest future bundle reduction energy.

Likewise, I hope the analysis produced by this application will inspire grounded discussion of potential bundle reduction strategies, in addition to helping validate them.

### DIFF
--- a/examples/rollup/.gitignore
+++ b/examples/rollup/.gitignore
@@ -1,0 +1,3 @@
+node_modules
+*.min.js
+*.min.js.map
--- a/examples/rollup/index.html
+++ b/examples/rollup/index.html
@@ -1,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="app-with-react.min.js"></script>
+  </body>
+</html>
--- a/examples/rollup/main.js
+++ b/examples/rollup/main.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { HttpLink } from 'apollo-link-http';
+import { Query, ApolloProvider } from 'react-apollo';
+import gql from 'graphql-tag';
+
+const client = new ApolloClient({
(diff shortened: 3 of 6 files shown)


benjamn:
For the sake of comparison, if you go back by a minor version (`apollo-client@2.4.13`, `apollo-cache-inmemory@1.4.1`, `react-apollo@2.4.1`, and `apollo-link-state@0.4.2`), the total size of the bundle according to this analysis is **197KB**.

In other words, we've reduced the bundle size contribution of Apollo Client and React Apollo by **27%** since we started focusing our attention on this effort (see https://github.com/apollographql/react-apollo/issues/2743 and https://github.com/apollographql/apollo-client/issues/4324 for more details).

I'm not sure how best to share historical snapshots of this analysis, but here's an archive of the HTML output, along with the relevant `package.json` and `package-lock.json` files, which should be enough to reproduce my results: [comparison.zip](https://github.com/apollographql/react-apollo/files/2928665/comparison.zip)

lishine:
This app have errors, cannot manage to run.
What is the current status, what is the bundle size?

PR #2840, opened 2019-03-05
Update lastResult whenever data changes to ensure components re-rende…

Glennrs:
I've been unable to write a test case to reproduce this scenario, however, the logic is fairly simple to follow. I'm happy to be informed if I'm missing something and this PR is not required. Scenario is:

1) Component renders with value from cache, which sets lastResult.
2) Value of cache is set to another value. Component re-renders.
3) Value is set back to the original value, therefore shallowEquals does not pass since lastResult is fixed to the original render.

This PR ensures that lastResult is updated if set, when new data is available.

Fixes https://github.com/apollographql/react-apollo/issues/2857.
Fixes https://github.com/apollographql/react-apollo/issues/2887.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,10 @@
   component, after an error was received, variables were adjusted, and then
   the good data was fetched.  <br/>
   [@MerzDaniel](https://github.com/MerzDaniel) in [#2718](https://github.com/apollographql/react-apollo/pull/2718)
+- Fixed an issue that prevented `Query` component updates from firing (under
+  certain circumstances) due to the internal `lastResult` value (that's used
+  to help prevent unnecessary re-renders) not being updated.  <br/>
+  [@Glennrs](https://github.com/Glennrs) in [#2840](https://github.com/apollographql/react-apollo/pull/2840)
 
 ### Improvements
 
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -369,6 +369,9 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
         }
 
         initial = undefined;
+        if (this.lastResult) {
+          this.lastResult = this.queryObservable!.getLastResult();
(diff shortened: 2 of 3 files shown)


hwillson:
Thanks for submitting this PR @Glennrs. We can't accept this PR as it stands currently, as we have no way to verify that it's fixing a problem. A test showing what this fixes (to help avoid regressions), or a  small runnable reproduction that demonstrates how this is a problem, would definitely help. I'll close this for now, but to help push the discussion forward, I'd suggest opening a new issue with as many details as you can provide that shows how this is a problem. Once we've been able to confirm this is an issue, we'll then be ready to accept a PR. Thanks!

Glennrs:
The following Gist has a repro for the issue. You can see that after the Query component has its properties updated, it will set lastResult to true and subsequently always early out when the value is set back to true again (meaning it remains in a state where it renders false).

hwillson:
Re-opening as this issue has been demonstrated by the reproduction in https://github.com/apollographql/react-apollo/issues/2887. I'll add a regression test, then get this merged.

hwillson:
I've updated the original PR message to reference a few issues this will help fix, but I'm sure there are others that this will help with (either directly or indirectly), like https://github.com/apollographql/react-apollo/issues/1931.

yloeza:
@hwillson when are you guys planning to release this? is it going to be just a patch version?

mac-s-g:
also hoping to test this out!

mac-s-g:
update:
2.4.1 has other issues... mainly, the issue with data getting set to an empty object (resolved @ ^2.5).  so, i gotta move back to 2.5.4 and wait for the release.  loading states will be a little clunky til then.

maybe a good time to give a shout out to all the contributors on this project.
it's an AWESOME module. really amazing work.  thank you for creating and maintaining it!

PR #2856, opened 2019-03-09
Sync & Update versions of react packages

### LINKED ISSUE #2792: package.json devDependencies for React outdated?
**Intended outcome:**

An error is stated when trying to use React hooks:

> Hooks can only be called inside the body of a function component. 

I believe the cause may be react & react-dom devDependency mismatch.

**Actual outcome:**

Thanks to @ryancogswell for the catch.

The devDependencies in `package.json` have React listed as: 16.5.2
> "react": "16.5.2", "react-dom": "16.8.1"

This may be causing an issue with React hooks.

Open package.json

Possibly related issue: [https://github.com/facebook/react/issues/14823](url)

**Version**

- apollo-client@2.4.12
- react-apollo@2.4.1

sapkra:
This PR will update react including the react test renderer.

The reason is that react and react-dom should always have the same version but the bots are ignoring this rule and for some applications it will produce bugs e.g. broken react hooks.

Fixes #2792


sapkra:
I definitely need some help with fixing the breaking changes - can someone have a look? You can see some errors in the logs of the failed builds.

hwillson:
Thanks for this @sapkra. We're currently working on updating React Apollo to drop its use of deprecated React API elements. That work overlaps with the items you've listed here, so I'll close this off. Thanks again though - these items will be addressed and ready shortly.

sapkra:
Ok, nice to hear that.
For the future it would be great if someone of the apollo team can comment on those tickets so that the community is informed and will not invest time in things they are already WIP. 😉

hwillson:
Agreed @sapkra - it has been mentioned in a few other issue threads, but communication across many related issues is definitely challenging. We're always striving to improve communication though, so thanks for the reminder!

PR #2872, opened 2019-03-16
Create useApolloClient.ts

FredyC:
### Checklist:

Quick fix for a leftover from context refactor so we can use it with hooks. I suppose tests are not needed for such a simple case?


FredyC:
Well, I am stumped now. With old `@types/react` this will never pass the tests. What is the preferred way of dealing with it? Just `@ts-ignore` or rather update types?

hwillson:
@FredyC I'll be pushing updates shortly that update `@types/react`.

hwillson:
@FredyC `master` now includes the most recent version of `@types/react`.

FredyC:
@hwillson Thanks. Looks like all is green, ready to merge :)

hwillson:
Thanks for this @FredyC. Since this requires `useContext`, we're going to merge it into the `release-3.0.0` branch shortly.

FredyC:
@hwillson That's not correct...It should be in the master branch, that's why there is that check in case someone tries to use it without React 16.8.

hwillson:
Thanks for the feedback @FredyC. Unfortunately, we don't want to introduce internal Hook use into `react-apollo` 2.x. The React 16.8 dependency in `react-apollo` 2.5 is a `devDependency`; we're actually supporting as far back as `react` 15 since it's a `peerDependency` (yes, we're supporting a 3 year old version of `react` :scream:). We don't want to introduce further gated checks for React functionality (like checking to make sure people can use `useContext`) in `react-apollo` 2.x, since 3.0 is going to land very quickly, and will have its `react` peer dep set to 16.8. Gated  checks for functionality are a pain to manage, and as an example, the check that was there was incorrect. It was checking to see if `ApolloContext` existed, which meant `React.createContext` could be used, which was released in `react` 16.3. Someone using `react` >= 16.3 but < 16.8 would have passed the check, only to have the call to `useContext` fail.

For now people who want to use the new context API in `react-apollo` 2.x (after the next patch release) will be able to using the `ApolloConsumer` component.

notgiorgi:
the workaround for 2.x:

You need to wrap the `ApolloConsumer` into your own context:
```js
// MyApolloContext.js
const MyApolloContext = React.createContext()

export const MyApolloContextProvider = ({ children }) => (
  <ApolloConsumer>
    {client => <MyApolloContext.Provider value={client}>{children}</MyApolloContext.Provider>}
  </ApolloConsumer>
)

export const useApolloClient = () => {
  return useContext(MyApolloContext)
}


// app.js
  <MyApolloContextProvider>
    <MyComponent />
  </MyApolloContextProvider>
<ApolloProvider>


// MyComponent.js
const MyComponent = () => {
    const client = useApolloClient()
    // ...
}
```

TidyIQ:
@notgiorgi Your example doesn't work for me.

First, I get an error as no default value is provided to React.createContext(), then I get an error when I attempt to use client.query() in MyComponent saying `Property 'query' does not exist on type '{}'`

PR #2873, opened 2019-03-16
Full dep/devDep updates to latest versions

hwillson:
This PR updates all `dependencies` / `devDependencies` to their latest versions, and includes code (mostly Typescript) changes required by the new versions.

One thing to note - these changes only include the bare minimum needed to make `walkTree` compile after the dep updates, and work exactly like it did before (bugs and all). Re-capping from https://github.com/apollographql/react-apollo/commit/98adf9261fccb1c0123566ea8f8b712928f27b38):

> `walkTree` is no longer used by `react-apollo` (see #2533), but
since it was previously exported and made available externally,
it is still available (albeit deprecated) until `react-apollo`
3.0. Unfortunately, the current `walkTree` implementation does not
fully work with React's new Context API. It can be updated to work
with modern versions of React, but the entire point of #2533 is
to avoid spending anymore time working on `walkTree`, since its
implementation relies on React internals (and is very brittle).
Since external use of `walkTree` is very minimal, and since there
have been no complaints of `walkTree` not working with React's
new Context API since it has been made a standalone module
(these problems were only noticed when `walkTree` was being used
by `react-apollo`), I think we should leave `walkTree` as is.
> 
> This commit makes sure `walkTree` passes Typescript compilation,
but disables failing tests caused by render tree differences in
newer versions of React.

I don't think we should spend anymore time on `walkTree` (and will add the above to the `CHANGELOG`). I know in theory we should make sure `walkTree` works fully with current day React until it's removed (in `react-apollo` 3.0), but given that its use is now quite minimal, and it still works just like it did before (which is broken with current day React), I really feel spending further time on it is not productive. I also don't want us to be held up on updating the rest of `react-apollo` to work with modern React versions, until we hit 3.0 and can remove `walkTree`. Definitely let me know if anyone disagrees.

Thanks!


src/mutation-hoc.tsx
```diff
@@ -31,7 +31,7 @@ export function withMutation<
   if (typeof mapPropsToOptions !== 'function') mapPropsToOptions = () => options as MutationOpts;
 
   return (
-    WrappedComponent: React.ComponentType<TChildProps & TProps>,
+    WrappedComponent: React.ComponentType<TProps & TChildProps>,
```
hwillson: Switching the `TProps & TChildProps` order around has no technical merit - it just bugged me!

PR #2889, opened 2019-03-21
Provide shouldInvalidatePreviousData prop

### LINKED ISSUE #2202: When variables change, data is the previous result in Query Component
I am using the "cache-and-network" policy.

**Intended outcome:**
Rerender Query with different variables. Data is {} until new request is resolved from cache or network.

**Actual outcome:**
Rerender Query with different variables. Returns previous data until next query is resolved without error.

https://stackblitz.com/edit/react-snlwzf
change the input and hit submit.

**Version**

* apollo-client@<version number>
* react-apollo@<version number>

- [x] has-reproduction
- [ ] feature
- [ ] docs
- [ ] blocking
- [ ] good first issue

**Backgroud**
I only want to display a loading state if I don't have a response from the cache or network, but I do want to show loading state when variables change and I don't have a response from cache. Right now it is impossible to tell if data is the response from cache or if it is the previous result. Loading property is no help because it is allows true until the network request is resolved.

jcready:
Fixes #2202

### Checklist:

### DIFF
--- a/src/Query.tsx
+++ b/src/Query.tsx
@@ -94,6 +94,7 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
   static propTypes = {
     client: PropTypes.object,
     children: PropTypes.func.isRequired,
+    shouldInvalidatePreviousData: PropTypes.func,
     fetchPolicy: PropTypes.string,
     notifyOnNetworkStatusChange: PropTypes.bool,
     onCompleted: PropTypes.func,
@@ -202,6 +203,10 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
       this.queryObservable = null;
       this.previousData = {};
       this.updateQuery(nextProps);
+    } else if (typeof nextProps.shouldInvalidatePreviousData === 'function') {
+      if (nextProps.shouldInvalidatePreviousData(nextProps.variables, this.props.variables)) {
+        this.previousData = {};
+      }
     }
 
     if (this.props.query !== nextProps.query) {


hwillson:
Thanks for this PR @jcready! I'm wondering if we're over-complicating things here a bit. What if we just added something like a `clearPreviousDataOnError` prop. When it's `true`, we make sure `previousData` is emptied. This would simplify the RA codebase, and if an application really only wants to disable `previousData` based on previous / current variable differences, they can track variables and do the comparison outside of the `Query` component, and just pass the boolean result in via `clearPreviousDataOnError`.

src/utils/shallowEqual.ts
```diff
@@ -7,11 +8,21 @@ function is(x: any, y: any) {
...
+  const keys = Object.keys(objA);
+
+  if (keys.length !== Object.keys(objB).length) {
+    return false;
+  }
+
+  return keys.every(key => hasOwnProperty.call(objB, key) && is(objA[key], objB[key]));
+}
+
+export function shallowEqualSansFirst(objA: any, objB: any) {
```
hwillson: @jcready Can you explain why this is necessary?

jcready: I didn't want to re-write all the tests that depend on this functionality and it does seem like `first` is typically only used for pagination which means you generally wouldn't want to invalidate the previous data if it changed.

jcready:
@hwillson `clearPreviousDataOnLoad` makes it sound like you would only clear the previous data once the network call finished. I think many people want the previous data cleared the moment their `variables` change. Additionally I know that many people wanted this to be the default behavior (e.g. pagination should be the exception).

> [...] if an application really only wants to disable `previousData` based on previous / current variable differences, they can track variables and do the comparison outside of the `Query` component, and just pass the boolean result in via `clearPreviousDataOnLoad`.

To me that sounds like it would make things more difficult for the consumer since they would then have to keep track of the previous variables themselves. The `Query` component's `componentWillReceiveProps` already has access to both and can easily provide them to a passed in function.

hwillson:
Thanks for the feedback @jcready. Regarding making this the default behavior, I agree, but we're kinda of stuck with things the way they are now (due to backwards compatibility issues). You touched on part of the bigger issue when you mentioned `componentWillReceiveProps` - we're in the process of moving away from React lifecycle methods in `react-apollo` 3.0. Another part of the issue I see here is that people will likely want similar behavior when they change other things, like changing the `query` for example. Hmmm ... thinking ...

Titozzz:
@hwillson This is still an issue, and after the 3.1 release, it is not fixed. By the way, you might want to reopen #2202.
If you want a CodeSandbox reproducing this: https://codesandbox.io/s/context-vs-queries-f2079.

iamswain25:
is it getting merged?

PR #2892, opened 2019-03-22
Release 3.0.0

hwillson:
This PR (and the `release-3.0.0` branch) will be used to track all `react-apollo` 3.0.0 work. The main focus of React Apollo 3.0 is to better align with current day React, and do some spring cleaning on the codebase. This means:

- Remove all `react-apollo` deprecated API elements
- Stop using deprecated React API's
- Provide React Hooks integration

~~The current (non-deprecated) `react-apollo` API will stay the same, but this release will include breaking changes (mostly around dependencies). For example, use of >= React 16.8 will be required, and we're dropping Preact support.~~

As we progress on this, the number of breaking changes are growing. The core functionality of the existing HOC and Component approaches is not changing however, and there will be a super easy migration path (that will likely only involve renaming your imports).


hwillson:
The CJS and UMD tests will be left failing here for a bit; the testing infrastructure is being revamped as part of this work (e.g. `MockedProvider`, `MockLink`, `MockSubscriptionLink`, etc. are being moved to a new package).

mattfwood:
Is there any help needed with react hooks integration or writing docs for react-apollo hooks? I’ve been using this library a lot and would be really interested in finding a way to contribute

FredyC:
@mattfwood Please have a look at the https://github.com/trojanowski/react-apollo-hooks/pull/93
It's lacking documentation, so it would be great if you are willing to write one.

sessionboy:
Hi guys.
When will v3 be released?

hwillson:
@sessionboy We don't have a release date to share yet, but RA 3.0 is a priority. We're taking this opportunity to do a lot of housecleaning / preparing for the future, so it will take some time. That being said, we're talking weeks, not years 🙂.

orrybaram:
@FredyC will the new hooks api be based on https://github.com/trojanowski/react-apollo-hooks?

FredyC:
@orrybaram I can't speak for the Apollo team, but I think for the sake of easy adoption the API should be similar. I don't even think it's possible to invent much in there, depends on how Suspense for data fetching turns out I guess.

Personally, I wouldn't worry about it. I've already migrated two production apps to a hooks version. The fact is that the library is not going anywhere so even if Apollo makes some hugely incompatible API, you can always stick to that library that will work no matter what.

hwillson:
The public Hooks API will be very similar - `useQuery`, `useMutation` and `useSubscription`. So if you're already using `react-apollo-hooks` you'll find the transition (if you want to switch) very painless. That being said, the internals will be different. The reason for this is that React Apollo 3.0 will maintain support for its `graphql` HOC, Components, and introduce Hooks support. To achieve this, we've converted the project to a monorepo structure, and will be splitting these mains areas of functionality up into their own packages. The core data management code is being abstracted out, and will be shared by each of these different approaches behind the scenes. We've also acquired the `@apollo` NPM organization, which we're going to start using when React Apollo 3.0 is released. You will be able to include only the parts of React Apollo you want to use, by adding the package that you want to work with (to help with bundle sizes):

- `@apollo/react-components`
- `@apollo/react-hooks`
- `@apollo/react-hoc`
- `@apollo/react-testing` 

In addition, we'll be providing a new top level "this is our preferred way of working with Apollo, GraphQL and React" entry point package, that will take care of wiring all dependencies up we think are a great starting point for most projects moving forward. That package will likely be published as `@apollo/react`, and will likely promote Hooks as the way forward.

One caveat - none of this work will be touching React Suspense yet. If you want to use Hooks with Suspense, then `react-apollo-hooks` is the way to go (since it has experimental support). We're going to wait until the React team has nailed down the Suspense + Data approach, before diving into it.

trojanowski:
The plan is to deprecate react-apollo-hooks after there is official support for hooks in react-apollo (and recommend official hooks instead). There is more info about this in this thread: <https://github.com/trojanowski/react-apollo-hooks/issues/40>. Until then, I'm going to maintain that library. We're going to release a new version soon: <https://github.com/trojanowski/react-apollo-hooks/issues/138>.

> The public Hooks API will be very similar - useQuery, useMutation and useSubscription.

Have you already thought what the exact API for `useMutation` would look like? We're going to introduce a breaking change here: <https://github.com/trojanowski/react-apollo-hooks/pull/93>, but I'd prefer the migration to the official hooks to be as painless as possible.

> The reason for this is that React Apollo 3.0 will maintain support for its graphql HOC, Components, and introduce Hooks support. To achieve this, we've converted the project to a monorepo structure, and will be splitting these mains areas of functionality up into their own packages. The core data management code is being abstracted out, and will be shared by each of these different approaches behind the scenes. 

> You will be able to include only the parts of React Apollo you want to use, by adding the package that you want to work with (to help with bundle sizes):

Have you considered creating render-props and HOC versions based on the hooks version? E.g. in theory the `<Query />` component could be implemented this way:

```
const Query = ({ children, query, ...rest }) => children(useQuery(query, rest));
```

benseitz:
> Have you considered creating render-props and HOC versions based on the hooks version? E.g. in theory the <Query /> component could be implemented this way:
> 
> ```javascript
> const Query = ({ children, query, ...rest }) => children(useQuery(query, rest));
> ```

If the peerDependency will be `react >=16.8` anyway, this seems the most elegant solution to me.
And the HOCs can be achieved in a similar way.

packages/common/config/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../config/tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "../lib"
+  },
+  "include": ["../src/**/*"],
+  "exclude": ["../src/**/__tests__/**/*", "../src/**/__mocks__/**/*"]
+}
```
benjamn: Shouldn't `packages/common` have a `tsconfig.json` file at its root, rather than nested in `packages/common/config`? Seems awkward to have so many `../`s here, and the `tsconfig.json` file is what defines the root directory of a TypeScript project (or so I thought).

packages/components/config/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../config/tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "../lib"
+  },
+  "include": ["../src/**/*"],
+  "exclude": ["../src/**/__tests__/**/*", "../src/**/__mocks__/**/*"]
+}
```
benjamn: Same comment: shouldn't this `tsconfig.json` file live in the root `packages/components` directory?

examples/rollup/main.js
```diff
@@ -1,50 +0,0 @@
...
-        "logged " + (data.isLoggedIn ? "in" : "out"),
-        <br/>,
-        <button onClick={toggle}>
-          log {data.isLoggedIn ? "out" : "in"}
-        </button>,
-      ]}
-    </Query>
-  </ApolloProvider>,
-  document.getElementById('root'),
-);
```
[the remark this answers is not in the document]
hwillson: @benjamn Absolutely (and we'll figure out a plan of attack for keeping them up to date separately).

packages/common/src/parser/parser.js
```diff
@@ -0,0 +1,57 @@
+import { invariant } from 'ts-invariant';
+export var DocumentType;
+(function (DocumentType) {
+    DocumentType[DocumentType["Query"] = 0] = "Query";
+    DocumentType[DocumentType["Mutation"] = 1] = "Mutation";
+    DocumentType[DocumentType["Subscription"] = 2] = "Subscription";
+})(DocumentType || (DocumentType = {}));
```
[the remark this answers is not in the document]
hwillson: @benjamn We shouldn't be (I missed this in https://github.com/apollographql/react-apollo/pull/2892/commits/5315dd1cc683ad01724e8b7aa6a51c81a801d41f) - thanks!

packages/common/config/tsconfig.json
hwillson: @benjamn I like the idea of keeping all project config related files in a common location, outside of the project root, hence the `/config` directory (which houses the Jest, Rollup, TS, etc. configs). Yes, `tsconfig` defines the root, which is why the `include`s / `exclude`'s are overridden. I'm not overly attached to this approach though, so if you'd prefer `tsconfig` in the root, then no problemo!

packages/components/config/tsconfig.json
hwillson: See above comment.

hwillson:
@trojanowski Sorry for the delay in responding; I think the API outlined in https://github.com/trojanowski/react-apollo-hooks/pull/93 (and https://github.com/trojanowski/react-apollo-hooks/pull/135) looks good! We'll plan for the same approach. And as for having the render props and HOC versions use the Hooks version behind the scenes, that's exactly what we're planning. We follow a similar approach right now with the HOC using the RA components behind the scenes, and it has worked well. Thanks!

mptorz:
@trojanowski @hwillson Do you have any estimates on when would hook support be released? How hard is it going to be to switch from react-apollo-hooks to official hooks support? I am not sure whether I should wait for the official release or start using the repo by @trojanowski

FredyC:
@mptorz Check https://github.com/apollographql/react-apollo/pull/2892#issuecomment-486205832

blakedietz:
Regarding api changes is there any help needed in making a codemod? I can ptich in.

jrop:
#2539 
Just referencing the above issue so that it links here.

Dattaya:
btw, latest alpha (`3.0.0-alpha.2`) cannot be installed neither in the normal Windows cmd: `'cp' is not recognized as an internal or external command,` nor in bash provided by git: `cp: cannot create regular file './node_modules/enzyme-adapter-react-16/build/': No such file or directory`

hwillson:
@Dattaya That `3.0.0-alpha.2` release isn't for this work; it was for testing some initial React Apollo 3.0 ideas about a year ago. I've removed it - sorry for the confusion! A beta release for this PR will be available shortly. Thanks!

SimenB:
Congrats on the beta release! 🎉 

I noticed when testing it out that `useQuery` is set to return `QueryResult<any, any>`: https://runpkg.com/?@apollo/react-hooks@0.1.0-beta.2/lib/useQuery.d.ts. Is there any (hah) particular reason for that? The code sandbox demo (https://codesandbox.io/s/github/peggyrayzis/redux-to-graphql/tree/master/client) does `const res: QueryResult<TData, TVariables> = useQuery()` to get proper types - wouldn't it be better if `useQuery` accepted the generics and defaulted to `any`, like the current `Query` component (and `react-apollo-hooks`) does?

FredyC:
I am wondering where is the proper channel to talk about issues regarding the beta. I mean NPM points to this repo, but this PR does not even contain the code for hooks and I only found a branch with some stubs. Would be a shame to keep piling up any discussion in this single PR.

djirdehh:
@SimenB As per Peggy's tweet (https://twitter.com/peggyrayzis/status/1128301517235159041) - the type defs are to be improved upon - which could be as to why the `useQuery` hook doesn't appropriately accept the generic types (yet).

I second @FredyC's comment. It would really be great if there was an open channel where we can discuss the library while still in the beta stage 🙏 .

hwillson:
Hi all - the types work wasn't quite done before Peggy's demo at Apollo Day NY, so gratuitous use of `any` was dropped in 🙂. Everything will be ready and dropped into this PR very soon. We aren't intentionally working on things in a secret manner - we've been a bit tied up in other areas, and didn't want to release partially working code. Once we get the full code dropped in, we'll be leaving a good chunk of time for review, feedback, suggestions, etc. before cutting an RC. Thanks for your patience!

FezVrasta:
@hwillson is Flow type support something you'd consider to include in this release?

hwillson:
@FezVrasta not at this time, sorry. We used to support Flow but the types quickly became out of date due to Flow losing the popularity race to TS (and because we don't use Flow internally). We've standardized on Typescript (and TS alone) across all Apollo repos.

mattiasewers:
@hwillson Does apollo with hooks not support ssr yet? I tried to set it up with next.js with no luck

FezVrasta:
@hwillson does that mean you wouldn't even accept external contributions?

hwillson:
@mattiasewers SSR support is coming for sure.

@FezVrasta Correct, we won't accept external Flow contributions. Again, we tried in the past, but it didn't work out (and the people who originally maintained the Flow types moved on). We're always looking for ways to reduce repo maintenance overhead, and dropping support for Flow was an easy win. Sorry!

hwillson:
@mattiasewers Hooks SSR support is now available in `@apollo/react-hooks@0.1.0-beta.6`.

packages/hooks/src/useApolloClient.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { invariant } from 'ts-invariant';
+import { getApolloContext } from '@apollo/react-common';
+
+export function useApolloClient() {
+  const { client } = React.useContext(getApolloContext());
+  invariant(
+    !client,
```
djirdehh: 👋 Unsolicited comment - so apologies in advance.

I believe the logical not operator here is causing the `useApolloClient` hook in release `@apollo/react-hooks@0.1.0-beta.6` to throw an error and not work. The `invariant` function itself places the `!` on the condition applied and seems to be used [appropriately elsewhere](https://github.com/apollographql/react-apollo/blob/15dea607a6caf71c0902c788fba43b323aabe12f/packages/common/src/context/ApolloProvider.tsx#L25).

Thanks for all the work you've done thus far! 🙏

packages/hooks/src/data/OperationData.ts
```diff
@@ -0,0 +1,114 @@
...
+    this._client = newClient;
+  }
+
+  public abstract execute(...args: any): any;
+  public abstract afterExecute(...args: any): () => void;
+
+  protected abstract cleanup(): void;
+
+  protected unmount() {
+    this.isMounted = false;
```
lucasecdb: another unsolicited comment 😅but aren't you missing calling `cleanup` here?

I'm getting a warning when unmounting any component using the hook.

packages/hooks/src/useApolloClient.ts
SimenB: also please do `return client!;` or something as it's currently typed as `ApolloClient | undefined` 🙂

hwillson:
Thanks very much for the comments @djirdehh @SimenB @lucasecdb - more changes are coming very soon (along with a full Hooks test suite that will help prevent these issues). I removed the WIP on this PR to show that we're getting close, but I should probably add a `here be dragons` label 🙂. Please keep the feedback coming though - all great stuff! 🙇

packages/hooks/src/data/OperationData.ts
hwillson: @lucasecdb Can you share the exact warning you're seeing?

packages/hooks/src/useApolloClient.ts
hwillson: @djirdehh @SimenB these should be fixed in `@apollo/react-hooks@0.1.0-beta.7`. Thanks!

packages/hooks/src/data/OperationData.ts
lucasecdb: oh, so sorry, I saw the `Query` component in my component stack and though the error was from it. Apparently it's happening because of this custom hook I've created:

```ts
const useLoading = (loading: boolean) => {
  const [mutate] = useMutation<LoadingMutation, LoadingMutationVariables>(
    LOADING_MUTATION
  )

  useEffect(() => {
    mutate({ variables: { loading } })
  }, [loading, mutate])
}
```

and this is the error I'm seeing:

image

it's still weird that I'm getting this warning, because my component doesn't use any state, it only uses the `graphql` hoc and the hook I've pasted above.

lucasecdb: So, according to the stack trace of the warning above, the error is coming from the `MutationData.updateResult` function

image

packages/hooks/src/types.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/* Query types */
+
+export interface QueryOptions<TData = any, TVariables = OperationVariables>
+  extends QueryFunctionOptions<TData, TVariables> {
+  children?: (result: QueryResult<TData, TVariables>) => ReactNode;
+  query: DocumentNode;
+}
+
+export interface QueryHookOptions<TData = any, TVariables = OperationVariables>
```
SimenB: Might be on purpose, but this does not match the signature in `react-apollo-hooks`: https://github.com/trojanowski/react-apollo-hooks/blob/e590e8f21f1ae4871d641d9681f7932433c8daca/src/useQuery.ts#L35-L45

This means that https://graphql-code-generator.com generates the wrong code. Not sure where the fix should go though 🙂 

/cc @trojanowski

rajington: link to the issue for that lib: https://github.com/dotansimha/graphql-code-generator/issues/1863

but it seems like they'll swap based on the `importHooksFrom` signature

sessionboy:
Hi!, How to manage state in v3 ? I can't find any api or example about managing state....

FezVrasta:
Is `MockedProvider` supported with `@apollo/react-hooks`?

FezVrasta:
If anyone is looking for a temporary solution for `MockedProvider` + `@apollo/react-hooks`, this should work:

```
// @flow
import React, { type Node } from 'react';
import { InMemoryCache } from 'apollo-cache-inmemory';
import { ApolloClient } from 'apollo-client';
import { MockLink } from 'react-apollo/test-links';
import { ApolloProvider as ApolloHooksProvider } from '@apollo/react-hooks';

export function createClient({ addTypename, mocks }) {
  return new ApolloClient({
    cache: new InMemoryCache({ addTypename }),
    link: new MockLink(mocks),
  });
}

export const MockedProvider = ({
  mocks,
  addTypename = true,
  children,
}: {
  mocks: Array<any>,
  addTypename: boolean,
  children: Node,
}) => {
  const client = createClient({ addTypename, mocks });
  return (
      <ApolloHooksProvider client={client}>{children}</ApolloHooksProvider>
    </ApolloProvider>
  );
};
```

adamsoffer:
Hey guys - super excited for 3.0! I'm trying to apollo hooks working in a beta version of [`next-apollo`](https://www.npmjs.com/package/next-apollo) but I'm getting the following error: 

`Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.`

Am I using `ApolloProvider` correctly? https://github.com/adamsoffer/next-apollo/blob/apollo-hooks/src/withData.js#L3

jacquesnf:
> Hey guys - super excited for 3.0! I'm trying to get apollo hooks working in a beta version of [`next-apollo`](https://www.npmjs.com/package/next-apollo) but I'm getting the following error:
> 
> `Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.`
> 
> Am I using `ApolloProvider` correctly? https://github.com/adamsoffer/next-apollo/blob/apollo-hooks/src/withData.js#L3
> 
> `npm install --save next-apollo@beta` if interested in reproducing in your next project.

@adamsoffer, are you importing `ApolloProvider` from `react-apollo` or from `@apollo/react-hooks`? I ran into the same issue and it was because I was importing from `react-apollo`. 

I think the context value provided by `ApolloProvider` exported from `@apollo/react-hooks` contains some extra data that is needed for the hooks to work.

So try:
```javascript
import {ApolloProvider} from '@apollo/react-hooks'
```

adamsoffer:
> @adamsoffer, are you importing `ApolloProvider` from `react-apollo` or from `@apollo/react-hooks`?

Thanks @jacquesnf. I am indeed importing from `@apollo/react-hooks`. That's why I'm a bit confused. See [here](https://github.com/adamsoffer/next-apollo/blob/apollo-hooks/src/withData.js#L3).

adamsoffer:
Ah nevermind - seems to be working.

packages/hooks/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "author": "opensource@apollographql.com",
+  "keywords": [
+    "apollo",
+    "graphql",
+    "react",
+    "hooks"
+  ],
+  "license": "MIT",
+  "main": "./lib/react-hooks.cjs.js",
+  "module": "./lib/react-hooks.esm.js",
```
benjamn: Since the `@apollo/react-hooks` package is where we now define `getDataFromTree`, I think we need to apply my commit 58c9f6e6b9c77f26795e47a485252fd875eb0bdb to this `package.json` file, so that React Native will not be confused by the `react-dom/server` dependency.

packages/hooks/src/data/OperationData.ts
```diff
@@ -0,0 +1,114 @@
...
+    this._isMounted = mounted;
+  }
+
+  get options(): CommonOptions<TOptions> {
+    return this._options;
+  }
+
+  set options(newOptions: CommonOptions<TOptions>) {
+    if (!isEqual(this.options, newOptions)) {
+      this.previousOptions = this.options;
```
benjamn: Even this logic might be better implemented by an explicit `getOptions`/`setOptions` method pair, rather than a getter and a setter.

jest.cjs.config.js
```diff
@@ -1,14 +0,0 @@
-const { jest } = require('./package.json');
```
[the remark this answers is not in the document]
hwillson: Right - I was a bit torn on the value added by doing so (and the extra pain it introduced), but I'll find a way to get it back in.

packages/common/config/rollup.config.js
```diff
@@ -0,0 +1,96 @@
...
+};
+
+function external(id) {
+  return Object.prototype.hasOwnProperty.call(globals, id);
+}
+
+const SRC_DIR = './src';
+const LIB_DIR = './lib';
+
+export default [
```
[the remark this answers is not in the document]
hwillson: I think so - I'm going to work on getting your rollup removal changes in later today, and will tackle this at the same time. Thanks!

packages/common/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "repository": {
+    "type": "git",
+    "url": "apollographql/react-apollo"
+  },
+  "sideEffects": false,
+  "scripts": {
+    "clean": "rm -Rf ./node_modules ./lib/* ./meta/bundlesize/* ./meta/coverage/*",
+    "prepare": "npm run build",
+    "prebuild": "npm run clean",
+    "build": "npx -p typescript tsc -p ./config",
```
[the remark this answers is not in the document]
hwillson: Great - that's what I had originally, but decided to let npx handle things (instead of seeing the relative references). I'm cool either way though, so this is great. Thanks!

packages/components/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "sideEffects": false,
+  "scripts": {
+    "clean": "rm -Rf ./node_modules ./lib/* ./meta/bundlesize/* ./meta/coverage/*",
+    "prepare": "npm run build",
+    "prebuild": "npm run clean",
+    "build": "npx -p typescript tsc -p ./config",
+    "postbuild": "npx rollup -c ./config/rollup.config.js",
+    "watch": "node ../../node_modules/tsc-watch/lib/tsc-watch.js --onSuccess \"npm run postbuild\" -p ./config",
+    "predeploy": "npm run build",
+    "deploy": "npm publish --tag beta"
```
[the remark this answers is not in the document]
hwillson: For sure, I'll add something in. I've made using `jest --testRegex ...` muscle memory, so I'm all for freeing that memory space up 🙂.

packages/testing/src/utils/catchAsyncError.ts
```diff
@@ -0,0 +1,7 @@
+export const catchAsyncError = (done: jest.DoneCallback, cb: () => void) => {
```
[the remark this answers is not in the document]
hwillson: Let's drop it!

packages/testing/src/utils/wrap.ts
```diff
@@ -0,0 +1,10 @@
+export const wrap = <TArgs>(
```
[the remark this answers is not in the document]
hwillson: Good question - I'll take a look.

packages/hooks/package.json
hwillson: Oh right! I'll get that in - thanks!

benjamn:
@hwillson Ok, I'm done pushing commits (and reviewing) for now, so you can take it from here!

Dattaya:
`getDataFromTree`/`getMarkupFromTree` for some reason returns an empty string on the server (`const content = await getDataFromTree(InitialView)`), does anyone else experience this? I'll try to create a repro case if it's still an issue in the next beta/rc.

FezVrasta:
I just wanted to report that `stopPolling` doesn't seem to work properly, or anyway something related to the polling mechanism.


```jsx
const { loading, error, data, stopPolling } = useQuery(QUERY, {
  variables: { id: documentId },
  pollInterval: 2500,
  skip: shouldSkip,
});

if (data && data.document.status === 'COMPLETED' || error) {
  stopPolling();
}
```

this will not stop polling, but if I change it to:

```jsx
const { loading, error, data, stopPolling } = useQuery(QUERY, {
  variables: { id: documentId },
  pollInterval: shouldSkip ? 0 : 2500,
  skip: shouldSkip,
});

if (data && data.document.status === 'COMPLETED' || error) {
  stopPolling();
}
```

it works properly, so probably there's something wrong with the combination of `skip` and `pollInterval`?

packages/components/package.json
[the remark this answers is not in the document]
hwillson: I have changes coming for this that allow `npm test` in the parent for all tests, or `npm test` in child packages for self contained testing, all sharing the same central jest config.

packages/testing/src/utils/flowRight.ts
```diff
@@ -0,0 +1,11 @@
+export function compose(...funcs: Function[]) {
```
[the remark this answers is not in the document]
hwillson: Turns out we are using `compose` for one single internal HOC test. I'll move this function into the test as a helper function, and drop the export to keep it fully internal.

packages/testing/src/utils/wrap.ts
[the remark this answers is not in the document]
hwillson: I'll go with removing `catchAsyncError` and `wrap`. In both cases simple try-catches work well as an alternative, and help make test logic easier to follow.

Negan1911:
Latest 3.0.0 release fails because `this.currentObservable.query!.getCurrentResult()` is not available:
Screen Shot 2019-06-11 at 3 24 08 AM
I made [this commit](https://github.com/Negan1911/react-apollo/commit/2a68bf6452d7efd8579009956085034ec2a1cd62)

which actually fixes it, please review and include it on this PR if necesary, proof of app working again:

Screen Shot 2019-06-11 at 3 29 27 AM

packages/common/config/buildJestModuleMap.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (existingModuleMap, library) => ({
+  ...existingModuleMap,
+  '\\.\\./ApolloProvider': library,
+  '\\.\\./ApolloConsumer': library,
+  '\\.\\./ApolloContext': library
```
hwillson: I'm not super happy with this approach, but I haven't found a better workaround (without sacrificing in other areas). Here we're basically overriding relative calls to source included within a package, to hijack the resolution of the source to instead pull from either the CJS or UMD bundle. While this works, it means that anytime we want to reference a file using a relative path in a test file, we have to make sure we've added a matching line in the `buildJestModuleMap.js` file to properly CJS/UMD test it. Right now this isn't that big of a deal since the number of files we have to hijack is pretty small, and isn't likely going to change anytime soon. But ideally we'd have a more generic way to intercept and redirect the lookup for all relative files under test.

We could consider always including `src` or some other common prefix in imports, then intercept based on `src` being in the lookup reference (like current day React Apollo [is doing](https://github.com/apollographql/react-apollo/blob/0adeb79e886003f57d0f77df875923cc75b4c332/jest.cjs.config.js#L3-L9)). Due to the fact that RA 3 leverages `__tests__` directories within `src` however, this would mean imports like

```js
import { graphql } from '../../src/graphql';
```

which just look strange, given that 

```js
import { graphql } from '../graphql';
```

is what most people would expect to see/use. I'm definitely open to better suggestions here.

hwillson: We're going with a different (more dynamic) approach for this, in https://github.com/apollographql/react-apollo/pull/2892/commits/6cf9f28409edf72d14a47fd5ab2865ee00c45809.

bummey:
Small question - is it possible to skip a subscription hook? I looked through the types but I couldn't find anything about it.

hwillson:
@bummey I was about to say no as I didn't remember seeing any existing subscription `skip` logic in the codebase when I was refactoring things, but I just double checked to make sure. Turns out in React Apollo 2 we actually support (and expose via types) a `skip` param in the `graphql` HOC for subscriptions, but behind the scenes we don't actually do anything with it 🤦‍♂️. I've made a note of this and we'll see what we can do about getting this into the RA 3 release (we're getting pretty close to feature locking things, so this might have to wait until a minor/patch release).

packages/hoc/src/__tests__/queries/api.test.tsx
```diff
@@ -188,13 +188,15 @@ describe('[queries] api', () => {
...
-                wrap(done, (result: any) => {
+              .then((result: any) => {
+                try {
                   expect(stripSymbols(result.data.allPeople.people)).toEqual(
                     data1.allPeople.people
                   );
-                })
-              );
+                } catch (error) {
+                  done(error);
```
benjamn: Is there any difference between `done(error)` and `done.fail(error)`?

packages/components/package.json
```diff
@@ -29,8 +29,8 @@
     "deploy": "npm publish --tag beta",
     "test": "../../node_modules/jest/bin/jest.js --config ../../config/jest.config.js --testPathPattern packages/components",
     "test:watch": "../../node_modules/jest/bin/jest.js --config ../../config/jest.config.js --testPathPattern packages/components --watch",
-    "test:cjs": "../../node_modules/jest/bin/jest.js --config ./config/jest.cjs.config.js --testPathPattern packages/components",
-    "test:umd": "../../node_modules/jest/bin/jest.js --config ./config/jest.umd.config.js --testPathPattern packages/components"
+    "test:cjs": "npm run build && ../../node_modules/jest/bin/jest.js --config ../../config/jest.cjs.config.js --testPathPattern packages/components",
+    "test:umd": "npm run build && ../../node_modules/jest/bin/jest.js --config ../../config/jest.umd.config.js --testPathPattern packages/components"
```
benjamn: Should we be using `npx jest` here instead of `../../node_modules/jest/bin/jest.js`?

packages/common/src/types/types.ts
```diff
@@ -0,0 +1,171 @@
...
+  | 'updateQuery'
+  | 'refetch'
+  | 'variables'
+> & {
+  fetchMore: (<K extends keyof TVariables>(
+    fetchMoreOptions: FetchMoreQueryOptions<TVariables, K> &
+      FetchMoreOptions<TData, TVariables>
+  ) => Promise<ApolloQueryResult<TData>>) &
+    (<TData2, TVariables2, K extends keyof TVariables2>(
+      fetchMoreOptions: { query?: DocumentNode } & FetchMoreQueryOptions<
```
benjamn: Wow, Prettier is doing an especially bad job of formatting these types.

packages/hoc/src/__tests__/queries/api.test.tsx
SimenB: No, they're the same (from an API standpoint)

benjamn:
I was able to clear the forever-waiting CircleCI statuses by changing the `master` branch protection rules, as described in https://circleci.com/docs/2.0/workflows-waiting-status/.

packages/common/src/types/types.ts
hwillson: Very much 🙁

packages/components/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  OperationVariables,
+  QueryFunctionOptions,
+  QueryResult,
+  BaseMutationOptions,
+  MutationFunction,
+  MutationResult,
+  SubscriptionResult
+} from '@apollo/react-common';
+import { DocumentNode } from 'graphql';
+import { BaseSubscriptionOptions } from '@apollo/react-common/src/types/types';
```
patroza: This breaks in distribution: https://github.com/apollographql/react-apollo/issues/3117

patroza:
It seems `ssr: false` queries are executed on server anyway, and appear to have a max processing time of around 5s, is that a new feature?

packages/components/src/types.ts
hwillson: Argh - that's vscode's `typescript.suggest.autoImports` at work 🙁. Fix coming shortly - thanks for pointing this out @patroza!

FezVrasta:
Is there any chance to see this fixed before the first release? It's gonna bite a lot of users. Thanks!

https://github.com/apollographql/react-apollo/pull/2892#issuecomment-499891694

hwillson:
@patroza That isn't intended - I'll take look, thanks!
@FezVrasta For sure, we'll be reviewing / addressing all outstanding issues before releasing, thanks as well!

patroza:
When I create a hook with `useQuery`, set `notifyOnNetworkStatusChange: true`, and I use this hook in two different places on the same page, if I call `refetch()`, the `networkStatus` changes for the hook that also provided this `refetch` function, but not for the other one.

Is that by design?

hwillson:
@patroza Can you share a quick code example of how you're creating the hook and re-using the result?

patroza:
@hwillson 
```
const useSomeQuery = () => {
   const result = useQuery(gql`...`, {notifyOnNetworkStatusChange: true})
   return result
}
```

Now call `useSomeQuery` from two different components, call `result.refetch()` inside component1; the `result.networkStatus` will change in component1, but not in 2.
however the `result.data` in both are updated.

package.json
```diff
@@ -1,176 +1,103 @@
...
-    "react"
-  ],
+  "author": "opensource@apollographql.com",
   "license": "MIT",
-  "main": "lib/react-apollo.cjs.js",
-  "module": "lib/react-apollo.esm.js",
-  "react-native": {
-    "react-dom/server": false
-  },
-  "typings": "lib/index.d.ts",
```
zackify: was this supposed to be removed? I am trying this on react native and it doesnt work

zackify:
Just wanted to throw out there, this only works in react native after adding back the deleted react-native json object: https://github.com/apollographql/react-apollo/pull/2892/files#r294482777

package.json
hwillson: Thanks @zackify - no this wasn't supposed to be removed. I'll get it added back in.

hwillson:
@blakedietz regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-490722885, thanks for the offer! I think we're okay on this for now, but we'll keep this in mind.

@lucasecdb regarding https://github.com/apollographql/react-apollo/pull/2892#pullrequestreview-242056659, I've created a new issue to track this: https://github.com/apollographql/react-apollo/issues/3128

@Dattaya regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-499543304, are you still seeing this happen?

@FezVrasta regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-499891694, I've created a new issue to track this: https://github.com/apollographql/react-apollo/issues/3129

@Negan1911 regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-500706541, are you still seeing this happen?

@bummey regarding your question about skipping with the subscription hook, I misread your question
https://github.com/apollographql/react-apollo/pull/2892#issuecomment-502010580 when I answered with https://github.com/apollographql/react-apollo/pull/2892#issuecomment-502058784 (I thought you said subscription HOC). I'll look into subscription hook skipping.

@patroza regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-502437305, I've created a new issue to track this: https://github.com/apollographql/react-apollo/issues/3130
And regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-502784831, I've created a new issue to track it as well: https://github.com/apollographql/react-apollo/issues/3131

Dattaya:
@hwillson, yep, but no one's complaining, I must be doing something wrong, I'll try to create a simple example soon and see if this problem reappears.

Dattaya:
False alarm, it's my own fault, sorry.

mugwump:
Circling back to one of the initial questions: Is there a release somewhere on the horizon? We're just starting out a new project, scheduled to go live somewhere in October this year. Should we base our developments on 2.5.x or rather on the 3.0.0-alpha-branch, with the hope that it will graduate out of alpha before we go live?!

Dattaya:
But I've forgotten to report one inconsistency, initial `data`, while apollo is in the process of fetching it, is `undefined` on the server, but on the client `data` is an empty object. If you want to see it in action,

- `git clone --branch apollo https://github.com/Dattaya/ts-react-ssr-basis apollo-data-undefined && cd apollo-data-undefined`
- `npm i && npm start`
- When you visit http://localhost:3000/test directly, the first logged value is `undefined` 
image
and when you open http://localhost:3000 and press on `Test`, browser's console reports `{}`
image

tafelito:
@hwillson will react-apollo 3.0 be part of a release of apollo client 3.0 as well? If that's the case, will it include any sort of cache persitence in SSR? 
There were some talks in this [issue](https://github.com/apollographql/apollo-cache-persist/issues/32) saying that in there will be a way to handle this in 3.0, not sure if it's on the roadmap

zackify:
I found another bug while testing in React Native. It looks like when I trigger a refetch, loading is never changed to true. Going to look through the code and see if I can figure out why. Here's my code:


```
let { data, refetch, loading } = useProviderOffersQuery();
if (!data || !data.stuff) return null;
  console.log(loading, 'loading ??');
  return (
    <FlatList
      data={data.stuff.items}
      keyExtractor={item => item.id}
      onRefresh={() => refetch()}
      refreshing={loading}
      renderItem={({ item }) => <Item offer={item} />}
    />
  );```

**Comment by patroza** [comment]
> I found another bug while testing in React Native. It looks like when I trigger a refetch, loading is never changed to true. Going to look through the code and see if I can figure out why. Here's my code:

afaikl `loading` never changes to `true` in that case, just the `networkStatus` changes if you enable that behavior.

**Comment by zackify** [comment]
> > I found another bug while testing in React Native. It looks like when I trigger a refetch, loading is never changed to true. Going to look through the code and see if I can figure out why. Here's my code:
> 
> afaikl `loading` never changes to `true` in that case, just the `networkStatus` changes if you enable that behavior.

Thanks for the info, I tested this out, it looks like the networkStatus never changes either, unless I add this: 

```
// Temporary, bug in apollo where loading state does not change with refetch
        setLoading(true);
        await refetch();
        setLoading(false);
```
Because I am forcing the state to change, i can see the network status changes. However if I refetch without setting a loading state myself, neither the loading or networkStatus changes in the component

**Comment by Dattaya** [comment]
@zackify, I think you need to add `notifyOnNetworkStatusChange: true,` to the query's options for `loading` to change when you load more or refetch.

**Comment by zackify** [comment]
@Dattaya thank you that worked, and sorry for thinking this was a problem with the new release. Everything is looking great to me so far, used a bunch of queries and mutations in RN :D

**Comment by IsaacHardy** [comment]
I receive an error Hooks can only be called inside the body of a function component. after upgrading to 3.0.0-beta.1. The odd thing here, is that I am not using any hooks but just a simple <Query> component inside a class.

Is this error a symptom of the deprecation of support for class base components within react-apollo?

Error:
Simulator Screen Shot - iPhone X - 2019-06-19 at 10 09 06

Code from top of stack:
```js

1. import React, { Component } from 'react';
2. import { Query } from 'react-apollo';
3. 
4. import FullscreenPlayer from './FullscreenPlayer';
5. 
6. import { GET_MEDIA_PLAYER_VISIBILITY } from './queries';
7. 
8. /**
9.  * Selectively renders FullscreenPlayer component is MediaPlayer is visible
10.  */
11. class MediaPlayer extends Component {
12.   shouldComponentUpdate() {
13.     return false; // 🚀
14.   }
15. 
16.   renderPlayer = ({ data = {} }) => {
17.     if (!data.mediaPlayer || !data.mediaPlayer.isVisible) return null;
18.     return <FullscreenPlayer />;
19.   };
20. 
21.   render() {
22.     return (
23.       <Query query={GET_MEDIA_PLAYER_VISIBILITY}>{this.renderPlayer}</Query>
24.     );
25.   }
26. }
27. 
28. export default MediaPlayer;

```

ekfn:
@IsaacHardy which version of react are you using? try to update to latest if it's not.

patroza:
When using `useQuery`, `error` does not seem to reset for me, ie if a query fails, and I do a refetch with different variables, the query succeeds, but `error` will still contain the old error.

hwillson:
@mugwump Regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-503075764, I'd recommend basing things off of the RA 3 beta release. Fingers crossed we're soft launching the end of next week.

@Dattaya Regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-503134546, I've created https://github.com/apollographql/react-apollo/issues/3143 to track this.

@tafelito Regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-503155889, no RA 3 and AC 3 won't be tied together. RA 3 is landing very soon, whereas we have several AC minor releases planned before AC 3 will be available.

@IsaacHardy Regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-503579070, I've created https://github.com/apollographql/react-apollo/issues/3144 to track this.

@patroza Regarding https://github.com/apollographql/react-apollo/pull/2892#issuecomment-504008145, there's a good chance this was addressed by https://github.com/apollographql/react-apollo/pull/2892/commits/f7c420afebf148974433779ececbff3cc1d22820. Definitely let me know if this problam still persists after the next beta is available (very soon).

mfpiccolo:
I am getting this from 3.0.0-beta.1:

`this.currentObservable.query.getCurrentResult is not a function`

It sounded like this was addressed. Will there be a beta.2 going out soon?

Also what are the steps to run locally?

hwillson:
@mfpiccolo The next beta is coming very shortly (this afternoon).

On that note, I believe all outstanding issues in this thread have been broken out into separate issues, and are being tracked under the [Release 3.0](https://github.com/apollographql/react-apollo/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Release+3.0%22) milestone. We released `react-apollo` 2.5.7 earlier today, and aren't planning on releasing any further RA 2 changes (other than critical bug fixes). I'm excited to say the time has come to merge this PR! 🎉 Further RA 3 changes will now be based off of `master`, so we're getting closer to launch!

Thanks for all of the feedback in this thread everyone!

tzarger:
@IsaacHardy Had the same issue, and strangely changing your `renderPlayer` to `RenderPlayer` should fix it ... at least for me, it did.

FezVrasta:
I'm trying to upgrade from @apollo/react-hooks beta.8 to beta.10 but our functional tests are now failing with:

```
Uncaught Error: Uncaught TypeError: e.currentObservable.query.resetQueryStoreErrors is not a function
```

Any clue on what's wrong? The app seems to work fine outside Cypress.

noah79:
What is the recommended way to migrate from using compose()?  It's been removed in 3.0 and I have many, many, **many** existing class components that use @compose(@graphql, @graphql,...) patterns.

hwillson:
@noah79 React Apollo's `compose` function was just a copy of lodash's [`flowRight`](https://lodash.com/docs/4.17.11#flowRight) function.

vladshcherbin:
@hwillson what? `compose` is removed? It's needed for HOCs usage, don't force us to install another libraries to get functionality we used for years just because hooks are a new react hype.

bwlt:
@vladshcherbin `compose` function is pretty easy to write. It is so common that you can find It in many modules, like redux, recompose etc.
You can copy It [from here](https://github.com/reduxjs/redux/blob/v4.0.4/src/compose.js) for example

vladshcherbin:
@bwlt so in every project with react-apollo I'll need to copy-paste a function or add another package just to make HOCs work? You can find it in redux, recompose exactly for the same reason - they are needed for users how use HOCs.

Unbelievable, how new updates ruin experience for old users. First, apollo-server unusable, not apollo-client. Guess it's time to search for alternatives. 🤦‍♂

enriquemorenotent:
@vladshcherbin We are talking about a new major release, not a minor. Libraries always end up dropping some methods and implementing new ones. Saying that you need an alternative because upgrading to 3.0.0 is too much hassle, because you need to use another library, seems unfair.

Example: Ruby on Rails decided to drop Form helpers to create HTML because they felt it would make the framework more agnostic, and gave more options to the users. It is not an unreasonable decision, even if it made upgrading a little more complicated.

After all you have an option. To not upgrade to 3.0.0. And I'm not saying that you are not entitled to think that it does not fit your needs and to search for alternatives, but the light you put it under seems unfair. No need for facepalm.

patroza:
wow quite strong emotions here about such a small and basic function.
I think major releases are a great time for such cleanup, as long as the old release still gets bug fixes you can stick to that one, instead of join the 'hype'.

i dont think it's apollo's job to provide you with a general hoc composition util. nor that they have to carry that baggage forever onward because they included it before ;) otoh, if it's such a pain, maybe deprecating it first and removing it later may be smoother sailing :)

vladshcherbin:
Exactly, first HOCs are removed from all examples and can be found only in API section, replaced with render props hype. How hooks come and a small helper function a few bytes is removed (it's actually not removed and used in tests). All of this is a big fu for old users who used apollo since long time ago.

If HOCs are not welcome anymore, remove it altogether and tell users to migrate to render props or hooks or find other libraries with HOC support.

I'm not the only one who noticed this - https://github.com/apollographql/apollo-client/issues/3253. I'm just tired of such things as apollo is bringing lots of pain lately, starting from server where I'm forced to use old version for > 1 year just because [v2 api sucks](https://github.com/apollographql/apollo-server/issues/1308) (most upvoted issue). Now apollo-client and suggestion to stay on v2 forever.

Take redux as an example, hooks were added there too, was `compose` removed? No. Tell me a good reason why it was "removed" here.

vladshcherbin:
You can also see the reaction to HOC removal from docs - https://github.com/apollographql/apollo-client/issues/3253#issuecomment-396326078, it clearly shows how users appreciate this changes.

dantman:
It's worth noting that a `compose` in Apollo besides being redundant, isn't even useful to most people.

Firstly, many users use HOCs with decorators instead of composition. If you're using decorators you don't need compose at all.

Secondly, even if you're not using decorators you still don't need compose. `compose(graphql, graphql)(Component)` is basically just another way of writing `graphql(graphql(Component))` or `Component = graphql(Component); Component = graphql(Component);`, you do not need to import a `compose` function at all to use HOCs.

And of course if you are within the subset of users whose teams avoid use of decorators and use HOCs enough that writing `compose(graphql, graphql)(Component)` instead of `graphql(graphql(Component))` will make your codebase nicer. Then you are probably already using a HOC helper library like recompose for the other HOCs you use (and if not it may still be in your codebase because there are 2000+ packages depending on it). Which is perfectly reasonable because this is not an Apollo specific issue, this is a function regarding how you use any HOCs in your codebase.

I might even say that including `compose` in Apollo was a bad idea from the start. Instead of teaching people how HOCs worked and allowing users to choose which one of the HOC coding styles best suits their team, it probably ended up with a lot of people copying and pasting example code thinking that `compose` is magic and they can't mix Apollo's HOCs with other HOCs.

It's also worth noting that redux's [`compose`](https://redux.js.org/api/compose) is a special case. It may be the same function implementation wise but it's not used in the same way. In redux `compose` is **not** used to compose HOCs like the `compose` in Apollo and recompose, in Redux `compose` is used to compose **store enhancers**.

mikl:
@vladshcherbin this is an open source project, the Apollo maintainers are not your paid slaves, and are not obliged to maintain old code forever, so stop whinging.

1. You're free to continue using the old versions. 
2. You are also free to create your own HOCs for Apollo.

The source is there if you have the talent to use it.

patroza:
@dantman in case of decorators those are only usable on classes right? I think that's definitely against the grain of recent class-less movement :)

dantman:
@patroza Yeah, but that's a recent movement that's grown in tandem with hooks. When HOCs were introduced many things could only be done with class components so most components were classes and there was no issue with using decorators. ^_^ And even if you tried to avoid classes by using recompose to manage state, all you ended up doing was wrapping your "class free" component in several layers of class based wrapper components.

Making a codebase without any class components only really became viable when hooks were introduced.

vladshcherbin:
@dantman I don't use decorators and don't have another libraries which have `compose` inside. While you find `compose` bad, lots of users find it useful in apollo and in other libraries. Not so long ago HOCs were the only way to use apollo and I know lots of users who love them and prefer over render props till today.

@mikl with such toxic attitude as paid slaves, please don't mention me in any of your messages, thanks.

The only reason a new issue about bringing back `compose` is not opened is that v3 is not released and it's still possible to return it and make everyone happy.

mikl:
@vladshcherbin the toxic attitude is yours: complaining that the people who've created the open source software you've profited from (presumably without paying) are not doing it the way you want. That is treating them like slaves.

vladshcherbin:
@mikl treating like a slave by just asking the reason it was removed? Okay, man, sure. I've added you to black list, please don't waste your time by replying to me. Thanks and have a nice day 👋

willisplummer:
I use compose from redux whenever I need it, but I would also be happy to switch to just using the lodash function of redux and all other relevant HOC libraries decided to stop supporting it.. 

generally agree that it’s unintuitive for so many HOC libraries to ship and maintain their own versions of a very basic and identical function.

heymartinadams:
@vladshcherbin The reason why the slave comment was brought up wasn’t because you asked the reason why it was removed. You complained about it having been removed.[1]  You’re correct in that the removal causes us developers some additional pain (having to retool our libraries).

Nevertheless, we developers live *by the grace and generosity of other people’s unpaid contributions* and it’s not right to impose our will on others.

1: `“Unbelievable, how new updates ruin experience for old users. First, apollo-server unusable, now apollo-client. Guess it's time to search for alternatives. 🤦‍♂”`
`”...don't force us to install another libraries...”`
`“ I'm just tired of such things as apollo is bringing lots of pain lately”`, etc.

rajington:
GraphQL, React, and JavaScript in general represent rapidly evolving ecosystems and I appreciate the diversity of opinions. There is a simple open-source solution to this problem however: fork the repo and add back `compose`.

Tooling can even make re-releasing/rebasing automatic, and other users can even contribute other tools that help a HoC-based strategy. The fork could also represent an active PR against this repo, to help the community find/discuss this strategy (even if it's closed). In fact I think the best place to continue this discussion is in that PR/issue.

Specifically for whether hooks are hype worth investing into, thankfully the React team itself addresses [this question directly](https://reactjs.org/docs/hooks-faq.html#do-hooks-replace-render-props-and-higher-order-components), and I can tell you from months of using hooks/`Suspense`/`ErrorBoundary` in production, it helps me write code that is safer and cleaner. The popularity of adopting this pattern in other frameworks also helps convey community sentiment.

hiroppy:
@hwillson I have one question. Why did you remove `GraphqlQueryControls` and `MutationFunc`? I wanna know substitutes. Thanks. https://github.com/apollographql/react-apollo/pull/2892/commits/ade881f07b1175d28b0aae79915bfc5ed8dd9e5a

mptorz:
@vladshcherbin Hey man, this is just not the way to submit your feedback and criticism in the open source community. It is perfectly ok to have strong opinions, but you cannot come here with the attitude like that. You could have submitted the same concerns without making it sound like apollo team owes you anything. Just be respectful and nice dude.

dunika:
@vladshcherbin Take a deep breath and compose yourself.

hansololai:
I just want to point out about that the `compose` was originally from the `recompose` project, because it was very useful, then a low of other projects start to also contain this method. 

I'll give you good reason why it should not be in this repo.

1. This actually created a lot of redundancy code when packaging the client code into vendor.js. You basically have multiple `compose` function in it. The same thing happens a lot with `lodash` and some other very useful but small packages. (some of them are caused by incorrect peer dependency setup). 

2. A good developer should always know what you have imported, if I were to use the `compose` function, I would use the one from `recompose`, not just import it from any package that has it. 

3. Adapting to unix philosophy of "a good tool should do one thing and does it well" is just good

4. per `recompose` README, even the original author of `recompose` is supporting hooks now, the HOC structure could be considered "deprecated" (this is too strong, but you get the idea). And trying to move to the new thing, and remove supporting deprecated stuff is always the first thing a MAJOR release does. see how java stop supporting deprecated APIs in every marjo release?

kinyat:
Sorry, this might be a dumb question.

Since this MR is merged, can I assume that 3.0.0 with hooks support is already published to the public?

The reason I am asking is that I see nowhere, including the official doc, mentioned anything about version 3.0 nor hooks support.

The newest version I see in the official doc is 2.5.

Cheers

oxy88:
@kinyat 
https://www.npmjs.com/package/@apollo/react-hooks
there are some docs in npm but not in github

MidnightDesign:
3.0 milestone: https://github.com/apollographql/react-apollo/milestone/4

hutber:
@vladshcherbin I enjoyed your rant, it reminded me of some of previous employess who I couldn't wait to get away from :D

PR #2953, opened 2019-04-11
Add result prop to withMutation HOC

### LINKED ISSUE #2952: withMutation should support loading/error/called/data states
Previous (closed) issues: #421 #1842

The `graphql()` (and `withMutation()`) HOCs currently have a surprising limitation in that the mutation results aren't passed by them, which isn't explained by the docs, and there also isn't an obvious reason for it. The natural expectation from reading the docs would be that HOCs and components would have feature parity (despite HOCs having fallen by the wayside to render props and more recently hooks) and would just be alternative APIs for the same functionality, especially since `graphql()` delegates to components under the hood. In other words, passing mutation results in `withMutation()` would follow the principle of least surprise.

The answer also shouldn't be "just use components or hooks"; the reason to pick HOCs currently is that components suffer from "wrapper hell" unless a helper library like react-adopt is used, and that hooks are still immature.

slikts:
Closes #2952

### Checklist:

* [X] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
* [X] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

### DIFF
--- a/src/mutation-hoc.tsx
+++ b/src/mutation-hoc.tsx
@@ -56,12 +56,16 @@ export function withMutation<
 
         return (
           <Mutation {...opts} mutation={document} ignoreResults>
-            {(mutate, _result) => {
+            {(mutate, result) => {
               const name = operationOptions.name || 'mutate';
-              let childProps = { [name]: mutate };
+              let childProps = {
+                [name]: mutate,
+                [`${name}Result`]: result,
+              };
               if (operationOptions.props) {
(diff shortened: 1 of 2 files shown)


hwillson:
Great idea @slikts, thanks for working on this. Any chance you could add one or two tests to verify this is working as expected?

slikts:
Closing in favor of #3008

Issue #2801, opened 2019-02-14
Typescript typing for graphql HOC with variables not used in component


VincentLanglet:
**Intended outcome:**

I do like to split the UI component and the data. That is why when I use react-apollo, I prefer HOC : 

ClientForm.component.tsx
```
export interface ClientFormProps {
  client?: Client;
}

export class ClientForm extends PureComponent<ClientFormProps> {
  public render(): ReactNode {
    return ...
  }
}
```

ClientForm.container.tsx
```
export interface Data {
  client?: Client;
}
export interface Variables {
  id: string;
}

export type ContainerProps = Omit<Variables & ClientFormProps, keyof Data>

export default graphql(..., {
  options: (props: ContainerProps): QueryOpts<Variables> => ({
    variables: {
      id: props.id,
    },
  }),
  props: (props: OptionsProps<ContainerProps, Data, Variables>): Data): ({
    client: props.data && props.data.client,
  })
})(ClientForm)
```

The component is re-usable without the data, if we provide a `client` props.
And the container is usable directly with an `id` props.

But this way I have a typescript error : 
Because I use `id` in the options of the container, they expect `id` to be in ComponentProps.
There is for me no reason to write
```
export interface ClientFormProps {
  client?: Client;
  id: string;
}
```
Cause both <Component client={...} /> and <Container id={...} /> are valid.

**Actual outcome:**

Property "id" is missing in Readonly<ClientFormProps> but required in ...

**How to reproduce the issue:**
I have a minimal reproduction this way

ClientForm.component.tsx
```
export type ClientFormProps = DataProps<any>

export class ClientForm extends PureComponent<ClientFormProps> {
  public render(): ReactNode {
    return ...
  }
}
```

ClientForm.container.tsx
```
export interface Variables {
  id: string;
}

export default graphql(..., {
  options: (props: Variables & DataProps<any>): QueryOpts<Variables> => ({
    variables: {
      id: props.id,
    },
  }),
})(ClientForm)
```

**Solution**
I think the problem is in this type : 
```
[...]

Issue #2083, opened 2018-06-08
TypeScript - props function of graphql HOC fail type checking


lirbank:
The example in the docs does not work, and I have not found a way around it either:

https://www.apollographql.com/docs/react/recipes/static-typing.html#props

Reproduction:

https://stackblitz.com/edit/react-ts-rpb3ex

TS Error:
```
Argument of type '{ options: ({ episode }: InputProps) => { variables: { episode: string; }; }; props: ({ data }: O...' is not assignable to parameter of type 'OperationOption<InputProps, Response, Variables, ChildDataProps<InputProps, Response, Variables>>'.
  Types of property 'props' are incompatible.
    Type '({ data }: OptionProps<InputProps, Response, OperationVariables>) => { error?: ApolloError; netwo...' is not assignable to type '(props: OptionProps<InputProps, Response, OperationVariables>, lastProps?: void | ChildDataProps<...'.
      Type '{ error?: ApolloError; networkStatus: number; loading: boolean; variables: OperationVariables; fe...' is not assignable to type 'ChildDataProps<InputProps, Response, Variables>'.
        Object literal may only specify known properties, and 'error' does not exist in type 'ChildDataProps<InputProps, Response, Variables>'.
```

**Version**

* apollo-client@2.3.2
* react-apollo@2.1.4

Skovy:
@lirbank hopefully this PR will resolve this issue: https://github.com/apollographql/react-apollo/pull/2094

A bit of a sledgehammer 🔨 approach you could take in the short term would be to use `any`:

```typescript
graphql<InputProps, Response, Variables, ChildProps>(HERO_QUERY, {
  options: ({ episode }) => ({
    variables: { episode }
  }),
  props: ({ data }) => ({ ...data } as any)
});
```

Based off that example, it looks like a slightly more correct type cast might look something like this:

```typescript
type Props = ChildDataProps<InputProps, Response, Variables>;

graphql<InputProps, Response, Variables, Props>(HERO_QUERY, {
  options: ({ episode }) => ({
    variables: { episode }
  }),
  props: ({ data, ownProps }) => ({ data, ...ownProps} as Props)
});
```

skovy:
@lirbank FYI the above PR was merged recently. Hopefully, it resolves the issue for you! Is this issue good to close?

jameslaneconkling:
I just forked @lirbank 's example and bumped `react-apollo` to `2.1.9` and it's still an issue.

[...]

PR #2998, opened 2019-04-29
Fix types to pass own props through hoc

tricoder42:
This PR extends types of HOCs to allow wrapped component's props to be used in parent component.

Following the [Using Apollo with TypeScript](https://www.apollographql.com/docs/react/recipes/static-typing) and source code, it seems I have to set `InputProps` explicitly and only these props I'm allowed to use in parent component. However, sometimes there're props passed down to wrapped component, which aren't related to graphql HOC at all.

Possibly related issues: https://github.com/apollographql/react-apollo/issues/2801 https://github.com/apollographql/react-apollo/issues/2083

---

I have a [query](https://github.com/tricoder42/apollo-typings-fix/blob/master/query.ts) as HOC where I transform received data in `options.props`:

```ts
interface ResponseType {
  remoteValue: number
}

export interface withDataProps {
  loading: boolean
  remoteValue: number | undefined
}

export const withData = graphql<{}, ResponseType, {}, withDataProps>(
  gql`
    query getRemoteValue {
      remoteValue
    }
  `,
  {
    props: ({ data }) =>
      data == null
        ? { loading: false, remoteValue: undefined }
        : {
            loading: data.loading,
            remoteValue: data.remoteValue
          }
  }
)
```

I use this HOC to provide data to a [component](https://github.com/tricoder42/apollo-typings-fix/blob/master/DataComponent.tsx):

```ts
type DataComponentProps = {
  // own props, not provided by HOC
  value: number
} & withDataProps

export const DataComponent = withData(
  ({ remoteValue, value }: DataComponentProps) => (
    <div>
      Value: {value}, Remote Value: {remoteValue}
    </div>
  )
)
```

**Intended outcome**

I should be able to set `DataComponent` props, which are not provided by HOC:

```ts
<DataComponent value={42} />
```

**Actual outcome**

Typescript complains that `value` props doesn't exist on `DataComponent`:

```
test.tsx:8:8 - error TS2322: Type '{ value: number; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<{}, any, any>> & Re
adonly<{}> & Readonly<{ children?: ReactNode; }>'.
  Property 'value' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<{}, any, any>> & Readonly<{}> & Readonly<{ children?: Reac
tNode; }>'.

8       <DataComponent value={42} />
```

**How to reproduce the issue**

1. Checkout https://github.com/tricoder42/apollo-typings-fix
2. Run `yarn; yarn tsc`

### DIFF
--- a/src/mutation-hoc.tsx
+++ b/src/mutation-hoc.tsx
@@ -30,11 +30,11 @@ export function withMutation<
   let mapPropsToOptions = options as (props: any) => MutationOpts;
   if (typeof mapPropsToOptions !== 'function') mapPropsToOptions = () => options as MutationOpts;
 
-  return (
-    WrappedComponent: React.ComponentType<TProps & TChildProps>,
-  ): React.ComponentClass<TProps> => {
+  return <OwnProps extends {}>(
+    WrappedComponent: React.ComponentType<OwnProps & TProps & TChildProps>,
+  ): React.ComponentClass<OwnProps & TProps> => {
     const graphQLDisplayName = `${alias}(${getDisplayName(WrappedComponent)})`;
-    class GraphQL extends GraphQLBase<TProps, TChildProps> {
+    class GraphQL extends GraphQLBase<OwnProps & TProps, TChildProps> {
       static displayName = graphQLDisplayName;
       static WrappedComponent = WrappedComponent;
       render() {
@@ -69,7 +69,7 @@ export function withMutation<
 
               return (
                 <WrappedComponent
-                  {...props as TProps}
+                  {...props as OwnProps & TProps}
                   {...childProps as any}
                 />
               );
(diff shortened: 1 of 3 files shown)


tricoder42:
Looking at errors in CI, it seems that these types fails when component doesn't reiceve any props. I tried several approaches, but the best workaround seems to be use `compose` and set props to final component explicitly:

```ts
type DataComponentProps = {
  // own props, not provided by HOC
  value: number
}

export const DataComponent: React.ComponentType<DataComponentProps> = compose(withData)(
  ({ remoteValue, value }: DataComponentProps & withDataProps) => (
    <div>
      Value: {value}, Remote Value: {remoteValue}
    </div>
  )
)
```

Composing multiple HOCs is very common actually, so this is not a big deal. I have to idea how it could be further improved.

Closing PR as suggested changes are incomplete.

PR #3008, opened 2019-05-03
Adds mutation result to mutation hoc. Closes #1967

### LINKED ISSUE #1967: Mutation HoC WrappedComponent additional props
Hello guys, thanks for the great react integration. It definitely got better with each release.
There is one thing I don't understand. What is the reason to omit all the useful properties like `loading` in mutation HoC (https://github.com/apollographql/react-apollo/blob/e2147f8448e691ff5a7244c8aab68c28e94df4be/src/mutation-hoc.tsx#L61)? Why don't pass them under some key like `mutation` or even `[name]` like it is done for queries?

andycarrell:
This change adds the result of the a mutation to the HoC. It follows the same pattern as the Query HoC. Closes #1967 

### Checklist:


andycarrell:
May be a duplicate of #2953 🤦‍♂

PR #3029, opened 2019-05-10
fix: require react-apollo based on module env

### LINKED ISSUE #2900: Could not find "client" in the context of ApolloConsumer when using react-apollo 2.5.3 with MockedProvider
**Intended outcome:**

I have a React component that makes use of `<ApolloConsumer>` in order to access its `client` render prop and explicitly call `client.query` when a user action occurs. In order to test my component in my Jest/Enzyme test suite, I wrap it in `<MockedProvider>` and mount it with Enzyme. My tests should run with no issue.

Example component:

```
import React, { Component } from "react";
import { graphql, ApolloConsumer } from "react-apollo";
import gql from "graphql-tag";

export const query = gql`
  query ErrorTemplate {
    people {
      id
      name
    }
  }
`;

class App extends Component {
  constructor(props) {
    super(props);

    this.state = {
      people: []
    };
  }

  render() {
    return (
      <ApolloConsumer>
        {(client) => (
          <main>
            <button onClick={(event) => {
              client.query({
                query
              }).then(response => {
                this.setState({ people: response.data.people });
              });
            }}>Fetch People</button>
            <ul>
              {this.state.people.map(person => <li key={person.id}>{person.name}</li>)}
            </ul>
          </main>
        )}
      </ApolloConsumer>
    );
  }
}

export default App;
```

Example tests:

```
import React from "react";
import { MockedProvider } from "react-apollo/test-utils";
import Enzyme, { mount } from "enzyme";
import Adapter from "enzyme-adapter-react-16";

Enzyme.configure({ adapter: new Adapter() });

import App, { query } from "./App";

const mock = {
  request: {
    query
  },
  result: {
    people: [
      { id: 1, name: 'John Smith' },
      { id: 2, name: 'Sara Smith' },
      { id: 3, name: 'Budd Deey' },
    ]
  }
};

describe("App", () => {
  it("mounts", () => {
    const app = mount(
      <MockedProvider addTypename={false} mocks={[mock]}>
        <App />
      </MockedProvider>
    );
  });
});
```

**Actual outcome:**

My test fails with a message saying: `Invariant Violation: Could not find "client" in the context of ApolloConsumer. Wrap the root component in an <ApolloProvider>.`

Screenshot:
image

Reproduction: https://github.com/bkuzma/react-apollo-error-template/tree/context-error-example

If you look at the commit history of the `context-error-example` branch, the single test passes in the second to last commit (412d2d57f52d2b10c1620dd2259bd8269eab74e5), but when I update `react-apollo` to 2.5.3 in the last commit (26474d6aab4cbdca7d9c08c9af386f4be8f0a2de), the test fails with the same error described above.

**Version**

- apollo-client@2.3.2
- react-apollo@2.5.3

jure:
Attempts to fix #2900 again. #2907 fixed MockedProvider for cases where `cjs` is used, but not in situations that use ESM, e.g. webpack. This PR requires `./` instead, which causes Node.js to load the `main` (`react-apollo.cjs.js`), and `webpack` to load the `module` (`react-apollo.esm.js`) field in `package.json`.

This works for both my use cases: tests with Jest & styleguide with webpack, but I might be missing something obvious.


jure:
Any thoughts on when this could land?

jure:
So... I could keep updating this with the latest master, but perhaps there's different plans, and this is going to be handled in a different way? I imagine it's not rare for people to have mocked providers in styleguides/storybooks these days?

Anyway, please let me know if this is not relevant or according to your roadmap, so I can stop worrying/checking this MR and implement a local alternative solution.

PR #3031, opened 2019-05-11
Apollo Client 2.6 integration work

hwillson:
Apollo Client 2.6 integration work.


hwillson:
Published and available as `react-apollo@beta`.

Issue #2890, opened 2019-03-22
Variable defaults to null but has been omitted before


eliias:
This [commit](https://github.com/apollographql/react-apollo/commit/3f72fef0263d12769c41a6d8ce103594fc7d364f#diff-752fa63ef9932eb40562c367c6ad6d1c) has removed a condition that previously made the function return if a variable was optional and not set. From 2.5.0 on, every variable is set to null, if optional and not provided any value. I wonder if the new behaviour is intentional and just right or if this is a regression (could not find anything related in [changelog](https://github.com/apollographql/react-apollo/blob/master/Changelog.md)).

**Intended outcome:**

```
variables: {}
```

<img width="165" alt="Bildschirmfoto 2019-03-22 um 01 13 19" src="https://user-images.githubusercontent.com/1222696/54801974-b8e24d00-4c3f-11e9-9979-7a523f86df6b.png">

**Actual outcome:**

```
variables: {
  transform: null
}
```

<img width="161" alt="Bildschirmfoto 2019-03-22 um 01 09 55" src="https://user-images.githubusercontent.com/1222696/54801980-c0095b00-4c3f-11e9-99e9-b4593b55787e.png">

**How to reproduce the issue:**

Create a mutation with at least one optional variable. Call the mutation without a value for the optional variable. 

Compared versions:
- 2.4.1 (variable is not set)
- 2.5.0, 2.5.1, 2.5.2 (all 2.5.x seems to have the regression)

The GraphQL query has an optional variable `transform`.

```
mutation saveX ($transform: Transform) {
  myX(transform: $transform) {
    …
  }
}
```

**Version**

- apollo-client@2.5.1
- react-apollo@2.5.1

hwillson:
@eliias Can you confirm if this is still happening with `react-apollo` 2.5.3?

eliias:
@hwillson Can report in the evening. 👋 to a fellow Ottawa citizen :D

eliias:
Did a quick check on `2.5.3` and tested locally. It still adds `{transform: null}` to the request payload in case the variable is not set.

ShepelievD:
Can also confirm in `2.5.3`

AdamTyler:
This issue is preventing me from creating queries with default values. Since omitted variables come in as `null` it won't replace them with the defaults

hwillson:
This was fixed in https://github.com/apollographql/react-apollo/pull/3056 and will be published in the next RA release. Thanks!

[...]

PR #3056, opened 2019-05-22
fixes variable default to null, where it should be undefiend

### LINKED ISSUE #1569: Optional input is set to null
Hello. I have a "users" request with userFilterInput which is optional. In my case i do not set the filter. Apollo will send this request to the server 

```
{"operationName":"users","variables":{"userFilterInput":null},"query":"query users($userFilterInput: UserFilterInput) {\n  users(userFilterInput: $userFilterInput) {\n    id\n    email\n    lastName\n    firstName\n    __typename\n  }\n}\n"}
```

Why userFilterInput is set to null and not to undefined? Bellow you can find my code and you can see that I did not set up any filter.

```
export const USERS = gql`
  query users($userFilterInput: UserFilterInput) {
    users(userFilterInput: $userFilterInput) {
      id
      email
      lastName
      firstName
    }
  }
`;

export default graphql(USERS, {
  options: () => ({
  }),
  props: ({ data: { loading, error, users, networkStatus, refetch } }) => ({
    usersLoading: loading,
    usersError: error,
    usersRefetch: refetch,
    users,
    networkStatus,
  }),
});
```

ZhengYuTay:
Fixes #1569.

The problem is when we did not implicitly set a nullable variable value to a mutation query. It will be set to null which is not intended.


axax:
@ZhengYuTay Thanks for the pull request. Fix is also for issue #2890. Hope we can merge this soon.

johansteffner:
@hwillson do you have any eta on when this fix will be released?

yashrajpchavda:
@hwillson same question as @johansteffner asked, any idea when this be released?

PR #3079, opened 2019-05-29
useQueryLoader() hook

mbrowne:
For an overview of this PR and the motivation for it, see https://github.com/apollographql/apollo-feature-requests/issues/106.

I'm very open to feedback on this...all the tests and examples work, but there are multiple potential variations on the API.

I added a new component to the example app just for the sake of demonstrating usage of the hook, but of course `useQueryLoader()` doesn't necessarily need to be in the example app, or could be used in a different way. Happy to adjust or remove the example if needed.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
   "editor.formatOnSave": true,
-  "prettier.singleQuote": true
+  "prettier.singleQuote": true,
+  "prettier.semi": true
 }
--- a/examples/hooks/client/src/RocketDetail.tsx
+++ b/examples/hooks/client/src/RocketDetail.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import { useQueryLoader } from '@apollo/react-hooks';
+import gql from 'graphql-tag';
+
+import { RocketInventory } from './types';
+
+const GET_ROCKET_DETAILS = gql`
+  query getRocket($id: ID!) {
+    rocket(id: $id) {
+      id
+      model
+      year
+      stock
+    }
+  }
+`;
+
+export interface RocketDetailProps {
+  rocketId: number;
+}
+
+export function RocketDetail({ rocketId }: RocketDetailProps) {
+  return useQueryLoader(GET_ROCKET_DETAILS, {
+    variables: { id: rocketId }
+  })(({ data: { rocket } }: { data: { rocket: RocketInventory } }) => {
+    if (!rocket) {
+      return <div>Rocket ID {rocketId} not found</div>;
+    }
+    return (
+      <article>
+        <h2>Rocket Details</h2>
+        <dl>
+          <dt>Model</dt>
+          <dd>{rocket.model}</dd>
+          <dt>Year</dt>
+          <dd>{rocket.year}</dd>
+          <dt>Stock</dt>
+          <dd>{rocket.stock}</dd>
+        </dl>
+      </article>
+    );
+  });
+}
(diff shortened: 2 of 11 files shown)


patroza:
I don't think this can be considered a Hook. It uses a Hook, but it doesn't have the same semantics as a Hook;
e.g `useState`, `useMutation` and `useQuery` all return state/functions that can be immediately used. `useEffect` executes lifecycle hooks. However `useQueryLoader` returns a React component, and is therefore more like a Component or HOC.

mbrowne:
Actually it returns React *elements*, not a React component, which is an important distinction (rendering a new component type each time could be bad for performance). One advantage of this over a render-props component is that you avoid an unnecessary `QueryLoader` element in the tree (avoiding unnecessary elements in the tree is one of the selling points the React team has listed for hooks). See also this quote from Dan Abramov:

> you can return elements directly. But don't return components.

https://www.reddit.com/r/reactjs/comments/9yq1l8/how_do_you_feel_about_a_hook_returning_components/ea3q7dt?utm_medium=android_app&utm_source=share

I have not seen any official recommendation stating otherwise...this hook is more about the logic of rendering the loading and error components than it is about rendering something specific. I was originally on the fence about it as well, but after reflecting on it I realized that this is definitely in the spirit of hooks. And it isn't breaking any rules or best practices as far as I'm aware.

patroza:
@mbrowne fair enough, yes you are right - a component factory, versus a HOC or wrapper component.
I've just applied a similar approach locally and I start liking it too!

mbrowne:
If, for whatever reason, the Apollo team and community decide against including this hook in react-apollo, I just thought of an alternative solution that would make it easier to implement this sort of thing in app code or a 3rd party library—there could be a `customOptions` prop or something similar that you could pass into `ApolloProvider` and then you'd be able to access those custom options via the `useApolloClient` hook. In this case, `customOptions` would be used to set the default loading and error components. Perhaps there are other use cases that would benefit from this as well.

patroza:
@mbrowne what would be the benefit compared to it’s own context provider/consumer?

mbrowne:
It just seems logical to me...everywhere you're using the same `ApolloProvider`, you're probably going to want to use the same default loader and error components (except in special cases where you're handling it manually anyway). But admittedly the benefits are pretty minor...one less provider to create, one less if statement validating that wherever the hook is used, it's a descendent of a valid provider (since useApolloClient() already takes care of that)...

hwillson:
Thanks for working on this @mbrowne! We're still mulling this (and your alternate suggestion in https://github.com/apollographql/react-apollo/pull/3079#issuecomment-502467648) over. I'll post back shortly - thanks again!

patroza:
One thing I was just bit with @mbrowne is that when you want to use other hooks, you have to use them outside the body of the loader, not inside. If you do it inside, and you change the query in a way that requires it to refetch and render the loader again, React will complain that not the same amount of Hooks / order of the hooks was executed.

mbrowne:
@patroza That's true, but it makes me wonder if you are using React's [ESLint plugin for hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks). I'm not at my computer to verify right now, but I'm pretty sure that will give you a linting error right away if you try to use a hook within the callback, which IMO mitigates this issue. And I think using the ESLint plugin can be considered a best practice and something most teams will be using.

patroza:
@mbrowne fair point, im using Typescript with TSLint, but I am going to switch it over now to ESLint now that it seems the way forward with the Typescript parser support :)

hwillson:
Thanks very much for working on this @mbrowne. Unfortunately, this isn't something we can accept at this time. While we can definitely see the advantages this PR helps provide for certain scenarios, we're trying to be super sensitive to bundle sizes. For this reason, we're trying to keep React Apollo as basic as possible (while still providing value), letting people build and expand on top of it as needed. That being said, we are open to considering smaller API changes that help make custom hooks easier to integrate (we can continue that discussion in https://github.com/apollographql/apollo-feature-requests/issues/106 or elsewhere). Thanks again!

PR #3098, opened 2019-06-06
Align to React version 16.8.3, fix build issues

matthargett:
* Align to React version 16.8.3, used by current stable release of React Native (0.59). 
* Fix build issues on Windows by using 'npx' instead of direct execution. 
* Eliminate dependency warning by using new namespace react testing library.

### Checklist:


packages/all/package.json
```diff
@@ -24,15 +24,15 @@
     "clean": "rm -Rf ./node_modules ./lib/* ./meta/bundlesize/* ./meta/coverage/*",
     "prepare": "npm run build",
     "prebuild": "npm run clean",
-    "build": "../../node_modules/.bin/tsc -p ./config",
-    "postbuild": "../../node_modules/.bin/rollup -c ./config/rollup.config.js",
+    "build": "npx tsc -p ./config",
```
hwillson: We actually recently moved away from using `npx` with scripts in https://github.com/apollographql/react-apollo/pull/2892/commits/cda6b37f81ef19926df2eede7d4e7c623a869c7e. @benjamn any objections to moving back?

packages/all/package.json
```diff
@@ -24,15 +24,15 @@
...
+    "postbuild": "npx rollup -c ./config/rollup.config.js",
     "predeploy": "npm run build",
     "deploy": "npm publish --tag beta"
   },
   "peerDependencies": {
     "apollo-client": "^2.6.0",
     "graphql": "^14.2.1",
-    "react": "^16.8.6"
+    "react": "^16.8.3"
```
hwillson: I'm not a huge fan of basing our peerDeps on the fact that another library (React Native in this case) hasn't been updated to use the latest version. I guess in this case though, the impact is fairly harmless. If we're going to open up the range a bit though, maybe we should backup all the way to `^16.8.0`. We aren't dependent on anything in the 16.8.x patch bumps, but are dependent on 16.8.0. This way we're going back about 5 months, and would likely cover more cases than just React Native. 

```suggestion
    "react": "^16.8.0"
```

packages/all/package.json
matthargett: I was just trying to fix the build on Windows, the error was ```.. is not a command```. I'm open to any solution that lets me run the build/tests in my environment :)

packages/all/package.json
matthargett: SGTM! should I do it in the other packages as well?

matthargett:
@hwillson ping? Anything else I need to do for this to be merged?

hwillson:
Thanks again for this @matthargett and don't worry about having to keep this up to date with recent changes. We've been adjusting the way we handle tests, which I know is annoying to have to rebase against. I'll get this in synch with the other changes shortly, and get it merged in.

packages/all/package.json
hwillson: @matthargett I'm going to change this around a bit, as `npx` can't use a parent's `node_modules` (it can only use a `node_modules` in the current package or global location). This means each child package `npx` call ends up re-installing the referenced packages first, since all of our devDeps are in the root of the monorepo. As a quick workaround to this, we'll add the utilities we're referring to with `npx` as devDeps in each child package. Not ideal, but we're only talking about 4 deps, and Renovate will help us keep them up to date.

Issue #3090, opened 2019-06-03
Query stuck in loading state when rendered after a query or a network error


ducarroz:
**Description**
When a `<Query>` component which has previously resulted in a qraphQL or network error is re-rendered, its state changes from error to loading and no request is executed. The component is now stuck in loading state.

**Intended outcome:**
If the query variables do not change, the `<Query>` stay in error state. Else the state change to loading and a new request is executed (cache or network).

**Actual outcome:**
The `<Query>` component's state changes to loading and no request is executed, whatever the query's variables change or not.

**How to reproduce the issue:**
I made a sample project to demonstrate the issue at https://codesandbox.io/s/apolloquerystuckaftererror-heyxu

1) open https://codesandbox.io/s/apolloquerystuckaftererror-heyxu
2) click on 'Next' and wait for response (300ms)
3) click on 'GQL Error' or 'Link Error' and wait for response (300ms)
4) click on 'Next' or 'Same'
==> `<Query>` stuck in loading state

**Version**

- apollo-client@2.6.0
- react-apollo@2.5.6

hwillson:
We're just waiting on related Apollo Client changes, then we can get the fix for this (https://github.com/apollographql/react-apollo/pull/3107) merged in. This will be included in the next RA release (actually, we're holding up the next release until this issue is resolved).

tsjohns9:
It looks like this issue has been merged, but I still consistently get this error.

I am using the `useLazyQuery` hook from `react-apollo`

Here are my dependencies:
```
	"dependencies": {
		"@apollo/react-ssr": "^3.0.1",
		"apollo-boost": "^0.4.4",
		"next": "9.0.4",
		"next-with-apollo": "^4.2.0",
		"react": "16.9.0",
		"react-apollo": "^3.0.1",
		"react-dom": "16.9.0"
	},
```

I am working on a login form. So far I am having two issues.

1. When logging in with an incorrect username/password the graphql server returns an error. This error gets rendered to the UI. When attempting to log in again with the same incorrect credentials no request is made, and the error clears out. The returned data is always `undefined`. This happens even though I have `fetchPolicy` set to `network-only`.

2. When logging in with an incorrect username/password for the first time, and then trying again with a different user/pass combination the query gets stuck in a loading state, and a network request is performed and resolved but I receive no update.

ducarroz:
@tsjohns9 sound like similar to https://github.com/apollographql/react-apollo/issues/3425.

Issue #2070, opened 2018-06-04
Cannot recover from errors via `refetch` or `resetStore`


tmeasday:
**Intended outcome:**

If a query throws an error, recover, either by `refetch`-ing (if intermittent, this could work) or `resetStore`-ing and changing something (in my case logging in / thus changing an auth header).

**Actual outcome:**

`react-apollo` will not process the successful result of the re-run query (gory details below).

**How to reproduce the issue:**

Here is a reproduction: https://codesandbox.io/s/qv1q541v2j

Note this points at: https://launchpad.graphql.com/vm3qvkz743 a really simple launch pad that throws an error 50% of the time.

To reproduce the problem in its pure form:

1. Load https://qv1q541v2j.codesandbox.io/ , ensuring it gets the error the first time (ie refresh the page until it does)
2. Hit the refresh button. Notice that even if the graphql network request returns actual data (it will 50% of the time), it never updates the component.

I can make a less random reproduction, this just seemed an easy way to show what's going on.

**Version**

* apollo-client@2.3.2
* react-apollo@2.1.4 [NOTE this doesn't happen with older versions, see gory details below)

**Gory details**

Why does this happen?

The reason is the confluence of two things:

1. [This PR](https://github.com/apollographql/react-apollo/pull/1849) which added an optimization to ignore query results if the previous query result was complete

2. If an query is in the store as an error, a refetch will *not* inform listeners when it begins, due to this line:

https://github.com/apollographql/apollo-client/blob/09b26bd8d14abcc9f3f7c4648d56b66df83ef658/packages/apollo-client/src/core/QueryManager.ts#L469-L472

So what happens is:
a) first query returns, error. Picked up by RA, renders.
b) refetch happens (either directly or via `resetStore`). Due to (a) being an error and 2. above, RA is never informed.
c) refetch query returns. In 1. above, RA compares it to the first query, which has `networkStatus === 7`, so it throws it away.

I'm happy to submit a PR with a fix, if someone can give me some guidance on where to fix it. @jbaxleyiii do you think the optimization (1.) is perhaps broken in other places and should be removed? I won't pretend to fully understand all the code paths that lead to `next()` being called, but I suspect other similar ones could have been missed.

tmeasday:
Note: You can work around the issue above by using `fetchPolicy: 'cache-and-network'` on your query.

bryandowning:
[...]

PR #3107, opened 2019-06-10
Fix always loading issue caused by error handling

hwillson:
The React Apollo `Query` component relies on Apollo Client's `ObservableQuery->getCurrentResult` method to get the results of a query to display. When `ObservableQuery` encounters an error, it stores that error in its associated query store, so that it can be passed back to React Apollo for handling. Unfortunately, the error stored in the query store isn't cleared out until a component unmounts, which means components that handle an error response followed by a valid response, won't get the chance to render the valid response. They're stuck rendering the error state, since the error stored in the `ObservableQuery` query store is never removed.

This commit calls into a new `ObservableQuery.resetQueryStoreErrors` method, to clear out previously used errors, before moving on to handle subsequent responses. This means a component can handle an error, then receive and display a valid subsequent response.

Fixes: https://github.com/apollographql/react-apollo/issues/3090

**Note:** This PR is dependent on https://github.com/apollographql/apollo-client/pull/4941 being merged.


src/Query.tsx
```diff
@@ -349,14 +349,8 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
...
       error: error => {
-        if (!this.lastResult) {
-          // We only want to remove the old subscription, and start a new
-          // subscription, when an error was received and we don't have a
-          // previous result stored. This means either no previous result was
-          // received due to problems fetching data, or the previous result
-          // has been forcefully cleared out.
-          this.resubscribeToQuery();
-        }
+        this.resubscribeToQuery();
```
hwillson: I haven't added any additional tests to validate this behavior, as switching to always resubscribe in the error callback here lead to existing test failures. These were then fixed by clearing query store errors below.

jasonpaulos:
@hwillson Any plans to bring this fix to RA 3 as well? Always resubscribing after an error fixes #2070 as well.

hwillson:
@jasonpaulos Definitely! I have it ready (I just haven't committed it yet). I should have it in the RA 3 branch shortly. As soon as it's there, I'm then planning on publishing a final version of RA 2, then will merge the RA 3 branch into `master`. So🤞by tomorrow morning we should be working against only one codebase again (yay! 🙂). Thanks!

sambaek:
any updates on when the final version of RA 2 will be published with this fix?

jasonpaulos:
@sambaek I believe it was just published as 2.5.7

developdeez:
I was using 2.5.6 and it was getting stuck.
I upgraded to 2.5.8 now it blinks several times before showing the page. 
What gives?

jasonpaulos:
Hi @developdeez, could you create a new issue with this bug? It would also help a lot if you could provide a runnable reproduction that shows this. Thanks.

Issue #2899, opened 2019-03-25
Child component stuck in loading state, but the data is already loaded.


OurMajesty:
**Intended outcome:**

There are two React components:
parent: Requests (stateful component)
child: graphql(List), having prop `state`

There's buttons in Requests, which are setting state.
In turn, List component's query variable is set by state prop, which is equals to state in Requests component.
So the data for List should be refetched every time I click button in Requests.

**Actual outcome:**

Data is refetched successfully, but sometimes (80% of times) List is stuck in loading state. If I click some other button again, loading is resolving, the data is new.

I did try install react-apollo 2.4.1 and it fixed the error.
Also, if I recreate component every time (`key` prop) it helps too.
BTW, react-apollo@2.5.2 was having the same issue

**How to reproduce the issue:**

~~I did create code example using local state, but it works fine, so I suggest the problem reproduction is requiring real graphql endpoint.~~
Real GraphQL endpoint example is failing as expected:  
[https://codesandbox.io/s/039pyz78pn](https://codesandbox.io/s/039pyz78pn)

**Version**

- apollo-client@2.5.1
- react-apollo@2.5.3

bogdansoare:
also having this issue

hwillson:
@OurMajesty Any chance you could try building a reproduction using https://github.com/apollographql/react-apollo-error-template? It will simulate having a backend, so hopefully that will help.

OurMajesty:
@hwillson Yes, I did try it right away, but it worked fine. Furthermore, I have found additional places in my app, where it stucks too. For example, dynamic search, that's why I mentioned this issue in #2808.
I wish Launchpad still worked, it would help to reproduce.
Had a lot of work lately, but still planning to find a way. Maybe use some open GraphQL API.

OurMajesty:
If someone needs a workaround, right now I am using react-apollo 2.4.1 along with client 2.5.1 and other latest libraries and doen't really found any compatibility issues, local state and client queries work fine too.

OurMajesty:
@hwillson I did finally reproduced it. The same Codepen link: https://codesandbox.io/s/039pyz78pn

First, click on "computer". Result: empty list.
Second, click on "noop". Result: empty list again.
And it stuck.

Cheesier:
Having this exact same issue, seems to me that the issue is when variables change but the response remains the same.

OurMajesty:
[...]

PR #3126, opened 2019-06-17
Fix component stuck in loading state for network-only fetch policy

jasonpaulos:
Due to inconsistencies between `Query.lastResult` and the actual previous result passed to the child component for rendering, sometimes the `Query` component would suppress a render that contained necessary data to bring a component out of a loading state. Concretely, this happens when using a `network-only` fetch policy and the variables of a query changes but the result remains the same.

See #2899 for a complete description of the issue.


hwillson:
Thanks very much for working on this @jasonpaulos - this has been a long standing React Apollo issue, so it's great to see traction here!

Looking this over, I think now might be a good time to take this solution further. The reason why your `lastRenderIsDifferent` approach works, is unfortunately because of a deep rooted issue that has been part of React Apollo for a long time. It's not necessarily a bug; it's more of a problem with the current architectural approach. I'll try to explain this issue below.

In the current `Query` component, after the initial creation and first render of the component (where things are in put into a loading state), we check for future data updates via Apollo Client by setting up a subscription on an Apollo Client `ObservableQuery` instance. This is the code that happens in the `Query` component's `startQuerySubscription` method. This idea here is that as soon as a data update is received through the `ObservableQuery` subscription, we can then force the `Query` component to re-render.

When we force a component to re-render in `startQuerySubscription`, after new subscription data is received, we end up calling the `getQueryResult` method. This is the method that takes the new data from Apollo Client, cleans it up a bit, then returns this as the data that is used in the render prop function. So `getQueryResult` ultimately gives us the data that is used in each render. 

This is where the crux of the issue really comes in. `startQuerySubscription` leverages `ObservableQuery` to watch for new data, then when new data is found, triggers a re-render by force updating the component. The call for that re-render / forced update happens within the same tick of the event loop as the call to `getQueryResult` to get and return the resulting data. `getQueryResult` gets the current data by calling into Apollo Client's `ObservableQuery.currentResult()` function, which in theory should have the most recent data available. But, `startQuerySubscription` has set up its own subscription on the `ObservableQuery` instance, so when it receives new data, and eventually calls down into `getQueryResult` (in the same tick), `ObservableQuery.currentResult()` won't return that new data yet - it will still return the previous data as it hasn't had a chance to refresh yet, using its own internals.

The real problem here is the fact that we're asking for data from an `ObservableQuery` instance in 2 places: 1) in `startQuerySubscription` via the subscription we setup, and 2) in `getQueryResult` via a call to `ObservableQuery.currentResult()`. We're then using the data received in `startQuerySubscription` to decide when to call `getQueryResult`, expecting the same new data to be used, but `getQueryResult` is then using its own version of the data which is out of date. 

This is definitely a bit convoluted to explain, but to make matters worse here we're leveraging `lastResult` to essentially work around the fact that the two places we load Apollo Client data from aren't always the same (and to block unnecessary renders). Your changes to use `lastRenderIsDifferent`, along with `lastResult` handling, do work around this issue, but the core problem still remains.

If we take a step back here a bit and think this through, we should be able to find a way to always make sure that if our `ObservableQuery` subscription says new data is available, and we're going to force a render based on that new data being available, that the render itself will always use the exact same data (taking into consideration a stored last render result of some kind, to decide of we should trigger a new render - ie, only if data has changed). I don't think Apollo Client is the problem here; the `ObservableQuery` API seems to be doing what it should. I think instead the way we've set ourselves up to watch for data changes in Apollo Client, then trigger a render based on that data, needs to be streamlined as our timing is off.

I'll setup a meeting to discuss this further (cc @benjamn). You're definitely headed in the right direction on this, and it would be amazing to finally nail this problem down once and for all. Thanks again @jasonpaulos!

hwillson:
I've added changes in https://github.com/apollographql/react-apollo/pull/3126/commits/5839fb9b1de451c2d4dfe86c928904dfe50302ae to help address the issues raised in https://github.com/apollographql/react-apollo/pull/3126#issuecomment-503189606. The current result to be used/displayed is now aligned across the entire component, and render tracking/control is now managed via a single instance variable called `lastRenderedResult` (which is always the last truly rendered result).

test/client/Query.test.tsx
```diff
@@ -524,14 +524,18 @@ describe('Query component', () => {
...
+      let count = 0;
       const Component = () => (
         <Query query={allPeopleQuery} fetchPolicy={'cache-only'}>
           {result => {
-            catchAsyncError(done, () => {
-              expect(result.loading).toBeFalsy();
-              expect(result.networkStatus).toBe(NetworkStatus.ready);
-              done();
-            });
+            if (count === 0) {
```
hwillson: I had to change this test a bit to get it to pass (due to timing issues in the test only), but it's still testing the same functionality and is accurate.

test/client/Query.test.tsx
```diff
@@ -545,14 +549,18 @@ describe('Query component', () => {
     });
 
     it('default fetch-policy', done => {
+      let count = 0;
```
hwillson: Same comment as above.

test/client/graphql/queries/lifecycle.test.tsx
```diff
@@ -581,20 +581,8 @@ describe('[queries] lifecycle', () => {
       { loading: false, a: 7, b: 8, c: 9 },
 
       // Load 6
-
-      // The first render is caused by the component having its state updated
-      // when switching the client. The 2nd and 3rd renders are caused by the
-      // component re-subscribing to get data from Apollo Client.
-      { loading: false, a: 1, b: 2, c: 3 },
```
hwillson: We no longer have to deal with these extra renders! 🎉

src/Query.tsx
```diff
@@ -455,11 +422,12 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
       ) {
         // Make sure data pulled in by a `no-cache` query is preserved
         // when the components parent tree is re-rendered.
-        data.data = this.previousData;
+        data.data = previousData;
       } else {
         const { partialRefetch } = this.props;
         if (
           partialRefetch &&
+          typeof currentResult.data === 'object' &&
```
benjamn: Might want to add a check that `currentResult.data !== null` here.

src/Query.tsx
```diff
@@ -443,8 +407,11 @@ export default class Query<TData = any, TVariables = OperationVariables> extends
       const { fetchPolicy } = this.queryObservable!.options;
       Object.assign(data, { loading, networkStatus, error });
 
+      const previousData =
+        this.lastRenderedResult ? this.lastRenderedResult.data : {};
+
       if (loading) {
-        Object.assign(data.data, this.previousData, currentResult.data);
+        Object.assign(data.data, previousData, currentResult.data);
```
benjamn: It's confusing that `data` here has a `data.data` property. Can we call the `data` variable something like `result` instead? This method is called `getQueryResult`, after all.

test/client/Query.test.tsx
```diff
@@ -545,14 +549,18 @@ describe('Query component', () => {
...
       const Component = () => (
         <Query query={allPeopleQuery}>
           {result => {
-            catchAsyncError(done, () => {
-              expect(result.loading).toBeFalsy();
-              expect(result.networkStatus).toBe(NetworkStatus.ready);
-              done();
-            });
+            if (count === 0) {
+              catchAsyncError(done, () => {
```
benjamn: Might want to go ahead and remove `catchAsyncError` usage, since we did that for 3.0?

hwillson: I think I'll leave this one as is, since these changes have to be merged into `release-3.0.0` by hand (and I'll make sure `catchAsyncError` isn't used). Thanks!

PR #3168, opened 2019-06-24
Move get{Data,Markup}FromTree into new @apollo/react-ssr package.

benjamn:
The `RenderPromises` class needs to remain in `packages/hooks/src/ssr` because it relies on the private `QueryData` type, but this reorganization means we no longer call `require("react-dom/server")` anywhere except in the new `@apollo/react-ssr` package.

Should address https://github.com/apollographql/apollo-feature-requests/issues/118.


packages/ssr/README.md
```diff
@@ -0,0 +1,49 @@
...
+### Installation
+
+```
+npm install @apollo/react-ssr
+```
+
+### Overview
+
+<a name="getDataFromTree"></a>
+#### a) [`getDataFromTree`](https://github.com/apollographql/react-apollo/blob/release-3.0.0/packages/ssr/src/getDataFromTree.ts)
```
benjamn: This `README.md` needs to be fleshed out before this PR is merged (and before this package is published).

hwillson: I'm going to do a full review/edit on all RA 3 package `README`'s before publishing, so feel free to merge this PR as is for now.

htulipe:
Just to be sure @hwillson @benjamn, this will only land on RA3 right ? It won't be release on 2.x

hwillson:
@htulipe Correct.

PR #3171, opened 2019-06-25
Make sure `useSubscription` cleans up when done

### LINKED ISSUE #3160: useSubscription hasn't cleaned up after unmount
**Intended outcome:**
subscription cleaned up after unmount

**Actual outcome:**
subscription hasn't cleaned up after unmount

sorry for the lack of details, but it really is. I didn't found in repo any code that clean ups subscription after unmount

**Version:**
3.0.0-beta.2

hwillson:
The `useSubscription` Hook isn't properly restting/removing the internal `ObservableQuery` instance it uses, when it's no longer needed. This means that the subscription used by the hook is never properly shut down, leading to unnecessary network communication. This PR ensures that the `useSubscription` internals are properly cleansed when the component wrapping the hook is unmounted.

Fixes #3160.

### DIFF
--- a/packages/hooks/src/useSubscription.ts
+++ b/packages/hooks/src/useSubscription.ts
@@ -36,6 +36,7 @@ export function useSubscription<TData = any, TVariables = OperationVariables>(
   subscriptionData.context = context;
 
   useEffect(() => subscriptionData.afterExecute());
+  useEffect(() => subscriptionData.cleanup.bind(subscriptionData), []);
 
   return subscriptionData.execute(result);
 }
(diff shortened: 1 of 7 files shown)


packages/hooks/src/useSubscription.ts
```diff
@@ -36,6 +36,7 @@ export function useSubscription<TData = any, TVariables = OperationVariables>(
   subscriptionData.context = context;
 
   useEffect(() => subscriptionData.afterExecute());
+  useEffect(() => subscriptionData.cleanup.bind(subscriptionData), []);
```
benan789, line 39: if i am not mistaken, this needs to be invoked? `subscriptionData.cleanup.bind(subscriptionData)()`

Issue #1759, opened 2018-03-06
Wrong return type of withApollo with TypeScript?


nilshartmann:
I wrap a React Component with the `withApollo` HOC using TypeScript. My React Component has no "own properties", only receives the ApolloClient via the HOC. 

```
interface MyAppProps {
  client: ApolloClient<any>;
}

function MyApp(props: MyAppProps) {
  props.client.resetStore(); // OK, WORKS!

  return <div>Lorem Ipsum</div>;
}
```

When I use my component, TypeScript complains due to missing `client` property (complete error see below):

```
const MyAppWithApollo = withApollo(MyApp);

<ApolloProvider client={apolloClient}>
    <MyAppWithApollo />  <!--- ERROR: Property client is missing
  </ApolloProvider>,
```

**Intended outcome:**
One should be able to use `withApollo` without type args if the wrapped component has no own props.

**Actual outcome:**
The TypeScript compiler complains when I use the wrapped component in my code:
```
[ts]
Type '{}' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<MyAppProps, ComponentState>> & Readonly<...'.
  Type '{}' is not assignable to type 'Readonly<MyAppProps>'.
    Property 'client' is missing in type '{}'.
```

When I add type arguments to withApollo it works:
```
const MyAppWithApollo=withApollo<{}, {}>(MyApp)
```
it works.

TypeScript shows the following types for the return value of `withApollo:`

*  without type args: `React.ComponentClass<MyAppProps>`
*  with "dummy" type args ({}, {}): `React.ComponentClass<{}>

The second is correct and would also be correct IMHO also for the first case.

**How to reproduce the issue:**

* Create a React Component with out own properties
* wrap the component with `withApollo` 
* embed the wrapped component in your code

You can find an example here: https://gist.github.com/nilshartmann/8b7ffb6429708dfae19bf6b4b1b5dfd5#file-withapollobug-tsx

**Version**

* apollo-client@2.2.5 (installed via apollo-boost 0.1.1)
* react-apollo@2.1.0-beta.2 (also tried beta.3 with same result)
* typescript@2.7.2
You can find my dependencies here: https://gist.github.com/nilshartmann/8b7ffb6429708dfae19bf6b4b1b5dfd5#file-package-json

clentfort:
[...]

PR #3178, opened 2019-06-25
Remove 'client' prop from withApollo type definition

RigoTheDev:
Fix for https://github.com/apollographql/react-apollo/issues/1759

Stops the `client` prop from being required with the `withApollo(...)` HOC


wszgxa:
Hi, @hwillson 
Could we release a version 2.5.9 with this change?

I don't have time to upgrade the react-apollo from 2 to 3 for now.

PR #3186, opened 2019-06-26
remove apollo-utilities in favor of just the isEqual needs from the shared root

jbaxleyiii:
After talking with @hwillson, this seems like a pretty easy swap to thin out the package by quite a lot! Let's see what bundlesize says after its done though...


jbaxleyiii:
The discrepancy between bundlesize and bundlephobia are pretty concerning here 
image

Issue #3130, opened 2019-06-17
`ssr: false` queries are being executed on the server


hwillson:
Moving this here from the [RA 3 PR](https://github.com/apollographql/react-apollo/pull/2892) thread: https://github.com/apollographql/react-apollo/pull/2892#issuecomment-502437305

cc @patroza

----

In latest (and previous) beta, it seems `ssr: false` queries are executed on server anyway, and appear to have a max processing time of around 5s, is that a new feature?

patroza:
The "max processing time of around 5s" part is incorrect, it was something else.

hwillson:
Thanks for reporting this @patroza. Unfortunately I haven't been able to re-create this (and https://github.com/apollographql/react-apollo/pull/3197 shows that SSR disabling is working). I'll close this for now, but if there's any chance you can fire over a small runnable reproduction that shows this happening, I'll happily dig in further. Thanks again!

patroza:
hi @hwillson we are definitely running into this issue.
The only way I can make it stop, is by setting `skip` to true on the server.

On the server I will still see `loading` turn to true, and on another pass `loading` turn to false, but at least it seems it's not executing or waiting for any queries.

why would `loading` even turn to true, if `ssr: false` is set, and still even so when `skip: true` is set?

patroza:
I have had success by treating `ssr: false && this.context && this.context.renderPromises` similar to the `skip` case.

Basically in the current release version, inside QueryData.execute:
instead of `if (!skip)`
`if (!skip && ((!this.context && !this.context.renderPromises) || _a.ssr !== false))`

I see that the current code is already different, but similar change should be applied.

imo: `skip` should throughout the queryData be equal to ` this.getOptions().skip || (this.context && this.context.renderPromises && this.getOptions().ssr === false)`

so in short, right now I think `ssr: false` is perhaps not waiting on results, but still executes the queries. And if there are other queries processing which take longer, then the result of the query still ends up inside the memory cache of the client, and gets put into the HTML document so that it is picked up by the frontend on initialization.

hwillson:
Thanks very much for the extra details @patroza - we'll get this fixed.

hwillson:
@patroza Are you still seeing this happen with the latest `beta`'s?

patroza:
@hwillson it looks like it's resolved, at least the client is doing the request, as the data is not coming in on the server side cache state.

hwillson:
@patroza Great to hear, thanks!

[...]

PR #3197, opened 2019-06-28
Test to verify SSR is skipped when setting `ssr` to `false`

hwillson:
Test to validate https://github.com/apollographql/react-apollo/issues/3130.


patroza:
As mentioned in #3130 this only tests and confirms that the query data is not considered in the render, but it does not test nor confirm that the actual query is not executed ;-)

PR #3199, opened 2019-06-28
Adjust `useMutation` result tuple to be more inline with React Core

### LINKED ISSUE #3189: useMutation Tuple signature inconsistent with built-in hooks
To help reduce potential confusion, would it make more sense for the `MutationTuple` to be ordered `[ MutationResult, MutationFunction ]` and remain consistent with all of the built-in hooks? I've been following this pattern in custom hooks, as it always ensures that the value at index 0 is the value to bind against.

hwillson:
This commit re-arranges the `useMutation` result tuple such that the mutation result is in the first position, with the mutation function in the second position. The intent of this change is to bring the `useMutation` Hook more in-line with how the React Core handles its Hook results (e.g. `useState`, `useReducer`).

**This is a breaking change, but since we're in beta hopefully that's okay.** To help lessen the blow a bit this commit includes changes to display a `console.warn` message describing the API changes. This message will be removed before the final version of React Apollo 3 is published.

Fixes #3189.

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -14,10 +14,12 @@ React Apollo [Hooks](https://reactjs.org/docs/hooks-intro.html).
 
 1. [Installation](#installation)
 2. [Hooks Overview](#hooks-overview)
-  - [`useQuery`](#useQuery)
-  - [`useMutation`](#useMutation)
-  - [`useSubscription`](#useSubscription)
-  - [`useApolloClient`](#useApolloClient)
+
+- [`useQuery`](#useQuery)
+- [`useMutation`](#useMutation)
+- [`useSubscription`](#useSubscription)
+- [`useApolloClient`](#useApolloClient)
+
 3. [Reference]()
 
 ### Installation
@@ -29,6 +31,7 @@ npm install @apollo/react-hooks
 ### Hooks Overview
 
 <a name="useQuery"></a>
+
 #### a) [`useQuery`](https://github.com/apollographql/react-apollo/blob/release-3.0.0/packages/hooks/src/useQuery.ts)
 
 **Function:**
@@ -37,7 +40,7 @@ npm install @apollo/react-hooks
 export function useQuery<TData = any, TVariables = OperationVariables>(
   query: DocumentNode,
   options?: QueryHookOptions<TData, TVariables>
-): QueryResult<TData, TVariables>
+): QueryResult<TData, TVariables>;
 ```
 
 **Options:**
@@ -122,6 +125,7 @@ export function RocketInventoryList() {
 ```
 
 <a name="useMutation"></a>
+
 #### b) [`useMutation`](https://github.com/apollographql/react-apollo/blob/release-3.0.0/packages/hooks/src/useMutation.ts)
 
 **Function:**
@@ -130,7 +134,7 @@ export function RocketInventoryList() {
 export function useMutation<TData = any, TVariables = OperationVariables>(
   mutation: DocumentNode,
   options?: MutationHookOptions<TData, TVariables>
-): MutationTuple<TData, TVariables>
+): MutationTuple<TData, TVariables>;
 ```
 
 **Options:**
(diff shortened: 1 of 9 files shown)


packages/hooks/src/useMutation.ts
```diff
@@ -5,10 +5,26 @@ import { DocumentNode } from 'graphql';
...
+
 export function useMutation<TData = any, TVariables = OperationVariables>(
   mutation: DocumentNode,
   options?: MutationHookOptions<TData, TVariables>
 ): MutationTuple<TData, TVariables> {
+  // This is a temporary message intended to help lessen the blow of a
+  // breaking `useMutation` API change during the RA 3 beta process. It will
+  // be removed before the official launch of RA 3.
+  if (showApiChangeWarning) {
+    console.warn(
```
benjamn: Let’s use `invariant.warn` from `ts-invariant` to avoid increasing the bundle size.

packages/hooks/README.md
```diff
@@ -130,7 +134,7 @@ export function RocketInventoryList() {
 export function useMutation<TData = any, TVariables = OperationVariables>(
   mutation: DocumentNode,
   options?: MutationHookOptions<TData, TVariables>
-): MutationTuple<TData, TVariables>
+): MutationTuple<TData, TVariables>;
```
benjamn: We should give the definition of the `MutationTuple` type in this documentation, since otherwise the named type is not super helpful.

packages/hooks/src/useMutation.ts
hwillson: Arg - I keep forgetting about `invariant.warn` - thanks!

packages/hooks/README.md
hwillson: Will do! The contents of this `README` are intended to be temporary (to help with beta testing). This `README` will be thoroughly cleaned up shortly. Thanks!

FredyC:
Just want to point out that docs ([README](https://github.com/apollographql/react-apollo/blob/master/packages/hooks/README.md#b-usemutation)) are kinda mixed, it shows correct signature coming from this change, but the example below has the "old way". That got me confused and had to look into a source to verify it.

hwillson:
Great point @FredyC - thanks! We're going to be overhauling all of the `README` very shortly.

FredyC:
Either way, I am a bit surprised by this change, to be honest. The point is, that user might want to execute a mutation and get the result from returned Promise more often than reading it out from scoped result. That means we will have to be skipping tuple variable, eg. `const [, mutate] = useMutation(...)`.

It's not a big deal breaker, but on the point of aligning with React Core, there is never use case like that (eg. calling setState without reading value) so the reasoning might be a little bit off on my opinion.

pie6k:
I'm also not fan of this. Having mutation result as promise data should be enough in all cases - if you need, you can just assign it to the state yourself or even create custom hook. It also means component will re-render after mutation which is not needed in many cases.

hwillson:
@pie6k Could you elaborate on your comment a bit further? All this is doing is switching the position of the mutation function and the mutation result object in the returned `useMutation` tuple. I don't quite follow how this will impact rendering?

FredyC:
@hwillson He probably means if there wouldn't be any internal state for mutation, it wouldn't need to rerender and you would get a result from the Promise... But then there wouldn't much of a difference from calling `client.mutate` imo :)

No, I think that the return result has its uses, especially the `loading` state which is great to express waiting on the operation to finish. It was actually the reason why I made this PR in first place https://github.com/trojanowski/react-apollo-hooks/pull/93

Thanks for reconsidering switching arguments over, it indeed makes more sense in most cases.

reverie:
This is an important one to get right.

To rephrase what FredyC said, I think the analogy to "how the React Core handles its Hook result" is misguided. Keeping data types in the same order might be the principle to follow if all else were equal. But the principle of 'optional values should come last' is a stronger convention (like in function arguments), and it's more strongly grounded in practice as well (it avoids `[,` noise all over the place).

FredyC:
@reverie Hold your horses, there are already considering switching tuple values around ... https://github.com/trojanowski/react-apollo-hooks/issues/187#issuecomment-509910983

reverie:
@FredyC When people are considering it is the right time to comment, right?

FredyC:
I wouldn't be making such a fuss around it really. If anyone is bothered by `[,` it's trivial to make a custom hook that switches it around. I will be content, either way. I already have some abstractions around it anyway (to handle errors and some other things).

reverie:
Seems like you are making a fuss about it :)

hwillson:
Hi all - we're going to switch it back; we just haven't had a chance yet (we just finished a company get together week, so time has been a bit tight). Everything will be back to normal this week, and this (along with other outstanding changes) will be wrapped up. Thanks!

Issue #119, opened 2016-07-21
Can we add SSR integration to docs?


ads1018:
I noticed support for SSR was merged into master with #83. I'm really looking forward to integrating server side rendering in my apollo stack application but I'm not quite sure how to implement it!

jbaxleyiii:
@ads1018 you bet! I've shipped it silently to make sure it is all working great before documenting it. With the upcoming 0.4.0 API changes, we will be redoing a lot of the docs!

ads1018:
Fantastic! You guys are the best. I <3 apollo.

PR #3214, opened 2019-07-04
Add `useQuery` lazy mode to defer query execution

hwillson:
Initial work to support deferring `useQuery` execution until some point in the future (after `useQuery` has been initialized).

~If a `lazy` option is set to a boolean, `useQuery` will return a tuple instead of a result object. The first position of the tuple is the standard query result object, and the second position is a function that can be triggered to kickstart the normal `useQuery` query cycle.~

~If `lazy` is `true`, the initial query result with have a `loading` status of `false`, and a `data` value of `undefined`. This will only change after query execution has been triggered using the delayed start function (`execute` in the following example):~

```js
const [{ loading, data }, execute] = useQuery(GET_ROCKET_INVENTORY, {
  lazy: true
});
```

~If `lazy` is `false`, a tuple is still returned, but `useQuery` will fire immediately, just like calling `useQuery` without a `lazy` option. So these are the same:~

```js
const { loading, data } = useQuery(GET_ROCKET_INVENTORY);
```

```js
const [{ loading, data }] = useQuery(GET_ROCKET_INVENTORY, {
  lazy: false
});
```

**UPDATE**

The above is no longer accurate. `useQuery` will always run the defined query, but `useLazyQuery` can be used to defer query execution. It works like:

```js
const [{ loading, data }, execute] = useLazyQuery(GET_ROCKET_INVENTORY);
```

Fixes https://github.com/apollographql/apollo-feature-requests/issues/119.

**TODO:**
- [x] Handle `skip`
- [x] Verify SSR
- [x] Consider a `called` option (like `useMutation`)
- [x] Handle `execute` variables
- [x] Tests, tests, tests


FredyC:
Thanks for accepting my suggestion and working on it.

At first glance, it feels somewhat strange having a different return shape based on the existence of a boolean option value. Good thing is that we can name the `execute` however we want which is certainly useful. I am honestly surprised it's possible to have correct type declaration for that, never seen that before 😮 👏 

On the point of `called` value, without it, we would have to track it in a separate state which means an extra re-render (before/after calling `execute`). Adding it means an extra variable has to be added to the standard query result in case of lazy mode.

FredyC:
Another aspect regarding the `called`. Since the lazy mode has `loading = false` initially, it would mean we would have explicitly check if `data` or `error` is set. With `called` we can make the simple assertion for rendering as `called && !loading` which definitely feels better semantically.

FredyC:
I wonder, why not to have the same order of variables in that tuple as the `useMutation` has? Is it intentional to keep people aware of a slightly different mechanics or just an oversight? I think it should be the same to lower the cognitive barrier. It's easy for TypeScript people to see the mistake right away, but regular JS folks might be bumping into it fairly often.

hwillson:
Thanks for the `called` feedback @FredyC - great points. As for the `useMutation` order, we actually switched it around in https://github.com/apollographql/react-apollo/pull/3199 (those changes haven't been published yet). So `useMutation` and these changes now follow more closely with the approach the React core is using.

FredyC:
Yea, just commented on useMutation change which I don't think is actually correct. However, for the `useQuery` the exact opposite applies and I would be contradicting myself here so it's like I haven't said anything 🙊 Ultimately, it is fine either way because people can just as easily make a custom hook to switch those values over to each liking.

hwillson:
Haha - we've struggled with this as well; there are pros/cons to both approaches for sure. For now we're going to go with the current approach of results first, function second (and we'll adjust if/as needed).

packages/hooks/src/data/QueryData.ts
```diff
@@ -330,7 +332,8 @@ export class QueryData<TData, TVariables, TLazy> extends OperationData {
         ...result,
         data: undefined,
         error: undefined,
-        loading: false
+        loading: false,
+        called: true
```
FredyC: I wonder if it shouldn't be `false` as well when skipping. Semantically, the query wasn't called, but it might have been called before and the `skip` was set later. That would mean it should derive the value from the previous result, but since it's `true` by default as well, it doesn't matter.

So I guess it's easier to document that non-lazy queries have a `called` always true.

pie6k:
I'd say better api would be like

```ts
const [shouldFetch, setShouldFetch] = useState(false)
useQuery(GET_ROCKET_INVENTORY, {
  enabled: shouldFetch,
});
```

and then it'll not get executed until `shouldFetch` is set to `true`

FredyC:
@pie6k That's already there with `skip` :) Please read through my original proposition at #119

rovansteen:
Wouldn't it be better to have a seperate hook for this? It feels a bit odd that the return types changes based on the parameters passed to the hook. A hook like `useLazyQuery` is more explicit about what to expect.

FredyC:
@rovansteen Considering its a totally optional feature and you must explicitly specify that option, there is no benefit of separating to an extra hook. Besides, you can always make your own `useLazyQuery` that wraps this. #powerofhooks !

hwillson:
@rovansteen I agree it is a bit strange, but I was interested in saving a few extra bytes. I'll defer to whichever approach @benjamn prefers, during code review.

packages/hooks/src/data/QueryData.ts
```diff
@@ -118,6 +115,85 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+
+  private initialLazyResult(): QueryTuple<TData, TVariables> {
+    return [
+      {
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        called: false,
+        data: undefined
+      } as QueryResult<TData, TVariables>,
+      this.executeLazy.bind(this)
```
benjamn:
```suggestion
      this.executeLazy,
```

packages/hooks/src/data/QueryData.ts
```diff
@@ -118,6 +115,85 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+    ];
+  }
+
+  private executeLazy(options?: QueryLazyOptions<TVariables>) {
```
benjamn:
```suggestion
  private executeLazy = (options?: QueryLazyOptions<TVariables>) => {
```
Let’s use class property syntax here to bind `this.executeLazy` once, in the constructor.

packages/hooks/src/useQuery.ts
```diff
@@ -6,23 +6,30 @@ import {
...
+  TVariables = OperationVariables,
+  TLazy = undefined
+>(
   query: DocumentNode,
-  options?: QueryHookOptions<TData, TVariables>
-): QueryResult<TData, TVariables> {
+  options?: QueryHookOptions<TData, TVariables, TLazy>
+): TLazy extends undefined
+  ? QueryResult<TData, TVariables>
+  : QueryTuple<TData, TVariables> {
```
benjamn: It appears the `QueryTuple` mode would be enabled even if someone passed `useQuery(query, { lazy: false })`, which seems like an argument for `useLazyQuery`?

packages/hooks/src/data/QueryData.ts
```diff
@@ -118,6 +115,85 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+    this.forceUpdate();
+  }
+
+  private getExecuteResult(): TLazy extends undefined
+    ? QueryResult<TData, TVariables>
+    : QueryTuple<TData, TVariables> {
+    const result = this.getQueryResult();
+    this.startQuerySubscription();
+    return (this.getOptions().lazy !== undefined
+      ? [result, this.executeLazy.bind(this)]
```
benjamn:
```suggestion
      ? [result, this.executeLazy]
```

packages/hooks/src/data/QueryData.ts
```diff
@@ -118,6 +115,85 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+      : result) as any;
+  }
+
+  private getExecuteSsrResult(): TLazy extends undefined
```
benjamn: If we don’t keep the `TLazy` approach, I think `getExecuteResult` and `getExecuteSsrResult` should stick to returning just the result, so we don’t have to juggle different return types.

packages/hooks/src/data/QueryData.ts
```diff
@@ -118,6 +115,85 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+      loading: true,
+      networkStatus: NetworkStatus.loading,
+      called: true,
+      data: {}
+    };
+
+    if (this.context && this.context.renderPromises) {
+      result = this.context.renderPromises.addQueryPromise(
+        this,
+        this.getExecuteResult.bind(this)
```
benjamn: Let’s avoid binding in favor of using the class property + arrow function trick, since rebinding functions is expensive, and arrow functions are faster in some engines than bound functions.

hwillson: I didn't realize rebinding was more expensive than using arrow functions, so TIL. That being said, I still visually prefer the explicitness of seeing `.bind` 😢. But, performance trumps my preference, so done!

hwillson:
As of https://github.com/apollographql/react-apollo/pull/3214/commits/157339f4d33ee27f8ebcf8354a60d3ab34eb80ad, `useQuery` will now always run the specified query, and `useLazyQuery` can be used for deferred query execution. E.g.:

```js
const [{ loading, data }, execute] = useLazyQuery(GET_ROCKET_INVENTORY);
```

FredyC:
I agree it's probably more coherent to have a separate hook for that. 

Once again about the order of arguments in the tuple. This is a similar case to `useMutation`. If you are going to switch it there, it would make sense to switch it here as well. Point is that without calling `execute` there will be no result and as such it's definitely more important.

Wouldn't be also desirable to get the Promise with a result from `execute`? In most cases query result is for rendering but in some cases, we might want to check it imperatively to for example send error report or redirect a user to a different page (and showing a toast error message). These are all doable with the inline result, but require `useEffect` and careful handling of deps.

Does a `skip` make any sense for `useLazyQuery`? It's possibly confusing, especially given it's always flipped to the `false`. What about excluding it from types there? It can surely be in underlying structures, but it shouldn't be documented as something that exists imo.

hwillson:
All great points @FredyC, and now that we have a separate hook for this, these are easier to implement. I'll adjust - thanks!

hwillson:
@FredyC Returning a promise with the lazy execution loaded result will require a bit more refactoring than we want to tackle before the RA 3 launch. It's a good idea, but we'll have to schedule it for a future point/minor release.

packages/hooks/src/utils/useBaseQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+import { DocumentNode } from 'graphql';
+
+import { QueryHookOptions, QueryOptions } from '../types';
+import { QueryData } from '../data/QueryData';
+import { useDeepMemo } from './useDeepMemo';
+
+export function useBaseQuery<TData = any, TVariables = OperationVariables>(
+  query: DocumentNode,
+  options?: QueryHookOptions<TData, TVariables>,
+  lazy?: boolean
```
benjamn:
```suggestion
  lazy = false,
```

packages/hooks/src/utils/useBaseQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+      queryDataRef.current = new QueryData<TData, TVariables>({
+        options: updatedOptions as QueryOptions<TData, TVariables>,
+        context,
+        forceUpdate
+      });
+    }
+    return queryDataRef.current;
+  }
+
+  const queryData = getQueryDataRef();
```
benjamn: The `getQueryDataRef` function doesn’t need to be a function if you’re only calling it here.

packages/all/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "sideEffects": false,
   "scripts": {
-    "clean": "rm -Rf ./lib/* ./meta/bundlesize/* ./meta/coverage/*",
+    "clean": "rm -Rf ./lib/* ./meta/bundlesize/* ./meta/coverage/* ./.rpt2_cache",
```
benjamn: This is fine for now, but we might want to use `npx rimraf` instead (if any Windows folks complain).

FredyC: I am Windows folk and I can imagine I could be complaining :)

packages/hooks/src/utils/useBaseQuery.ts
hwillson: Right! Left over from some previous refactoringmadness.

packages/all/package.json
hwillson: Definitely, great point. I've noted this down along with a few other outstanding script related changes, and will get them in place separately. Thanks!

PR #3362, opened 2019-08-13
Add test to verify error is persisted on re-render, when appropriate

### LINKED ISSUE #3295: useQuery's `error` object disappears on rerender
**Intended outcome:**

Error object for failed query should persist across rerenders

**Actual outcome:**

The error object disappears after rerendering

- have the `@apollo/react-hooks` branch installed
- have a proxy set up to force a request to fail. Charles is a good option: https://www.charlesproxy.com/
- The ParentComponent has a 5 second timer. Force the request for your query to fail before that timer expires. It will show the `there is an error` message. But once the timer expires and the Parent rerenders the child component, the error object will disappear and it will display "there is no error" without refiring the query
```
const ReproComponent = () => {
  const { error } = useQuery(/* insert your gql query here */)
  if (error) {
    return <div> there is an error </div>
  }
  return <div> there is no error </div>
}

class ParentComponent {
  constructor() {
    super()
    this.state = { count: 0 };
    this.increment = this.increment.bind(this)
    setTimeout(this.increment, 5000)
  }

  increment() {
    this.setState(prevState => ({ count: prevState.count + 1 })
  }

  render() {
    return <ReproComponent count={ this.state.count } />
  }
}
```
**Version**

System:
OS: macOS 10.14.5
Binaries:
Node: 8.12.0 - /usr/local/bin/node
npm: 6.5.0 - ~/.npm-packages/bin/npm
Browsers:
Chrome: 75.0.3770.142
Safari: 12.1.1
npmPackages:
apollo-cache-inmemory: ^1.1.7 => 1.6.2
apollo-cache-persist: ^0.1.1 => 0.1.1
apollo-client: ^2.2.3 => 2.6.3
apollo-link: ^1.1.0 => 1.2.12
apollo-link-context: ^1.0.8 => 1.0.18
apollo-link-error: ^1.1.0 => 1.1.11
apollo-link-http: ^1.5.2 => 1.5.15
apollo-link-rest: ^0.2.4 => 0.2.4
apollo-link-retry: ^2.2.4 => 2.2.14
apollo-link-state: ^0.4.0 => 0.4.2
apollo-upload-client: ^8.1.0 => 8.1.0
react-apollo: ^2.1.4 => 2.5.8

=> also @apollo/react-hooks, but that doesn't log from the npx command

hwillson:
Issue #3295 demonstrates a problem where errors aren't maintained on re-renders, when they should still be available. This issue was fixed by PR #3339, but this commit will add an extra test to verify that the issue is indeed fixed.

Fixes #3295.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,8 @@
   [@hwillson](https://github.com/hwillson) in [#3337](https://github.com/apollographql/react-apollo/pull/3337)
 - Add `@types/react` as a peer dep, to address potential TS compilation errors when using `ApolloProvider`. <br/>
   [@zkochan](https://github.com/zkochan) in [#3278](https://github.com/apollographql/react-apollo/pull/3278)
+- Make sure `error`'s are maintained after re-renders, when they should be. <br/>
+  [@hwillson](https://github.com/hwillson) in [#3362](https://github.com/apollographql/react-apollo/pull/3362)
 
 ## 3.0.0 (2019-08-06)
 
--- a/packages/hooks/src/__tests__/useQuery.test.tsx
+++ b/packages/hooks/src/__tests__/useQuery.test.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useReducer } from 'react';
 import { DocumentNode, GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
 import { MockedProvider, MockLink } from '@apollo/react-testing';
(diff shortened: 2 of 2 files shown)


dylanwulf:
I am still experiencing this issue when I have either `onCompleted` or `onError` specified in the query options. I tried adding `onError` to the query options in this test, and now it fails: https://github.com/dylanwulf/react-apollo/commit/f01d71478ac777dcb5e78df89b7d8f11f45ad22f

Issue #3365, opened 2019-08-14
TypeError: Cannot read property 'getCurrentResult' of undefined


Ayc0:
I have this error: `TypeError: Cannot read property 'getCurrentResult' of undefined`
when doing:

```js
import React from 'react';
import { useQuery } from '@apollo/react-hooks';

import QUERY from '../gql/queries/query.gql';

export default props => {
  const { data } = useQuery(QUERY);
  return null;
};
```

with:
```json
{
  "dependencies": {
    "@apollo/react-hooks": "3.0.0",
    "apollo-cache-inmemory": "1.6.3",
    "apollo-client": "2.6.4",
    "apollo-link": "1.2.12",
    "apollo-link-http": "1.5.15",
    "apollo-link-ws": "1.0.18",
    "apollo-utilities": "1.3.2",
    "blessed": "0.1.81",
    "graphql": "14.4.2",
    "kleur": "3.0.3",
    "node-fetch": "2.6.0",
    "react": "16.9.0",
    "react-apollo": "3.0.0",
    "react-blessed": "0.6.1",
    "subscriptions-transport-ws": "0.9.16"
  }
}
```

(I also tried `"apollo-client": "2.6.4"`)

The issue seems to come from `useQuery` because I don't have the error when the return is above 

And I don't have multiple `apollo-client` in my `yarn.lock`

hwillson:
@Ayc0 any chance you can put together a small runnable reproduction that shows this happening? There isn't enough to go on with the code sample you've provided. The `Cannot read property 'getCurrentResult' of undefined` error is likely happening due to an attempt to call `getCurrentResult` on the internal `ObservableQuery` after it has been destroyed, which happens during unmount. A repro would help confirm this. Thanks!

Ayc0:
@hwillson, I'm gonna push my buggy code tonight (EDT)

But it was really like in the code snippet that I wrote in the example:

I have a component that imports a `gql` file with parcel 1, and I have the error when the component that has `useQuery(GQL);` is rendered.

(I think the bugs doesn't appear in the 1st render but when it gets the data and tries to re-render)

Ayc0:
Okay, if you run `yarn start` in the folder `terminal-client`, you should see the error:

And the file that causes the issue is [this one](https://github.com/Ayc0/mini-datadog/blob/update-hook/terminal-client/src/Monitorings/index.js) (need to be on the branch `update-hook`)

Ayc0:
[...]

PR #3372, opened 2019-08-15
[QueryData] check if there is a query

Ayc0:
Related to this: https://github.com/apollographql/react-apollo/issues/3365

I got the error `TypeError: Cannot read property 'getCurrentResult' of undefined` so I changed the source in my node_modules and adding this check solved my issue

### Checklist:

### DIFF
--- a/packages/hooks/src/data/QueryData.ts
+++ b/packages/hooks/src/data/QueryData.ts
@@ -397,11 +397,14 @@ export class QueryData<TData, TVariables> extends OperationData {
   }
 
   private handleErrorOrCompleted() {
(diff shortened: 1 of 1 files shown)


garrettmaring:
I'm running into a similar issue here with an uncaught exception when using `!` for the current query.

```ts
  private obsRefetch = (variables?: TVariables) => this.currentObservable.query!.refetch(variables);
```

It's a rare case that I hit when using HMR under certain circumstances. However, it's probably wise to respect the types here and guard against a nil query. I'd be willing to jump in here and contribute but I wanted to check in first and ask if there was good reason for the forced unwrapping.

PR #3382, opened 2019-08-16
feat(packages/hooks): export MutateFunction type

orrybaram:
Howdy! When using the useMutation function and typescript we are typing the mutate function like this:
```ts
const [mutateSomething] = useMutation(QUERY);
const handleOnClick = (mutateSomething: MutationTuple<TData, TVariables>[0]) => (idx: number) => {
 return mutateSomething()
}

```

This PR will allow us to clean up the types a bit and write the same thing as 

```ts
const [mutateSomething] = useMutation(QUERY);
const handleOnClick = (mutateSomething: MutationFunction) => (idx: number) => {
 return mutateSomething()
}

```

If there's already a better way to do this let me know!

### DIFF
--- a/packages/hooks/src/types.ts
+++ b/packages/hooks/src/types.ts
@@ -85,10 +85,12 @@ export interface MutationOptions<TData = any, TVariables = OperationVariables>
   mutation: DocumentNode;
 }
 
+export type MutationFunction<TData, TVariables> = (
+  options?: MutationFunctionOptions<TData, TVariables>,
+) => Promise<ExecutionResult<TData>>;
+
 export type MutationTuple<TData, TVariables> = [
-  (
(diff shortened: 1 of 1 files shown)


jbhannah:
I had been doing

```js
import { MutationFn } from "react-apollo"
```

and the equivalent for 3.0 appears to be

```js
import { MutationFunction } from "@apollo/react-common"
```

orrybaram:
@jbhannah Ah yeah thats exactly what we were looking for! Thanks!

jbhannah:
No problem! 😄 For future reference, [here’s](https://github.com/apollographql/react-apollo/blob/master/packages/all/src/index.ts) where I found all of the exports from the new `@apollo/` packages that used to all come from the `react-apollo` package.

Issue #3390, opened 2019-08-19
querying local data server side returns loading true for first render causing errors 


kedarguy:
**Intended outcome:**

upgrade an existing Next.js application that has local state to use react-apollo 3.0 

I created a repo using nextjs with-apollo example (which was updated to use app/react-hooks and ssr, see below.

I expected that as in previous version of react-apollo, when querying only local data, the loading will always return false and I will get the data synchronously 

**Actual outcome:**
When querying only local data for the first render useQuery returns loading true and no data

![scrnli_8_19_2019_5-46-18 PM](https://user-images.githubusercontent.com/23150397/63274910-4be3e980-c2a9-11e9-85ae-91231bad9924.png)

this is not consistent with the apollo [docs](https://www.apollographql.com/docs/react/essentials/local-state/#querying-local-state) :
```
import React from "react";
import { useQuery } from "@apollo/react-hooks";
import gql from "graphql-tag";

import Todo from "./Todo";

const GET_TODOS = gql`
  {
    todos @client {
      id
      completed
      text
    }
    visibilityFilter @client
  }
`;

function TodoList() {
  const { data: { todos, visibilityFilter } } = useQuery(GET_TODOS);
  return (
    <ul>
      {getVisibleTodos(todos, visibilityFilter).map(todo => (
        <Todo key={todo.id} {...todo} />
      ))}
    </ul>
  );
}
```

**How to reproduce the issue:**

https://codesandbox.io/embed/github/kedarguy/apollo-3.0-next-local-state-server-side-bug-reproduction/tree/master/?fontsize=14

**Versions**
  System:
    OS: Linux 5.0 Ubuntu 18.04.3 LTS (Bionic Beaver)
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.16.0 - ~/.nvm/versions/node/v10.15.3/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  Browsers:
    Chrome: 76.0.3809.100
    Firefox: 68.0.2
  npmPackages:
    apollo-boost: 0.4.4 => 0.4.4 

Looking forward to be able to upgrade to the awesome apollo-hooks :)

youfoundron:
This may be related to #3338

hwillson:
Duplicate of https://github.com/apollographql/react-apollo/issues/3338. Let's continue the discussion over there. Thanks!

Issue #3338, opened 2019-08-10
different server-side rendering behaviour between v2 and v3


cusxio:
## Current Behaviour

`v2` and `v3` of `react-apollo` have vastly different behaviour/timings server side.

On `v3`, it seems that `react-apollo` has an additional render pass/intermediate state on the server side that breaks `v2` behaviour.

The **additional render pass** seems to output the following:

https://github.com/apollographql/react-apollo/blob/baf9a7648ac4ef2fc486f27719736b0d4e198340/packages/hooks/src/data/QueryData.ts#L172-L177

## Reproduction:

`v2`: https://codesandbox.io/s/with-apollo-xrj2c

`v3`: https://codesandbox.io/s/with-apollo-8luew

Check out the Terminal output difference of both `v2` and `v3`, 

`v2` only has two render pass, but `v3` has three render pass.

The additional render pass in `v3` currently breaks our app, as we rely on the `variable` data being there.

## Expected Behaviour

`v3` server side rendering behaviour should be similar to `v2`

NinjaOnRails:
I'm having a similar issue [here]( https://github.com/zeit/next.js/pull/8180#issuecomment-520226158)

HugoGiraudel:
I can confirm. `react-apollo@^3.0.0` has a rendering pass on the server with `loading` being `true`.

youfoundron:
Also seeing this issue. Updating Next did not fix this for me.

flipxfx:
We are seeing the same issue of the data on SSR still showing `loading: true`

mikebm:
Also experiencing this. The SSR is only returning our loading components.

mikebm:
I've been debugging this a little bit today and it appears to be isolated to `network-only` fetchPolicy. Still unsure why and will keep digging, but one could potentially use `cache-and-network` as workaround. I personally went with `network-only` because I was worried about the client side cache getting too big for my particular case since there is currently no TTL support.

mikebm:
@hwillson I was able to "resolve" my issue where a fetchPolicy of `network-only` causes only the loading component to render during SSR. It appears to be due to the initial observable getting initialized in execute() before fetchData is called. I was able to adjust the `updateObservableQuery` function with the below code:
```
	if (this.context.renderPromises && (newObservableQueryOptions.fetchPolicy === 'network-only' || newObservableQueryOptions.fetchPolicy === 'cache-and-network')) {
[...]

PR #3435, opened 2019-08-30
fix: ssr should support network-only & cache-and-network fetchPolicies. disabled ssr queries were still issuing requests

mikebm:
Related to issues discovered in topic #3338 
Might be related to #3390 

This PR addresses 2 issues and some cleanup.
Issue 1) `network-only` fetchPolicy during a SSR was resulting in the component being stuck in a loading state.

Issue 2) Disabling a query using `ssr:false` would result in the query still being executed on the server even though it was never used.

Cleanup:
GetOptions returns new objects with each call. This PR will reduce the amount of times it is called and thus reduces some garbage. In addition, with the changes on where the SSR observable was initialize, fetchData was greatly reduced in needed code.


packages/hooks/src/data/QueryData.ts
```diff
@@ -224,13 +218,21 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+        );
+      }
     }
   }
 
   private updateObservableQuery() {
     // If we skipped initially, we may not have yet created the observable
     if (!this.currentObservable.query) {
       this.initializeObservableQuery();
+      return;
```
mikebm: this was added as an optimization. updating the options doesn't need to happen because it was just initialized.

PR #3439, opened 2019-09-01
Update types.ts

lifeiscontent:
data is never undefined, and in typescript it should be Partial<TData> for better merging mechanics.

### Checklist:


abenhamdine:
> data is never undefined, and in typescript it should be Partial for better merging mechanics.

data *should have been* undefined when no data is available, see https://github.com/apollographql/react-apollo/pull/3388

lifeiscontent:
@abenhamdine I don't agree, data should always be an object, because the response of a graphql server is always an object, in cases where it should be undefined, it should just be an empty object, that way you can destructor keys that you expect and rather than having to default it to an object (which it will always be) just do it internally. @peggyrayzis explained this to me long ago at graphql conf. And I totally agree with her.

PR #3453, opened 2019-09-03
Better support for multiple `useLazyQuery` execution function calls

### LINKED ISSUE #3355: useLazyQuery ignores fetchPolicy 'network-only'
**Intended outcome:**

I had a useLazyQuery hook with a fetchPolicy of network-only. I expect that every time I run this I generate a network request and get a new response.

I was using this custom hook and then calling it from a click handler:
```
  const useCustHook = () => {
    const [executeQuery, result] = useLazyQuery(GQL_QUERY, {
      fetchPolicy: "network-only"
    });
    return async ({ foo }) => {
      await executeQuery({
        variables: { foo },
        fetchPolicy: "network-only"
      });
      console.log("response", result);
      return result;
    };
  };
```
...and then calling it from a click handler:
```
function QueryButotn() {
  const lazyQueryFn = useCustHook();
  return (
      <button
        onClick={() => {
          lazyQueryFn({ foo: "bar" });
        }}
      >
        Launch Lazy Query
      </button>
  );
}
```

**Actual outcome:**

Actual behavior was to send a network request on the first invocation. Subsequent runs would return the previously returned values and did not generate a network request.

I could fix the issue and get the desired behavior by swapping out useLazyQuery for useApolloClient like so:
```
const useCustHook = () => {
const client = useApolloClient();
    return async ({ foo }) => {
      const response = await client.query({
        query: GQL_QUERY,
        variables: { foo},
        fetchPolicy: "network-only"
      });
      console.log("Client.query response", response);
    };  }; 
```

I tried to use the codesandbox link to make a repro, but got tripped up on a cross origin error when I tried to update the linked app to use the ApolloProvider from @apollo/react-hooks. I can try again later, but wanted to at least submit.

**Version**

  System:
    OS: macOS 10.14.5
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.17.3 - ~/.nvm/versions/node/v10.16.0/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Browsers:
    Chrome: 76.0.3809.100
    Safari: 12.1.1
  npmPackages:
    apollo-cache-inmemory: ^1.6.2 => 1.6.2 
    apollo-client: ^2.6.3 => 2.6.3 
    apollo-link: ^1.2.12 => 1.2.12 
    apollo-link-context: ^1.0.18 => 1.0.18 
    apollo-link-error: ^1.1.11 => 1.1.11 
    apollo-link-http: ^1.5.15 => 1.5.15 
    apollo-link-ws: ^1.0.18 => 1.0.18 
    apollo-utilities: ^1.3.2 => 1.3.2 
    react-apollo: ^2.5.8 => 2.5.8 

not sure it's included in the output, but using "@apollo/react-hooks": "^3.0.0"

hwillson:
Before this PR, calling a `useLazyQuery` execution function multiple times in a row, when using a fetch policy of `network-only`, lead to unexpected results. Only the first network request was submitted as Apollo Client was blocking subsequent requests, since the query, fetch policy and variables remained the same. This commit adds additional cleanup to the `useLazyQuery` execution function, such that each call will start a new `ObservableQuery` instance. This means each query fired by the `useLazyQuery` execution function is treated as a fully new query.

Fixes #3355.


TitusEfferian:
thanks,
currently i'm upgrading to 3.1.0 version

Issue #3395, opened 2019-08-20
useQuery: on using refetch, 'loading' keeps true when same error returned twice or more


jet2jet:
When using `refetch` method on the return value of `useQuery`, and an error returned with the same `name` and `message` for the previous error, `loading` value of the return value keeps `true` (not changes to `false`), even if `notifyOnNetworkStatusChange` option is `true`.

**Intended outcome:**

`loading` changes to `false` whenever an error returned

**Actual outcome:**

`loading` keeps `true` when the same error returned (though it changes to `false` on the _first_ error)

**How to reproduce the issue:**

I created an repository with the reproduction code here:
https://github.com/jet2jet/react-apollo-test

After build and open `dist/index.html`, when clicking 'Refetch' at first, an error data will be shown. But when clicking 'Refetch' on the third time (`callCount = 4` on the log), 'Loading...' will be shown (an error data should be shown on this time).

**Version**

Note: I think this problem might not occur with react-apollo: 3.0.0

  npmPackages:
    apollo-cache-inmemory: ^1.6.3 => 1.6.3
    apollo-client: 2.6.4 => 2.6.4
    react-apollo: 3.0.1 => 3.0.1

jet2jet:
I think the following code is not working as expected...
https://github.com/apollographql/react-apollo/blob/be122f036c54892cc0632ed21f41fff9574ea096/packages/hooks/src/data/QueryData.ts#L288
(it seems that `equal` from `@wry/equality` does only compares `name` and `message` for Error object: https://github.com/benjamn/wryware/blob/master/packages/equality/src/equality.ts#L72 )

Kreozot:
I confirm the issue.
I've created the sandbox for this issue: https://codesandbox.io/s/quirky-dhawan-csdxw
First time error handled correctly, but when the query is updating, error is disappears and loading is forever true.

Kreozot:
❗️  Important notice: When I change version of @apollo/react-hooks from 3.0.1 to 3.0.0 the issue is not reproducing. So maybe we should find the error in changes in 3.0.1 version.

Kreozot:
So, for original issue I recommend to change version of react-apollo to 3.0.0 and try to reproduce.

hinok:
I struggle with the same issue.

I made codesandbox that shows this issue: https://codesandbox.io/s/react-apollo-3x-and-loading-true-i7ub1

This issue doesn't exist when `react-apollo@3.0.0` is used.

[...]

Issue #3468, opened 2019-09-07
fetchMore causes loading to be indefinitely true when getting an empty response


ostrebler:
**Intended outcome:**

`loading` being true while the fetchMore request is done, then merging the results with `updateQuery`, then rerendering with `loading` to false.

**Actual outcome:**

When fetchMore gets an empty array from the server, `updateQuery` is called as it should be, but the component never rerenders, thus `loading` being indefinitely true.

**Version**

@apollo/react-hooks : 3.1.0

The problem started with this version.
3.0.1 worked fine.

Victordmdb:
Not sure if this would help anyone but I resolved this because my `fetchMore` was unnecessarily being called immediately after render. After fixing that the issue stopped.

scf4:
I'm also experiencing this issue

scf4:
It seems like this happens when `updateQuery` returns `previousQueryResult`, regardless of the result.

dylanwulf:
I created a reproduction for this bug: https://codesandbox.io/s/reproduction-for-react-apollo-3468-jt3b5
@hwillson this seems like it could be related to PR #3433 


P.S. I love using `addMockFunctionsToSchema` in my client side tests, it's so much easier when the data is mocked automatically

scf4:
@dylanwulf I think your reproduction is for a different bug. This issue happens with a successful response of an empty array when no other data/state changes.

dylanwulf:
> @dylanwulf I think your reproduction is for a different bug. This issue happens with a successful response of an empty array when no other data/state changes.

@scf4 I am fairly confident that both issues are caused by the same bug, but just in case I created another reproduction with a successful response of empty array: https://codesandbox.io/s/reproduction-for-react-apollo-3468-with-successful-response-p4gmo

scf4:
@dylanwulf Oops, yeah I see what you're saying now. 👍

Quadriphobs1:
Same happens here, building pagination components and when I get a empty array response from the API the loading get stucked and it is not updated.

Quadriphobs1:
It looks like a fix has been proposed but it has not been merged https://github.com/apollographql/react-apollo/pull/3514

ehortle:
I am also experiencing the bug described by @scf4 . When I use `fetchMore`, it successfully makes the network request, but `networkStatus` is stuck on 3, and `loading` is stuck on `true`, but only when the result of `fetchMore` is the same as the previous result.

I am experiencing this with: 

[...]

PR #3477, opened 2019-09-08
fix: force update just after loading done with error even if the error is unchanged

jet2jet:
This PR will fix 'unchanging loading state' when an error, the _same_ with previous error, is occurred after `refetch` (described in #3395).  
\# *'same'* depends on `@wry/equality` implementation.

`previousData.result.loading` in `QueryData` should be `true` just before the first `error` handler called for any fetch/refetch, so this PR checks its value, and always calls `forceUpdate` if the value is `true`.  
(related PR: #3339)

After:

1. some fetch has triggered with `notifyOnNetworkStatusChange: true`
2. `loading` changes to true
3. an error occurrs for the fetch
4. `loading` changes to false and the error data can be used
5. call `refetch`
6. `loading` changes to true
7. an error occurrs for the refetch (the error message the same with the error occurred in step 3)
8. `loading` changes to false and the error data can be used (this does not occur in the previous source)

### Checklist:

### DIFF
--- a/packages/hooks/src/__tests__/useQuery.test.tsx
+++ b/packages/hooks/src/__tests__/useQuery.test.tsx
@@ -498,6 +498,226 @@ describe('useQuery Hook', () => {
         });
       }
     );
+
+    it('should render errors (different error messages) with loading done on refetch', async () => {
+      const query = gql`
+        query SomeQuery {
+          stuff {
+            thing
+          }
+        }
+      `;
+
+      const mocks = [
+        {
+          request: { query },
+          result: {
+            errors: [new GraphQLError('an error 1')]
+          }
+        },
+        {
+          request: { query },
+          result: {
+            errors: [new GraphQLError('an error 2')]
+          }
+        }
+      ];
+
+      let renderCount = 0;
+      function App() {
+        const { loading, error, refetch } = useQuery(query, {
+          notifyOnNetworkStatusChange: true
+        });
+
+        switch (renderCount) {
+          case 0:
+            expect(loading).toBeTruthy();
+            expect(error).toBeUndefined();
+            break;
+          case 1:
+            expect(loading).toBeFalsy();
+            expect(error).toBeDefined();
+            expect(error!.message).toEqual('GraphQL error: an error 1');
+            setTimeout(() => {
+              // catch here to avoid failing due to 'uncaught promise rejection'
+              refetch().catch(() => {});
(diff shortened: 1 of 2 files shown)


gburchfield:
Not sure if this is the place to ask, but this issue is the closest to what I'm seeing when using 'useLazyQuery' and I see in the package it uses the 'QueryData.js' that was fixed in this PR.

I'm using @apollo/react-hooks@3.1.0

**Experiencing the following:**

`const [login, {loading, error, data}] = useLazyQuery(LOGIN)`

1. Call `login` with bad credentials

1. `loading` goes from `true` to `false` as expected

1. trigger an `Alert` from the react-native package because `error` is not null/falsey for invalid authentication and click 'OK' to close message

1. Call `login` again with bad credentials

1. `loading` gets stuck on `true`

1. GraphQL error is the same, as expected, Authentication Invalid

**Expected Behavior:**

1. Call `login` with bad credentials

1. `loading` goes from `true` to `false`

1. Trigger `Alert` because of `error` then dismiss `Alert`

1. Call `login` again with bad credentials

1. `error` is cleared or nulled

1. `loading` goes from `true` to `false`

1. `error` is populated with a new Authentication Invalid error


Am I wrong to expect this behavior?  Should I be using `refetch`?  I tried to implement that but when calling `refetch`, `loading` was never changed to `true`.  I also tried to clone the master branch this morning to get this merge in to test with my project but I couldn't get it implemented correctly and have just relied on the npm install packages of botho `3.1.0` and `3.1.0.beta.0`.  For now my solution will just be to handle `loading` and `error` locally with `useState`.

dylanwulf:
@gburchfield this fix will be in the v3.1.1 release. If you still have the problem when v3.1.1 is released, please open a new issue

gburchfield:
Updated my @apollo/react-hooks package via npm to v3.1.1

I am still seeing the problem I described above:

 @https://github.com/apollographql/react-apollo/pull/3477#issuecomment-530424181

Also the issue could be similar to #3468 

My workaround is to just navigate to an entirely new component/screen, ensuring the component I'm using gets unmounted and then navigate back to my component fresh.  Probably the best for a login screen, but for other use cases I could see this being really annoying, especially if the work arounds I'm reading include using a forceUpdate()

Either way, great library so far! As Apollo matures, this will be my preference for both mobile and web.  Thanks guys!

PR #3490, opened 2019-09-12
Fix runtime error with fast-refresh

### LINKED ISSUE #3480: Throwing a runtime bug with useQuery and Fast Refresh (new React hot reloader)
**Intended outcome:**
Refresh of component. Hook state should be preserved.

**Actual outcome:**
On a fast refresh `useQuery` runs into a runtime error. See details in the image below. It seems that a value is undefined that is assumed to be defined. I think it is failing at this line:

https://github.com/apollographql/react-apollo/blob/65b3a88bea486e41b0a335f0afe65f110dae2498/packages/hooks/src/data/QueryData.ts#L346 

Interestingly there is a typescript nullability override, so maybe that is an assumption that we cannot make in the world of fast refresh?

image

Run any `useQuery` in React-Native 0.61 RC2.

**Version**
@apollo/hooks 3.10
React-Native 0.61 RC2

jfrolich:
Fixes #3480


TheRusskiy:
actually seems that `currentObservable` can be null as well after fast refresh

PR #3496, opened 2019-09-14
fix: argument type for onCompleted in query

G07cha:
'data' also could be an empty object according to documentation and real-life usage I've observed

See https://www.apollographql.com/docs/react/essentials/queries/#usequery-api


hwillson:
Thanks for the PR @G07cha. As of React Apollo 3.1.0 (and https://github.com/apollographql/react-apollo/pull/3388 in particular), `data` should no longer be an empty object. Are you still seeing this happen with 3.1.0? If so, would you mind opening a new issue with a small runnable reproduction? Thanks!

G07cha:
@hwillson indeed, I've missed that one. It works as expected in 3.1.0, thanks!

PR #3497, opened 2019-09-15
Call `onCompleted` each time `useLazyQuery` executes

### LINKED ISSUE #3488: useLazyQuery / useQuery hook still isn't fired after fetching data from the cache
**Intended outcome:**

`OnCompleted` should be called after query result is fetched.
The problem appears to be solved only with the <Query> component as stated here:
https://github.com/apollographql/react-apollo/issues/2177
With the useQuery or useLazyQuery hooks the problem still appears. 

If i set `{ fetchPolicy: "no-cache" }` for the query the onCompleted will be called as expected.
 
**Actual outcome:**
onCompleted is only called after the first call/render. On any subsequent calls it is not invoked.

Here is a sandbox to reproduce the problem: https://codesandbox.io/s/vigilant-maxwell-gqyf2?fontsize=14

onCompleted is only called after the first call to Load() as shown in the console.

**Version**
`react-apollo: v3.1.0 release
`

hwillson:
Make sure `onCompleted` is called each time a `useLazyQuery` based query completes, after the execution function is called.

Fixes #3488

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,8 @@
   [@jet2jet](https://github.com/jet2jet) in [#3477](https://github.com/apollographql/react-apollo/pull/3477)
 - Add back in the removed `ChildDataProps` and `ChildMutateProps` types. <br/>
   [@hwillson](https://github.com/hwillson) in [#3495](https://github.com/apollographql/react-apollo/pull/3495)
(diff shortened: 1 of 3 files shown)


vilvaathibanpb:
@hwillson Can I know the exact reason for adding `undefined` to useEffect ?

PR #3514, opened 2019-09-21
Fix an error that the networkStatus did not change to ready at the end of pagination

mu29:
Solves #3468

Although #3433 has improved performance by reducing the number of renders in pagination, it only works when the fetchMore function returns a new list.

If we reached to the end of the list, the fetchMore function returns the same length of result, which is determined as equal by the [@wryware/equality library](https://github.com/benjamn/wryware/blob/50e7d6181b0b937ad54b543118e823ad98e2bca9/packages/equality/src/equality.ts#L40) we currently use. Therefore, reading the last page of the list remains `networkStatus` to `NetworkStatus.fetchMore`.

I think [apollo-client](https://github.com/apollographql/apollo-client) needs to send the updated fetchMore result and the changed networkStatus at once.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,8 +6,10 @@
 
 - Make sure SSR is fully disabled when using `ssr: false` and `ssrMode: true`. <br/>
   [@maapteh](https://github.com/maapteh) in [#3515](https://github.com/apollographql/react-apollo/pull/3515)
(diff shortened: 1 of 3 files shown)


mattdell:
Have you signed the agreement @mu29 ?

I need this fix 🙂

mu29:
Sure :) As you can see, all checks have passed.

PR #3536, opened 2019-09-26
@apollo/client integration

hwillson:
This PR introduces the `@apollo/client` `beta` into React Apollo. This means common, hooks, and testing utilities are now pulled into this project from Apollo Client, and re-exported as needed. All `@apollo/react-*` packages will continue to work as they did before, but the common, hooks and testing packages are now just wrappers around `@apollo/client` functionality. 

Moving forward, our recommendation will be to just use `@apollo/client` and its integrated React hooks functionality, but this repo and its child packages will still be supported until sometime in 2020.


lifeiscontent:
@hwillson why would you call it `@apollo/client`? when it isn't generically something for JavaScript and more so react?

hwillson:
Tests will fail (caused by https://github.com/apollographql/react-apollo/pull/3536/commits/b049ee20052c23c69d608c592ab0e131e19fcb9e) until https://github.com/apollographql/apollo-client/pull/5639 is merged, and a new AC `beta` is released.

lifeiscontent:
@hwillson when do you plan on releasing this?

config/jest.config.js
```diff
@@ -10,8 +10,10 @@ module.exports = {
   },
   coverageDirectory: './meta/coverage',
   moduleNameMapper: {
-    '^@apollo\\/react-([^/]+)': '<rootDir>/$1/src',
+    '^@apollo\\/react-components': '<rootDir>/components/src',
+    '^@apollo\\/react-hoc': '<rootDir>/hoc/src',
+    '^@apollo\\/react-ssr': '<rootDir>/ssr/src',
   },
-  testMatch: ['<rootDir>/*/src/**/__tests__/**/*.test.ts?(x)'],
+  //testMatch: ['<rootDir>/*/src/**/__tests__/**/*.test.ts?(x)'],
```
lifeiscontent: @hwillson was it intentional to leave this commented out?

Changelog.md
```diff
@@ -1,5 +1,36 @@
 # Change log
 
+## 4.0.0 (TBD - not yet released)
+
+> ⚠️ **Deprecation Notice** ⚠️
+>
+> Please note that as of version 4.0.0, the React Apollo project has been deprecated. `@apollo/react-hooks` and `@apollo/react-testing` functionality is now available directly from the `@apollo/client` package (refer to the [Apollo Client 3 migration guide](https://www.apollographql.com/docs/react/) for more details). Both packages still work as is, but their functionality is now re-exported from `@apollo/client`. If you're only using hooks, we recommend using the hooks from `@apollo/client` directly, and dropping your React Apollo dependency(ies). React Apollo's `graphql` HOC and render proper components are now in maintenance mode, meaning they will continue to receive important bug fixes, but will not be updated with new functionality.
```
benjamn: This first sentence is kind of alarming, because it makes it sound like we're removing support for using React with Apollo Client.

Can we frame this more like "You should prefer to import React-related utilities from `@apollo/client`. While we still have separate `@apollo/react-*` packages for ease of migration, they are no longer under active development, as they simply re-export from `@apollo/client`."?

packages/all/package.json
```diff
@@ -30,30 +30,28 @@
...
+  },
+  "devDependencies": {
+    "rollup": "1.29.0",
+    "typescript": "3.7.4"
   },
   "dependencies": {
     "@apollo/react-common": "file:../common",
     "@apollo/react-components": "file:../components",
     "@apollo/react-hoc": "file:../hoc",
-    "@apollo/react-hooks": "file:../hooks",
```
benjamn: Why remove `@apollo/react-hooks` here? Don't we want to export hooks-related stuff from `react-apollo`?

Ah, I think I see now: since the exports of `@apollo/react-hooks` are a subset of the exports of `@apollo/react-common`, we don't need to import `@apollo/react-hooks` for the combined `react-apollo` package?

If that's accurate, I think I'd like to keep depending on `@apollo/react-hooks` here, just because the relationship between it and `@apollo/react-common` could change.

packages/common/package.json
```diff
@@ -1,57 +1,28 @@
...
-    "test:cjs": "npm run build && npx jest --config ../../config/jest.cjs.config.js --testPathPattern packages/common",
-    "test:umd": "npm run build && npx jest --config ../../config/jest.umd.config.js --testPathPattern packages/common"
-  },
-  "peerDependencies": {
-    "@types/react": "^16.8.0",
-    "apollo-client": "^2.6.4",
-    "apollo-utilities": "^1.3.2",
-    "graphql": "^14.3.1",
-    "react": "^16.8.0"
+    "clean": ""
```
benjamn: Does this still need these empty `clean` scripts?

packages/hooks/index.js
```diff
@@ -0,0 +1,19 @@
+const common = require('@apollo/react-common');
```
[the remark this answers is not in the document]
hwillson: The original idea here was that `@apollo/react-common` would give full access to `@apollo/client` for people that need it (and for the tests we have in the various React Apollo packages). The individual packages that depend on `@apollo/react-common` would then only re-export the specific parts they needed to share, which is why the hooks index only re-exports a subset. Same thing for the `all` (`react-apollo`), `components` and `hoc` packages. They’re only re-exporting parts of Apollo Client, from `@apollo/react-common`.

Thinking about this more now though, I don’t think we need to filter any of the Apollo Client exports from `@apollo/react-common`, or in any package. Exporting everything doesn’t impact bundle size much, and all we’re doing by restricting certain exports is making things a bit more tedious for developers when it comes to managing their imports. Also, the additional maintenance overhead of managing re-exports, and remembering to update re-export lists when we make changes, doesn’t seem worth it.

I’ll adjust things to keep the full `@apollo/client` re-export from `@apollo/react-common`, as well as a full `@apollo/client` re-export in all packages that use common. Thanks!

packages/common/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('@apollo/client');
```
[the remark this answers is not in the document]
hwillson: See https://github.com/apollographql/react-apollo/pull/3536#discussion_r368232773 - thanks!

config/rollup.config.js
```diff
@@ -15,10 +15,7 @@ const defaultGlobals = {
   '@apollo/react-hooks': 'apolloReactHooks',
   '@apollo/react-testing': 'apolloReactTesting',
   '@apollo/react-ssr': 'apolloReactSSR',
-  'apollo-client': 'apolloClient',
-  'apollo-utilities': 'apolloUtilities',
-  'apollo-cache': 'apolloCache',
-  'apollo-cache-inmemory': 'apolloCacheInMemory',
+  '@apollo/client': 'apolloClient',
```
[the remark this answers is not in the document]
hwillson: Definitely - will do, thanks!

packages/common/package.json
```diff
@@ -9,49 +9,20 @@
     "react"
   ],
   "license": "MIT",
-  "main": "./lib/react-common.cjs.js",
-  "module": "./lib/react-common.esm.js",
```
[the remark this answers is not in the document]
hwillson: Sure thing!

Changelog.md
hwillson: Much better - thanks!

packages/all/package.json
hwillson: Exactly, and no problem on adding the dependency back.

packages/common/package.json
hwillson: Without these `lerna` complains when we run `npm run clean` from the monorepo root. We could adjust lerna to silence its complaints, or handle multi package cleaning without lerna, but leaving them as empty script references was the easiest. Happy to adjust further if you think we should. Thanks!

hwillson:
This should be ready for another look @benjamn - thanks!

PR #3575, opened 2019-10-08
Prevent runMutation from throwing an error

brapifra:
This relates to https://github.com/apollographql/apollo-client/issues/4283

`useMutation` shouldn't throw any error when it's being run. The error is already passed in the `MutationResult` object.


brapifra:
Some tests are failing. I am happy to fix them if you agree that this should be the expected behaviour.

brapifra:
Could this be reviewed by somebody? Maybe @hwillson?

PR #3582, opened 2019-10-11
Add missing client property to useLazyQuery Result object.

bwlt:
From [the documentation](https://www.apollographql.com/docs/react/api/react-hooks/#result-1) the result object returned by `useLazyQuery` should contains the `client` property.

## Side note
In this point of code there is an unsafe casting which prevent to spot the error using the type system.
Probably other properties are missing.


bwlt:
duplicated of: https://github.com/apollographql/react-apollo/pull/3581

PR #3583, opened 2019-10-11
Ensure updateQuery is included in QueryResult with useLazyQuery

bwlt:
From [the documentation](https://www.apollographql.com/docs/react/api/react-hooks/#result-1) the result object returned by `useLazyQuery` should contains the `updateQuery` property.

## Side note
In this point of code there is an unsafe casting which prevent to spot the error using the type system.
Probably other properties are missing.

### Checklist:


bwlt:
The resolution does not convince me because of this scenario:
what about if a user wants to update the query data with data even if the execution was not ran?
With this changes the `updateQuery` throws an invariant error.
What do you think?

PR #3606, opened 2019-10-17
Make partialRefetch work like it did in v2.x

### LINKED ISSUE #3605: partialRefetch behaving differently (incorrectly?) from v2.x
**Intended outcome:**
If I have a query rendered like this:
```gql
query MyQuery {
  user {
    id
    todos { # list of todos
      id
      title
      completed
    }
  }
}
```

And I've just executed a mutation like this:
```gql
mutation AddTodoToUser {
  addTodoToUser { # appends a new todo to the user
    user {
      id
      todos {
        id
      }
    }
  }
}
```

If I have `partialRefetch` on for the query, I would expect the query to be re-fetched to get the missing data for the new todo. This is what happened before I upgraded from 2.x.

**Actual outcome:**
The query is not re-fetched, and the query continues to render the stale data that does not include the new todo.

I dug through the source code, and I think this code from the `QueryManager` is part of the problem:
https://github.com/apollographql/apollo-client/blob/d61e90926cb5f5c559f3daa1c974fac225b1c5b2/packages/apollo-client/src/core/QueryManager.ts#L627-L637

Because `returnPartialData` is `false`, the query manager interprets the new partial data as `stale`. And this causes the query manager to not provide its subscribers with the new data. Therefore, [this codeblock](https://github.com/apollographql/react-apollo/blob/07e404b089017f8a653d721c169a55191bf88f9b/packages/hooks/src/data/QueryData.ts#L373-L393) is never hit, so the refetch never happens.

I can't just start passing `returnPartialData` to my queries because I don't actually want to receive any partial data—I just want the `Query` component/hook to receive it so that it can trigger the partial refetch.

I'm happy to help with a PR for this but would appreciate any direction you can give me. Would, for example, changing [this code](https://github.com/apollographql/apollo-client/blob/d61e90926cb5f5c559f3daa1c974fac225b1c5b2/packages/apollo-client/src/core/QueryManager.ts#L627-L637) to this:
```typescript
        const stale = isMissing && !(
          options.returnPartialData ||
          options.partialRefetch || // added this line
          fetchPolicy === 'cache-only'
        );

        const resultFromStore: ApolloQueryResult<T> = {
          data: stale ? lastResult && lastResult.data : data,
          loading,
          networkStatus: queryStoreValue.networkStatus,
          stale,
        };
```
Be an appropriate solution? I'm guessing not, since I don't think it makes sense for the `QueryManager` to care about a feature like `partialRefetch`, which is implemented at the `react-apollo` level.

Set up a query/mutation as described above!

**Version**

```
  System:
    OS: macOS 10.15
  Binaries:
    Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.5.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
  Browsers:
    Chrome: 77.0.3865.120
    Firefox: 66.0.2
    Safari: 13.0.2
```

minznerjosh:
Fixes #3605.

### Checklist:

### DIFF
--- a/packages/hooks/src/__tests__/useQuery.test.tsx
+++ b/packages/hooks/src/__tests__/useQuery.test.tsx
@@ -1184,7 +1184,7 @@ describe('useQuery Hook', () => {
       cache.writeFragment({
         fragment: OWNER_FRAGMENT,
         data: OWNER_FRAGMENT_DATA,
-        id: '1612'
+        id: 'Owner:1612'
       });
       // Because the data of the third car is partial, and because
       // `partialRefetch` is `true`, the hook re-fetches data
--- a/packages/hooks/src/__tests__/useQuery.test.tsx
+++ b/packages/hooks/src/__tests__/useQuery.test.tsx
@@ -1127,7 +1127,9 @@ describe('useQuery Hook', () => {
             .mockReturnValueOnce({ data: OWNER_RESULT_DATA_WITH_NEW_CAR })
         }
       ];
+      const link = new MockLink(OWNER_MOCKS);
       const cache = new InMemoryCache();
+      const client = new ApolloClient({ link, cache });
 
       const dataSpy = jest.fn();
       const errorSpy = jest.fn();
@@ -1149,9 +1151,9 @@ describe('useQuery Hook', () => {
       };
 
       render(
-        <MockedProvider mocks={OWNER_MOCKS} cache={cache}>
+        <ApolloProvider client={client}>
           <Component />
-        </MockedProvider>
+        </ApolloProvider>
       );
 
       // Data loads as normal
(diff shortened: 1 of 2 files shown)


JKillian:
Any updates on this @hwillson? Would be amazing to have this fixed, would let us (and others I'm sure) unblock our upgrade to the hooks version of `react-apollo`!

glentakahashi:
Any status on this? We're hitting the same issue as well

minznerjosh:
Looks like all current development on the react library has moved to apollographql/apollo-client. Hoping the issue this PR fixes won't be an issue in that implementation. Hopefully that version lands soon!

PR #3645, opened 2019-10-30
[RFC] Allow startPolling without having to exec first

masad-frost:
### Checklist:

Related to https://github.com/apollographql/apollo-client/issues/5140

When using `useLazyQuery`, the return value does not include `ObservableQueryFields`, hence you cannot call `startPolling` until after you call `runLazyQuery` which is inconvenient.

At the very least the type of the return value of `useLazyQuery` should be modified to indicate that `ObservableQueryFields` maybe undefined.

### DIFF
--- a/packages/hooks/src/__tests__/useLazyQuery.test.tsx
+++ b/packages/hooks/src/__tests__/useLazyQuery.test.tsx
@@ -389,4 +389,43 @@ describe('useLazyQuery Hook', () => {
       });
     }
   );
+
+  it('should support polling programatically', done => {
+    let renderCount = 0;
+    const Component = () => {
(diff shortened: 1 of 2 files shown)


masad-frost:
Also wondering why lazy queries weren't just a convenience for switching `skip` from `true` to `false` and then `forceUpdate`ing, wouldn't that simplify the internals?

masad-frost:
@hwillson Looks like you were interested in looking at the problem in https://github.com/apollographql/apollo-client/issues/5140

This PR explains why this is happening and adds a POC for a solution

masad-frost:
Any body looks at PR's here? :/

Issue #3505, opened 2019-09-17
useLazyQuery onCompleted called on every re-render


PovilasSlekys:
**Intended outcome:**
useLazyQuery only calls onCompleted on a successful network request

**Actual outcome:**
useLazyQuery calls onCompleted after every re-render even if the result is being taken from the cache

**Version**

  npmPackages:
    apollo-cache-inmemory: ^1.6.3 => 1.6.3
    apollo-client: ^2.6.4 => 2.6.4
    apollo-link-context: ^1.0.19 => 1.0.19
    apollo-link-http: ^1.5.16 => 1.5.16
    react-apollo: ^3.1.1 => 3.1.1

huongdevvn:
The same happen for me. `onCompleted` of `useLazyQuery` is called only once when I execute simple functions like `console.log`
```
const [fetchLazy] = useLazyQuery(LAZY_QUERY, {
    variables: { id },
    onCompleted: data => {
      console.log('data ', data);
    }
  });
```

But if I execute a function returned by `React hooks` like: useState or useReducer, `onCompleted` will happen many times or infinite times.

```
  const [data, setData] = useState(null);

  const [fetchLazy] = useLazyQuery(LAZY_QUERY, {
    variables: { id },
    onCompleted: data => {
      console.log('data ', data);
      setData(data); // multiple times
    }
  });
```


```
  const [state, dispatch] = useReducer(reducer, initialState);

  const [fetchLazy] = useLazyQuery(LAZY_QUERY, {
    variables: { id },
    onCompleted: data => {
      console.log('data ', data);
      dispatch({ type: 'completed', payload: 'data' }); // infinite times
    }
  });
```

huongdevvn:
@hwillson Could you take a look?

vilvaathibanpb:
I can work on this issue if @hwillson  or @bsonntag haven't started working on this.

bsonntag:
@vilvaathibanpb  I took a look and figured the problem is here: https://github.com/apollographql/react-apollo/blob/master/packages/hooks/src/utils/useBaseQuery.ts#L57. I didn't start working on fix tho, so go ahead.

vilvaathibanpb:
@bsonntag  Sure, Let me raise a PR against it. Thanks

vilvaathibanpb:
Hey @bsonntag , I was checking the issue and my observations are: 

1. For a normal query, the `useEffect` hook is checked only once for any state change.
[...]

PR #3789, opened 2020-01-08
fix proper typing lazy hook options

### LINKED ISSUE #3787: types for lazy hook are incorrect (only two available)
I'm using 3.1.2 and see the generated types for a lazy hook are missing a lot of the default options.

The query (non-lazy) one extends on the base which has all options as documented and correctly.
```
export interface QueryOptions<TData = any, TVariables = OperationVariables> extends QueryFunctionOptions<TData, TVariables> {
    children?: (result: QueryResult<TData, TVariables>) => ReactNode;
    query: DocumentNode;
}
```

While the lazy options are only referring to:
```
export interface QueryLazyOptions<TVariables> {
    variables?: TVariables;
    context?: Context;
}
```

So at the moment, i'm unable to set for example `fetchPolicy` as documented on https://www.apollographql.com/docs/react/api/react-hooks/#uselazyquery when using your generated types. When using the `useLazyQuery` from the library:

```
export declare function useLazyQuery<TData = any, TVariables = OperationVariables>(
    query: DocumentNode,
    options?: LazyQueryHookOptions<TData, TVariables>
): [(options?: import("./types").QueryLazyOptions<TVariables> | undefined) => void, import("@apollo/react-common").QueryResult<TData, TVariables>];
```

This is a bug :) and created a PR for it

environment:
```
  System:
    OS: macOS Mojave 10.14.1
  Binaries:
    Node: 10.16.3 - ~/.nvm/versions/node/v10.16.3/bin/node
    Yarn: 1.19.1 - ~/.nvm/versions/node/v10.16.3/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.3/bin/npm
  Browsers:
    Chrome: 79.0.3945.117
    Firefox: 69.0.3
    Safari: 12.0.1
  npmPackages:
    @apollo/react-hooks: 3.1.2 => 3.1.2 
    @apollo/react-ssr: 3.1.2 => 3.1.2 
    @apollo/react-testing: 3.1.2 => 3.1.2 
    apollo-cache-inmemory: 1.6.3 => 1.6.3 
    apollo-client: 2.6.4 => 2.6.4 
    apollo-link: 1.2.13 => 1.2.13 
    apollo-link-batch-http: 1.2.13 => 1.2.13 
    apollo-link-error: 1.1.12 => 1.1.12 
    apollo-link-http: 1.5.16 => 1.5.16 
    apollo-link-logger: 1.2.3 => 1.2.3 
    apollo-link-schema: 1.2.4 => 1.2.4 
    apollo-link-state: 0.4.2 => 0.4.2 
```

maapteh:
The lazy types are incorrect, the base options are not being added.

closes #3787

(previous PR was borked because of misconfiguration on current machine)

### DIFF
--- a/packages/hooks/src/data/QueryData.ts
+++ b/packages/hooks/src/data/QueryData.ts
@@ -30,7 +30,7 @@ export class QueryData<TData, TVariables> extends OperationData {
   private forceUpdate: any;
 
   private runLazy: boolean = false;
-  private lazyOptions?: QueryLazyOptions<TVariables>;
+  private lazyOptions?: QueryLazyOptions<TData, TVariables>;
 
   constructor({
     options,
@@ -134,7 +134,7 @@ export class QueryData<TData, TVariables> extends OperationData {
     return options;
   }
 
-  private runLazyQuery = (options?: QueryLazyOptions<TVariables>) => {
+  private runLazyQuery = (options?: QueryLazyOptions<TData, TVariables>) => {
     this.cleanup();
 
     this.runLazy = true;
--- a/packages/hooks/src/types.ts
+++ b/packages/hooks/src/types.ts
@@ -61,13 +61,14 @@ export interface QueryCurrentObservable<TData, TVariables> {
   subscription?: ZenObservable.Subscription;
 }
 
-export interface QueryLazyOptions<TVariables> {
+export interface QueryLazyOptions<TData, TVariables>
+  extends QueryFunctionOptions<TData, TVariables> {
   variables?: TVariables;
   context?: Context;
 }
 
 export type QueryTuple<TData, TVariables> = [
-  (options?: QueryLazyOptions<TVariables>) => void,
+  (options?: QueryLazyOptions<TData, TVariables>) => void,
   QueryResult<TData, TVariables>
 ];
 


krvajal:
It seems, from this signature 
```
export interface QueryLazyOptions<TVariables> {
    variables?: TVariables;
    context?: Context;
}
```
that you can't pass a `onCompleted` or `onError` option when calling the lazy query function. 
Is this correct.

My use case:
```js 
const [lazyQuery] = useLazyQuery(query);
function getData(variables){
return new Promise((resolve,reject) => lazyQuery({variables, onComplete: resolve, onError: reject))
}
```

maapteh:
No that is not correct, and that's why this PR has been made. To solve the types. If you don't use TS like your example (and also in their tests) you will be able to do this since these are legit options.

krvajal:
> since these are legit options.

I don't think they are. They are never called.

maapteh:
But then that's a bug and not belonging to this PR, see https://github.com/apollographql/react-apollo/issues/3505 for example

maapteh:
@krvajal to help you and make this PR still usefull :) 

```
const [execute, data, error] = useLazyQuery(query, { onCompleted: () => {},onError: () => {}, );
...
execute(variables)
```

You dont need a Promise and your execution function is only accepting context and variables, the other options just are the second argument in the lazy Query. I read it wrong, just as you :) But the code above works :) Fijne dag en hopelijk hielp deze PR je dan toch een beetje

PR #3815, opened 2020-01-23
Test use case where you have resolvers for some data, but not all of it

lifeiscontent:
### Checklist:

* [X] Make sure all of the significant new logic is covered by tests

@hwillson @benjamn I found an issue with apollo client.

essentially if you have `resolvers` for apollo local state, but you don't specify resolvers for *ALL* of your local data, the Client doesn't do what you would expect.

this simple change in a test doesn't do a full test case of this functionality but from a high level this is what you would expect:

you might have the client data:

``` typescript
interface LocalState {
  todos: {title: string, done: boolean}[];
  visibilityFilter: 'NEW' | 'ALL' | 'DONE';
}
```

and you might have a localState resolver for the `todos` but maybe not for the `visibilityFilter` because it's rather simple. in this case, ApolloClient will not resolve the visibilityFilter data as you would expect.

### DIFF
--- a/packages/hoc/src/__tests__/MockedProvider.test.tsx
+++ b/packages/hoc/src/__tests__/MockedProvider.test.tsx
@@ -630,7 +630,16 @@ describe('@client testing', () => {
       const ContainerWithData = withNetworkStatus(Container);
 
       render(
-        <MockedProvider mocks={networkStatusMocks}>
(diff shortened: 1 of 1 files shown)


lifeiscontent:
ping @hwillson

PR #3902, opened 2020-03-27
Prevent new data re-render attempts during an existing render

### LINKED ISSUE #3863: React 16.13.0 'Warning: Cannot update a component from inside the function body of a different component.'
**Intended outcome:**

After updating to React 16.13.0, `useQuery` should not throw unintentional state changes warning.

**Actual outcome:**

After updating to React 16.13.0, in dev mode, every component that uses `useQuery` **without** `skip: true` will throw this new warning. 

`Warning: Cannot update a component from inside the function body of a different component.`

image

More info about this warning can be found in the [React changelog](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render).

The actual functionality appears to be the same, from what I can tell.

Update your React version to `16.13.0` where [this warning was introduced](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#react-dom) and it will appear in the console once for each component using `useQuery` without `skip: true`.

**Version**
```
  System:
    OS: Windows 10 10.0.18362
  Binaries:
    Node: 11.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.19.1 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    apollo-cache-inmemory: ^1.6.5 => 1.6.5
    apollo-client: ^2.6.8 => 2.6.8
    apollo-link: ^1.2.13 => 1.2.13
    apollo-link-error: ^1.1.12 => 1.1.12
    apollo-link-http: ^1.5.16 => 1.5.16
    apollo-upload-client: ^12.1.0 => 12.1.0
    apollo-utilities: ^1.3.3 => 1.3.3
    react-apollo-network-status: ^3.0.0 => 3.0.0
```

Update 05.03: 
 - thanks to @dannycochran, we have a narrow reproduction case https://github.com/bogdibota/usequery
 - my idea so far: This seems to be caused by a component A updating the cache and another component B that listens to the same query as component A gets updated during the rendering phase.
 - the reason my initial error indicated that every component throws this warning is that i am using `react-apollo-network-status` for a global loading indicator. This loading indicator was updated by all the components that had a `useQuery` hook (as expected)
 - IMPORTANT: the warning's stacktrace is showing the component that initialed the changes, not the one that gets re-renderd by those changes 🤦‍♂

hwillson:
As of version 16.13.0, React logs a warning when a function component is updated during another component's render phase (facebook/react#17099). In React Apollo, this warning can be triggered when nesting multiple components that leverage `useQuery`. To help avoid this, this PR ensures re-render requests to show new data are delayed until an effect hook is run to handle them (since we're then out of the render phase).

Fixes #3863


packages/hooks/src/data/QueryData.ts
```diff
@@ -206,8 +207,8 @@ export class QueryData<TData, TVariables> extends OperationData {
     // See if there is an existing observable that was used to fetch the same
     // data and if so, use it instead since it will contain the proper queryId
     // to fetch the result set. This is used during SSR.
-    if (this.context && this.context.renderPromises) {
-      this.currentObservable.query = this.context.renderPromises.getSSRObservable(
+    if (this.ssrInitiated()) {
+      this.currentObservable.query = this.context?.renderPromises?.getSSRObservable(
```
jasonwr: The optional chaining operator is a nice addition here.

packages/hooks/src/utils/useBaseQuery.ts
```diff
@@ -18,21 +18,39 @@ export function useBaseQuery<TData = any, TVariables = OperationVariables>(
   const context = useContext(getApolloContext());
   const [tick, forceUpdate] = useReducer(x => x + 1, 0);
   const updatedOptions = options ? { ...options, query } : { query };
+  const isRendering = useRef(true);
```
jasonwr: Why not just have a useState hook for each then if you really want have separate useEffect to subscribe to any variable changes? Just curious. I mean if it works it works I just don't understand why the useRef versus useState.

packages/hooks/src/data/QueryData.ts
hwillson: Thanks @jasonwr - I agree. I've started sprinkling its use in, but there are definitely more places it can be used for sure.

packages/hooks/src/utils/useBaseQuery.ts
hwillson: In this case I'm interested in using `isRendering` and `isRenderScheduled` to track render metadata, not to trigger an actual render. Doing this is easier with `useRef` since we don't ever have to worry about triggering an unintended render. We could likely achieve the same thing with `useState` and guarding/effect dep tracking, but in all honesty I kinda like refs 🙂.

jasonwr: @hwillson wow that's a clever trick. I would have just used ```useState``` but you're right in that ```useRef``` won't trigger an unintended render and it's mutable. That's pretty slick. I'll have to try this out sometime. From the React docs: However, useRef() is useful for more than the ref attribute. It’s handy for keeping any mutable value around similar to how you’d use instance fields in classes... useRef will give you the same ref object on every render.

Vadorequest:
This creates a SSR regression when used by Next.js framework.

https://github.com/lfades/next-with-apollo/issues/126 similar regression (SSR) when upgrading from v4 to v5 (which uses `@apollo/react-ssr`)

Tested on https://github.com/UnlyEd/next-right-now/pull/27

I don't think it's gonna be blocking (for me) but others could run into similar issue. If so, I believe following https://github.com/lfades/next-with-apollo/issues/126#issuecomment-608442105 is the way to go, but I'm about to try and will confirm it.

PR #4021, opened 2020-06-12
Fixed unmounted component state update

mccraveiro:
In my project I'm getting the following error with useQuery:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in %s.%s, a useEffect cleanup function,
```

After some debugging, I came into this line: 
https://github.com/apollographql/react-apollo/blob/4bd40a94b18102b1eb42864d2e59d52ced0303f5/packages/hooks/src/utils/useBaseQuery.ts#L34

Because of the re-render in a microtask this may cause the hook to update the state of an unmounted component. Adding a `isMounted` check fixed the issue. 

I'm not sure how to add tests for this. Any suggestions?

### DIFF
--- a/packages/hooks/src/utils/useBaseQuery.ts
+++ b/packages/hooks/src/utils/useBaseQuery.ts
@@ -16,6 +16,7 @@ export function useBaseQuery<TData = any, TVariables = OperationVariables>(
   lazy = false
 ) {
   const context = useContext(getApolloContext());
+  const isMounted = useRef(false);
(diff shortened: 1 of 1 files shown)


mccraveiro:
@benjamn @hwillson Should I move this PR to the apollo-client repository? This patch will need to be applied on both.

